PR #9, opened 2019-06-29
proved Kelly's coherence proof of monoidal category

HuStmpHrrr:
I built a bunch of infra to reason about isos in order to formalize Kelly's. not sure I overshoot anything.


HuStmpHrrr:
@JacquesCarette it would be nice if you can review this PR and see if all these are necessary.

I know that there is something I invented can get into stdlib, e.g. `IsoPlus`. this is something I can do in parallel, c.f. agda/agda-stdlib#828.

PR #19, opened 2019-07-11
define dinatural transformation

HuStmpHrrr:
c.f. https://github.com/agda/agda/issues/3912

there seems something wrong with generalizable variables in Agda. I will let this hang until problems are identified.


HuStmpHrrr:
the desired definitions should be just based on generalisable variables but due to an issue in Agda, it seems currently not feasible. ideally,
```agda
_<∘_ : NaturalTransformation G H → DinaturalTransformation F G → DinaturalTransformation F H
_<∘_ = {!!}
```

but this doesn't work.

Dinatural transformation can be used in the definition of closed categories, so I will merge it for now.

PR #22, opened 2019-07-14
define terminal category

HuStmpHrrr:
c.f. https://github.com/copumpkin/categories/blob/master/Categories/Terminal.agda

### DIFF
--- a/Categories/Category/Instance/One.agda
+++ b/Categories/Category/Instance/One.agda
@@ -1,0 +1,57 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Level
+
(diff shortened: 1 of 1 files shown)


HuStmpHrrr:
To make the definition truly range over all `Category o ℓ e`, it seems it is necessary to define a universe polymorphism unit type and the equality between morphisms must not be propositional equality.

JacquesCarette:
An alternative could be to use `Lift` everywhere -- but that would be quite ugly. I think this is best.  Although let me ask about universe-polymorphic `Unit` (as an option) for stdlib.

PR #26, opened 2019-07-19
[ refactor ] let Agda infer the easy parts

gallais:
Removed all the fields that could be solved by eta-expansion.


JacquesCarette:
Hmm, that's going to clash mightily with a different version of the same file that I was about to check in. Let me try to do the same change to my version, and if that works, I'll close this PR but your idea will still make it in.  Otherwise I'll probably revert my changes, merge this, and see which way to go forward.

JacquesCarette:
I was able to use this idea successfully (see commit 7fb3daaf40) here -- and elsewhere as well. Thanks.

PR #33, opened 2019-07-23
do not use Yoneda lemma

HuStmpHrrr:
is it really better than applying Yoneda lemma though? the compilation time definitely differs.


JacquesCarette:
No, Yoneda does not have to be used. This is quite clear too.  How much faster? I mean, on your machine, how long did the original take versus this one?

HuStmpHrrr:
I did not measure it but I can feel I am more patient with this proof.

HuStmpHrrr:
I moved the proof over here: https://github.com/agda/agda-categories/blob/master/Categories/Category/Monoidal/Closed/IsClosed.agda

Issue #41, opened 2019-08-27
Challenge: prove that the Discrete functor is the (cartesian) right-adjoint of the forgetful functor from Cats to Sets


sstucki:
Categories currently don't carry an explicit notion of equality on objects. This seems to be a "standard" way to mechanize categories constructively (as far as I'm aware), it is in line with the [principle of equivalence](https://ncatlab.org/nlab/show/principle+of+equivalence), and it also simplifies the definition of `Category` greatly.

However, I'm worried that some rather elementary category theory silently assumes that objects really do have an equality (despite all the talk about this being evil). For example, one may define things like the hom-functor `Hom : Cᵒᵖ × C → Sets` from some category `C` to the category of sets. But in a constructive setting, we surely want to remember the notion of equality on morphisms in `C` so it's more accurate to type this as `Hom : Cᵒᵖ × C → Setoids`. (And this is indeed what seems to be implemented in `Categories.Functor.Hom`).

Another example is the adjunction between the forgetful functor from `Cats` to `Sets` and the `Discrete` functor in the opposite direction. In text-book category theory, both `Cats` and `Sets` are CCCs, so this adjunction should consist of a pair of cartesian functors. Following a [short discussion](https://twitter.com/stuckintheory/status/1162321902012108800) with @JacquesCarette on Twitter, I have tried to mechanize this IMO very elementary result, but have failed so far. Here's a summary of what I've tried so far:

 - Show that `Sets` is cartesian closed — this fails because it seems to require function extensionality (at least as `Sets` is currently defined).
 - Use `Setoids`, which is a CCC, instead of `Sets`. Now we need to pick a notion of equality on objects as we go from `Cats` to `Setoids`.
    *  Use isomorphism (don't be evil) — this isn't truly forgetful but results in a truncation because we are effectively quotienting objects by isomorphism. There is no left-adjoint because truncation forgets equality proofs of morphisms, which are required to define the counit.
    * Use propositional equality (be truly forgetful) — there is again no left-adjoint because the unit would have to (propositionally) identify setoid elements that are only equivalent.

There might be other options, but ultimately I think that if one wanted to mechanize a "classic" text-book on category theory, one would be forced to use a more "evil" notion of constructive categories, such as [Palmgren's H-categories](https://arxiv.org/pdf/1708.01924.pdf). I don't think that's too bad though. I see no reason that the two notions of categories could coexist in the same library, in principle.

JacquesCarette:
[...]

PR #42, opened 2019-09-04
New functor equivalence

sstucki:
While working with some strict constructions, I felt that the current version of (strict) functor equality is a bit hard to work with.

This PR proposes a new (but equivalent) version of functor equality, which is based on the intuition that functor equality is really a special case of natural isomorphism: the natural iso where each component is an identity.

To make this precise, the PR introduces a notion of 'heterogeneous identity', which is an identity morphism whose domain and codomain are propositionally equal (but not necessarily syntically equal) – see `Categories.Morphism.HeterogeneousIdentity`. My hope is that this construction is also useful in other circumstances.

The new definition is equivalent to the old one, and includes some conversion functions for backwards compatibility.

The only place where migration was hairy is in the module `Categories.Category.Construction.Graphs` where graph equality is defined similarly as the old functor equality. Hence there is now some mismatch. I mostly fell back to the compatibility functions to keep changes to a minimum, but this should be refactored properly if this PR is accepted.


JacquesCarette:
I had, on purpose, avoided all uses of Heterogeneously equality in this library.  So there's a bit of a hump to get over to convince me that this is a good idea.

Having said that, this code is quite clean (thanks!) and doesn't seem quite so foreign. And for those parts of the library where we're trying to do Strict category theory, it's probably the 'right' way to go. I just want to be as careful as possible to isolate the Strict parts away from the rest.

In particular, I'd be even happier with a larger refactor that would make an even cleaner separation between Strict and 'Weak' reasoning. I'm not against Strict, though I admit I have personally found working in that realm quite unpleasant (my guess is that it would be different in Cubical).

In other words, I'm going to ponder this for a bit of time, to figure out if there's an alternate design that would be even better. I don't really mean changing your code much at all, merely its organization, relative to the rest of the library.

sstucki:
> I had, on purpose, avoided all uses of Heterogeneously equality in this library. So there's a bit of a hump to get over to convince me that this is a good idea.

Yes, I am aware of that (I read the README :wink:).

Luckily, there are no heterogeneous equalities involved here. The name 'heterogeneous identity' may be a bit misleading, it's just a usual identity, with the codomain transported along a (homogeneous) equality `p`:

```agda
hid : ∀ {A B} (p : A ≡ B) → A ⇒ B
hid {A} p = subst (A ⇒_) p id
```

I call this 'heterogeneous' identity to distinguish it from the usual `id`s which – by definition – have (definitionally) identical domains and codomains.

> I'm not against Strict, though I admit I have personally found working in that realm quite unpleasant (my guess is that it would be different in Cubical).

Yes, I'm sympathetic to this. However, my impression is that it's hard to avoid 'strict' constructions if one is to follow the usual CT literature. One example is what I describe in issue #41. Another are classic categorical models of dependent type theory (pre-HoTT and cubical), which I'm particularly interested in. It seems that the usual syntax of DTT is essentially strict, which makes working with non-evil models at least cumbersome (see [this discussion in nLab](https://ncatlab.org/nlab/show/categorical+model+of+dependent+types), and [this paper by Clairambault and Dybier](https://arxiv.org/abs/1112.3456)).

JacquesCarette:
I've got tons more to say, but regarding strictness of DTT, the paper [A type theory for cartesian closed bicategories](https://arxiv.org/abs/1904.06538) by Fiore and Saville seems to remove that need. The usual syntax is indeed essentially strict, but that may turn out to be a historical over-constraint.

sstucki:
Very interesting, I wasn't aware of this work. Thanks for the pointer!

sstucki:
> In particular, I'd be even happier with a larger refactor that would make an even cleaner separation between Strict and 'Weak' reasoning.

That sounds like a good idea. Let me know if I can help somehow.

In the meantime, I'll continue to do some experiments with strict structures in my own fork (when I find time). I can continue to submit PRs if you think it's useful, or I can wait until you have decided how best to reorganize the library.

JacquesCarette:
Do continue with PRs!  The smaller the PRs, the better, in my mind. It's much easier to integrate small incremental changes. The larger things may sit for a while, especially when they involve 'new' design. Even though they might contain sub-parts that could be integrated right away...

Categories/Category/Construction/Graphs.agda
```diff
@@ -199,6 +200,12 @@ Graphs o ℓ e = record
         open GraphMorphism≈
         module EqR′ {G : Graph o ℓ e} {A B : Graph.Obj G} = EqR (Graph.setoid G {A} {B})
 
+-- FIXME: for easy migration to new functor equality.  Should probably
```
JacquesCarette: Yes, please do so if you could?  I agree that this is better, so may as well remove the old one entirely.

Categories/Category/Product.agda
```diff
@@ -135,9 +135,10 @@ _※ⁿⁱ_ : ∀ {D₁ : Category o ℓ e} {D₂ : Category o′ ℓ′ e′}
 
 module _ (C₁ : Category o ℓ e) (C₂ : Category o′ ℓ′ e′) (C₃ : Category o″ ℓ″ e″) where
 
-  module C₁ = Category C₁
-  module C₂ = Category C₂
-  module C₃ = Category C₃
+  private
```
JacquesCarette: This is a nice clean up, but I prefer if orthogonal clean-ups are in their own PRs. Those will get merged in way faster.

Categories/Functor/Equivalence.agda
```diff
@@ -1,94 +1,231 @@
...
+  open _≡F_
+
+  ≡F-assoc : {F : Functor B C} {G : Functor C D} {H : Functor D E} →
+             (H ∘F G) ∘F F  ≡F  H ∘F (G ∘F F)
+  ≡F-assoc = record { eq₀ = λ _ → refl ; eq₁ = λ _ → Equiv.sym id-comm }
+
+-- For backwards compatibility with the old functor equality.  Should
+-- probably be removed eventually.
+
+module OldEquiv {C : Category o ℓ e} {D : Category o′ ℓ′ e′} where
```
JacquesCarette: Go ahead and delete this entirely, please.

sstucki:
I'll make the requested changes as soon as I find time. Thanks for the feedback!

JacquesCarette:
Any idea when that time might appear?

sstucki:
Hopefully next week or the one after.

Categories/Functor/Equivalence.agda
sstucki: Done.

Categories/Category/Product.agda
sstucki: Done: now in #71.

Categories/Category/Construction/Graphs.agda
sstucki: Done.

sstucki:
(Note: this PR now depends on #70.)

As requested, the latest commit refactors `Categories.Category.Construction.Graphs` to get rid of the old functor equality and the backwards-compatibility code. With this, the PR should be ready for merging.

Please review @JacquesCarette.

A side note: the Graphs module could do with some refactoring (there are some comments to this effect in the code). There is a lot of stuff in this module (basic definitions, functors, an adjunction, helpers, etc.) A lot of this should go into separate modules, IMO. But this is orthogonal to the changes in the PR. I will open an separate issue about this (if there isn't one already).

PR #40, opened 2019-08-27
Two simple adjunctions between setoids, categories and groupoids.

sstucki:
Following @JacquesCarette's [suggestion](https://twitter.com/stuckintheory/status/1162321902012108800) I tried to mechanize a simple (or so I thought) adjunction. I failed to prove that the relevant pair of functors actually forms an adjunction (they probably don't), but ended up mechanizing some other stuff, which is in this PR now.

I don't think the PR should be merged as is, though.

My code probably doesn't follow the library conventions (I'm not sure what exactly these are) and I left quite a few comments in the code with questions around library design and such that I think should be removed first. My intention is to do that and update the PR if someone can guide me and help answer my questions.

Here is a summary of the questions/comments I have:
 - I'm not sure what should go into `Categories.Functor.Instance` and what should go into `Categories.Functor.Construction`.  For example, where should `Categories.Functor.Instance.Core` go?
 - I find it a bit confusing that properties and structures are sometimes named interchangeably. For example the `Groupoid` record is not really a groupoid, it's a predicate on categories that says that the category in question _is_ a groupoid. In the standard library, there are usually two distinct records for structures and properties, e.g. `Setoid` and `IsEquivalence` or `Poset` and `IsPartialOrder`.  Maybe this convention could be used here too? E.g. `Groupoid` and `IsGroupoid`?
 - Should the `Isos` category defined in `Categories.Morphism.IsoEquiv` be renamed to `Core`? Generally, I wonder whether there is some official "reference" for naming things in the library. nLab?
 - I reimplemented some existing functionality/proofs because I did not realize that it was already included in some other part of the library. In general, I removed my duplicates when I noticed the original, but I might have missed some. I at least one case, I decided to keep a reimplementation that is equivalent but different (equality of isomorphisms).

PR #46, opened 2019-09-08
The category of groupoids

sstucki:
Another slice of #40: the category of groupoids.

This is the "weak" version of said category, using natural iso as functor equality, just as `Categories.Category.Instance.Cats`. A strict version (akin to `StrictCats`) will follow in a separate PR.

~~The reason for also including the change to `Categories.Morphism.IsoEquiv` in this PR is that the proof of `F-resp-⁻¹` uses `to-unique`.~~ (This is now in #51.)

### DIFF
--- a/Categories/Category/Instance/Groupoids.agda
+++ b/Categories/Category/Instance/Groupoids.agda
@@ -1,0 +1,58 @@
+{-# OPTIONS --without-K --safe #-}
+module Categories.Category.Instance.Groupoids where
+
+-- The category of groupoids.
+--
+-- This category should maybe be called "Ho(Groupoids)" or "Ho(Gpd)"
+-- instead.  The "homsets" are not the "usual" ones consisting of
+-- functors, but consist instead of equivalence classes of functors up
+-- to natural isomorphism.  This is because homsets here are really
+-- hom-setoids and we pick natural isomorphism as the equivalence
+-- relation for these setoids.
+--
+-- See https://ncatlab.org/nlab/show/Ho%28Cat%29
+
+open import Level
+open import Categories.Category
+open import Categories.Category.Groupoid
+open import Categories.Functor as Fctr using (Functor; _∘F_)
+open import Categories.Functor.Properties using ([_]-resp-Iso)
+import Categories.Morphism.IsoEquiv as IsoEquiv
+open import Categories.NaturalTransformation.NaturalIsomorphism
+  using (NaturalIsomorphism; associator; unitorˡ; unitorʳ; isEquivalence; _ⓘₕ_)
+private
+  variable
+    o ℓ e : Level
+
+open Groupoid using (category)
+
+-- The category of groupoids.
+
+Groupoids : ∀ o ℓ e → Category (suc (o ⊔ ℓ ⊔  e)) (o ⊔ ℓ ⊔ e) (o ⊔ ℓ ⊔ e)
+Groupoids o ℓ e = record
+  { Obj       = Groupoid o ℓ e
+  ; _⇒_       = λ G H → Functor (category G) (category H)
+  ; _≈_       = NaturalIsomorphism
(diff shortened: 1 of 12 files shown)


sstucki:
This PR has been open for a while. Is there anything problematic about it?

If so, please let me know how I can improve it.

JacquesCarette:
It came in when I was super-busy, and I haven't been systematic enough to go back to look at it. Thanks for the ping, I'll look at it in detail now.

Categories/Morphism/IsoEquiv.agda
```diff
@@ -68,3 +69,59 @@ Isos = record
       }
   }
   where open Equiv
+
+----------------------------------------------------------------------
+
+-- An alternative notion of equality on isomorphisms that only
```
JacquesCarette: Could you please split this into its own PR?

Categories/Category/Instance/Groupoids.agda
```diff
@@ -0,0 +1,72 @@
...
+import Categories.Morphism.IsoEquiv as IsoEquiv
+open import Categories.NaturalTransformation.NaturalIsomorphism
+  using (NaturalIsomorphism; associator; unitorˡ; unitorʳ; isEquivalence; _ⓘₕ_)
+private
+  variable
+    o ℓ e : Level
+
+-- A groupoid is a category that has a groupoid structure
+--
+-- This is not the same as Categories.Category.Groupoid.Groupoid.
```
JacquesCarette: I agree that the current `Categories.Category.Groupoid.Groupoid` is ill-named and should be `IsGroupoid`. That should be done - and this local definition removed.

I would prefer if things were done in that order though.  But if you're ok with doing both, I'm also ok with pulling this in first and trust that it will shortly be refactored.

Categories/Morphism/IsoEquiv.agda
sstucki: Done in #51.

Categories/Category/Instance/Groupoids.agda
sstucki: Done in #55.

sstucki:
I refactored the PR further, as requested by @JacquesCarette.

It now depends on #51 and #55 (see above for details).

Categories/Category/Instance/Groupoids.agda
```diff
@@ -0,0 +1,67 @@
...
+import Categories.Morphism.IsoEquiv as IsoEquiv
+open import Categories.NaturalTransformation.NaturalIsomorphism
+  using (NaturalIsomorphism; associator; unitorˡ; unitorʳ; isEquivalence; _ⓘₕ_)
+private
+  variable
+    o ℓ e : Level
+
+-- A groupoid is a category that has a groupoid structure
+
+record Groupoid (o ℓ e : Level) : Set (suc (o ⊔ ℓ ⊔ e)) where
```
JacquesCarette: This should still be split off and put into `Categories.Category.Groupoid`.

sstucki: Hm. Should we then put `IsGroupoid` in that same module, or do you prefer to have both `Categories.Category.Groupoid` and `Categories.Category.IsGroupoid` as separate modules?

JacquesCarette: I have no strong opinions there.  I know the stdlib splits them, but for now, I would be fine with putting them all in the same module.

sstucki: OK, I put them in the same file. Less clutter.

sstucki:
In the latest version, `Groupid` and `IsGroupoid` are now both in `Categories.Category.Groupoid`, while the actual category of groupoids is in `Categories.Category.Instance.Groupoids`.

The reason so many files are changed is that the `Categories.Category.IsGroupoid` module disappeared.

PR #47, opened 2019-09-08
Core functor

sstucki:
Yet another slice of #40: the core functor from categories to groupoids.

Contains:
 - the core construction (the groupoid `Core C` for a given category `C`) in `Categories.Category.Construction.Core`;
 - the construction extended to a functor in `Categories.Functor.Instance.Core`;
 - a proof that the functor is right-adjoint to the forgetful functor from `Groupoids` to `Cats` in `Categories.Adjoint.Instance.Core`.

Caveats:
 - this PR depends on #46 for the definition of the category of groupoids;
 - the `Core` construction is essentially the same as the `Isos` category from `Categories.Morphism.IsoEquiv`, but I think the name and placement of the new module is better –  "core" is the generally accepted name of this construction, AFAIK.

### DIFF
--- a/Categories/Adjoint/Instance/Core.agda
+++ b/Categories/Adjoint/Instance/Core.agda
@@ -1,0 +1,110 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Adjoint.Instance.Core where
+
+-- The adjunction between the forgetful functor from Cats to Groupoids
+-- and the Core functor.
+
+open import Level using (_⊔_)
+import Function
+
+open import Categories.Adjoint using (_⊣_)
+open import Categories.Category using (Category)
+import Categories.Category.Construction.Core as C
+open import Categories.Category.Groupoid using (Groupoid)
+open import Categories.Category.Instance.Cats using (Cats)
+open import Categories.Category.Instance.Groupoids using (Groupoids)
+open import Categories.Functor using (Functor; _∘F_; id)
+open import Categories.Functor.Instance.Core using (Core)
+import Categories.Morphism as Morphism
+open import Categories.NaturalTransformation.NaturalIsomorphism using (refl; _≃_)
+
+-- The forgetful functor from Groupoids to Cats
+
+Forgetful : ∀ {o ℓ e} → Functor (Groupoids o ℓ e) (Cats o ℓ e)
+Forgetful = record
+  { F₀ = category
+  ; F₁ = Function.id
+  ; identity     = refl
+  ; homomorphism = refl
+  ; F-resp-≈     = Function.id
+  }
+  where open Groupoid
+
+-- Core is right-adjoint to the forgetful functor from Groupoids to
+-- Cats
+
+CoreAdj : ∀ {o ℓ e} → Forgetful {o} {ℓ ⊔ e} {e} ⊣ Core
+CoreAdj = record
+  { unit   = record { η = unit   ; commute = λ {G} {H} F → unit-commute {G} {H} F }
+  ; counit = record { η = counit ; commute = counit-commute }
+  ; zig    = λ {G} → zig {G}
+  ; zag    = zag
+  }
+  where
+    open Groupoid using (category)
+    module Core = Functor Core
+
+    unit : ∀ G → Functor (category G) (C.Core′ (category G))
+    unit G = record
+      { F₀ = Function.id
+      ; F₁ = λ f → record { from = f ; to = f ⁻¹ ; iso = iso }
+      ; identity     = Equiv.refl
+      ; homomorphism = Equiv.refl
+      ; F-resp-≈     = Function.id
+      }
+      where open Groupoid G
+
+    unit-commute : ∀ {G H} (F : Functor (category G) (category H)) →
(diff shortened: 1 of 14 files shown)


JacquesCarette:
This is great.  Modulo comments on PRs on which this 'depends', I'll pull this in as soon as the pieces are ready.

sstucki:
As requested, I'll further refactor this before it can be merged.

JacquesCarette:
So this one is now obsolete and should be closed, is that right?

sstucki:
No, #53 only addressed parts of this PR. The functor and adjunction are missing.

I am currently rebasing this on #46 and will push the updated version once it's ready.

JacquesCarette:
:+1:

sstucki:
OK, this PR is now ready, but it depends on #46, so let's wait until that one has been merged.

Also: there are now _two_ definitions of the core category in `Categories.Category.Construction.Core`:
- `Core` is what was previously called `Isos`;
- `Core′` is what was previously called `Core`.

They are identical, except the latter uses the simpler equality of isos (from #51).

The reason I kept both is that there is some code in `Categories.Morphism.Isomorphism` that relies on the old definition (it uses the more complex equality) while the implementation of the core functor and the adjunction use the new one (they were easier to write with the simplified equality). 

Having the two definitions like this is probably not ideal. I see three ways forward, but I don't know which you prefer:

1. Keep both definitions and prove they are equivalent as categories (in a separate module, e.g. `Categories.Category.Construction.Core.Equivalence`). One could then use whichever definition one finds easier to work with for the task at hand.

2. Remove the new definition and insert "coercions" everywhere in the definitions of the core functor and adjunction to convert the simple equality proofs into complex ones. The necessary conversion functions are already in place, but readability might suffer.

3. Remove the old definition and rework `Categories.Morphism.Isomorphism` (which probably needs to be reworked anyways as it uses`rewrite` repeatedly). This is my preferred option.

Whichever option is chosen, I suggest creating an explicit issue for this and solving it in a separate PR. Splitting and rebasing these PRs is error-prone and tedious 😓 .

JacquesCarette:
I'm happy with trying approach number 3.  And I'm also fine with the small amount of 'duplication' involved here (so I'll merge now). 

@HuStmpHrrr is the one who ported these files, so I'd really like to hear his opinion too regarding `Core` vs `Core'` and the different definitions involved there.

sstucki:
> @HuStmpHrrr is the one who ported these files, so I'd really like to hear his opinion too regarding `Core` vs `Core'` and the different definitions involved there.

OK, I'll wait for a comment from @HuStmpHrrr before making any changes.

HuStmpHrrr:
I think `Core′` looks a better idea. it just reduces the burden of an extra proof.

HuStmpHrrr:
I am not even sure the previous version is still needed provided the new definition is simply better. please refactor as you wish.

sstucki:
I gave this a try today and noticed that the simplified equality on isos does not play well with implicit arguments. Concretely, in functions that have signatures of the form
```agda
fun : {f : A ≅ B} ... → (eq : f ≃ ...)
```
The parameter `f` can often not be inferred because `eq` only involves the `from` part.

As a result the functions at the end of `Categories.Morphism.Isomorphism` that implement some "diagram chasing" all require quite a lot of explicit passing of implicit arguments, both in their proof but also in their application (e.g. in `Categories.Category.Monoidal.Properties`).

I'm not sure how best to address this. Some of these functions could maybe be generalized but that would lead to even more refactoring in modules that depend on them. I'm not sure if that's desirable. Then again, it seems that most of them are only ever used in `Categories.Category.Monoidal.Properties`.

@HuStmpHrrr WDYT?

sstucki:
It looks like these functions are only ever used in `Categories.Category.Monoidal.Properties` and are quite specific to Kelly's proofs. Maybe they should just be moved there?

HuStmpHrrr:
oh that's bad. the module actually is quite generic: diagrams are easier to chase if all morphisms provided are isomorphisms. the idea in coherence conditions of monoidal categories is that it's much easier to chase isomorphisms instead of just morphisms. so the module has nothing specific to the proofs itself. there is just one use case, i believe, because this library just does not contain enough conclusions.

but what puzzles me is then why the lifting functions have no problem, as they also just talk about one direction?
```agda
lift-triangle′ : {f : A ≅ B} {g : C ≅ A} {h : C ≅ B} → from f ∘ from g ≈ from h → f ∘ᵢ g ≃ h
```

sstucki:
> but what puzzles me is then why the lifting functions have no problem, as they also just talk about one direction?

You do have a problem when trying to apply them, but proving them is trivial (equality on  isos is already equality on the underlying morphisms).

sstucki:
I think I found a nice compromise.
 - #56 does the minimal refactoring to remove uses of `rewrite` and some cleanup.
 - #57 redefines `_≃_` to only contain `from-≈` whereas `to-≈` becomes a derived field.

The second PR combines the advantages of the new and old iso equality: Agda can still infer which isos are involved, but the overhead is minimal. Essentially, just wrapping proofs using the `⌞_⌟` constructor.

Let me know what you think.

HuStmpHrrr:
yes that looks much better. good that Jacques has merged it.

PR #48, opened 2019-09-17
Fix #44: the full Grothendieck construction for (contravariant) pseudofunctors

### LINKED ISSUE #44: is the equivalence of Grothendieck construction correct?
the equivalence of Grothendieck construction looks strange to me.

consider the following diagram:

IMG_20190907_164246

the current definition only has `f` and `g` equivalent, but I suppose the triangle in category `FB` does not necessarily commute, does it?

sstucki:
Now (hopefully) with the correct notion of equality on arrows.

Differences w.r.t. the old version:
 * The fixed version turned out _a lot_ more complicated. 
 * This version uses pseudofunctors (rather than plain functors). The associativity and unit laws require coherence conditions which plain functors don't provide. I believe it should be possible to instead give a construction using plain functors into `StrictCats` (instead of `Cats`), but the version using pseudofunctors is more general.
 * Interestingly, because of the way the coherence conditions are defined in `Pseudofuctor`, it was easier to prove the contravariant version of the construction.

### DIFF
--- a/Categories/Bicategory/Construction/1-Category.agda
+++ b/Categories/Bicategory/Construction/1-Category.agda
@@ -1,0 +1,126 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Categories.Category
+
+-- Lifts a 1-Category into a bicategory
+
+module Categories.Bicategory.Construction.1-Category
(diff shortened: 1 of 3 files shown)


JacquesCarette:
Superb!  Thanks a lot.

PR #51, opened 2019-09-22
Alternative (but equivalent) equivalence for isomorphisms.

sstucki:
Split out of #46.

UPDATE: Now depends on #55.


sstucki:
Please merge #55 first.

Even though there are not really any dependencies, I rebased this PR on top of #55 to avoid merge conflicts in later PRs that will depend on both #55 and this PR.

PR #53, opened 2019-09-22
Rename `Isos` to `Core` and move it into its own module

sstucki:
Depends on #52


sstucki:
Sorry for force-pushing changes. I missed a dependency.

On that note: compiling `Everything` fails with the error

    Categories/Category/Construction/Graphs.agda:29,8-86
    Failed to find source of module
    Relation.Binary.Construct.Closure.ReflexiveTransitive.Equivalence

PR #66, opened 2019-10-22
added some extra reasoning tools and proved 2 unitor identities

MirceaS:


Categories/Category/Monoidal.agda
```diff
@@ -207,6 +214,7 @@ record Monoidal : Set (o ⊔ ℓ ⊔ e) where
   refl⊗refl≃refl : ≅.refl {A} ⊗ᵢ ≅.refl {B} ≃ ≅.refl
   refl⊗refl≃refl = ⌞ identity ⌟
 
+
```
MirceaS: newline added by mistake

Categories/Category/Core.agda
```diff
@@ -47,6 +47,9 @@ record Category (o ℓ e : Level) : Set (suc (o ⊔ ℓ ⊔ e)) where
   ∘-resp-≈ʳ : ∀ {A B C} {f h : A ⇒ B} {g : B ⇒ C} → f ≈ h → g ∘ f ≈ g ∘ h
   ∘-resp-≈ʳ pf = ∘-resp-≈ refl pf
 
+  assoc² : ∀ {A B C D E} {f : A ⇒ B} {g : B ⇒ C} {h : C ⇒ D} {i : D ⇒ E} → ((i ∘ h) ∘ g) ∘ f ≈ i ∘ (h ∘ (g ∘ f))
+  assoc² = trans assoc assoc
```
JacquesCarette: This definitely shouldn't go into the `Category` record.  There should be a module (whose current home would be under `Morphism.Reasoning`) that has a bunch of 'associativity combinators' like this one.

MirceaS: Moved it.

PR #67, opened 2019-10-23
Proved part of the braiding coherence lB=r

MirceaS:
Proved the -⊗1 version of the coherence theorem.
I might get around to proving the final identity later but for now I thought it would be useful to have this as a PR and other people can do it in the meantime if they want.
Also, I wasn't sure where exactly to write it so I created a new module inside Braided.agda. Feel free to move stuff around if that is not appropriate.

### DIFF
--- a/Categories/Category/Monoidal/Braided.agda
+++ b/Categories/Category/Monoidal/Braided.agda
@@ -2,6 +2,8 @@
 
 open import Categories.Category
 open import Categories.Category.Monoidal
+import Categories.Category.Monoidal.Properties as MProperties
+import Categories.Morphism.Reasoning as Reasoning
 
 module Categories.Category.Monoidal.Braided {o ℓ e} {C : Category o ℓ e} (M : Monoidal C) where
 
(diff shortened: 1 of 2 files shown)


MirceaS:
TODO: Open the module publicly exposing the final identity once it's proved

Categories/Category/Monoidal/Braided.agda
```diff
@@ -51,3 +57,38 @@ record Braided : Set (levelOfTerm M) where
                  B                           ⇒⟨ Z ⊗₀ X ⊗₀ Y ⟩
                  associator.to
                ⟩
+
+  module Properties where
```
HuStmpHrrr: I would define this in a separate module in a new file.

Categories/Morphism/Reasoning/Core.agda
```diff
@@ -41,9 +41,12 @@ module Basic where
 open Basic public
 
 module Utils where
-  assoc² : ((i ∘ h) ∘ g) ∘ f ≈ i ∘ (h ∘ (g ∘ f))
+  assoc² : ((i ∘ h) ∘ g) ∘ f ≈ i ∘ h ∘ g ∘ f
```
sstucki: I'm not sure this should be changed. I actually like the more explicit version with parentheses better.

PR #69, opened 2019-10-27
Add 'heterogeneous' identities.

sstucki:
(This is a subset of #42, refactored into its own PR.) 

If Agda was an extensional type theory, any pair of morphisms `f : A ⇒ B` and `g : A ⇒ C` with `A ≡ B` would belong to the same homset, even if `A` and `B` are not definitionally equal. In particular, the identity on `B` could be given the type `id {B} : B ⇒ C`.

But Agda is an intensional type theory, so the identity cannot have this type, in general. Instead one needs to manually "transport" `id {B}` into the homset `B ⇒ C`. Given `p : B ≡ C`,
```agda
  idBC : B ⇒ C
  idBC = subst (B ⇒_) p (id {B})
```
Morphisms like `idBC` are no longer identities (in the strict sense) but they still enjoy many of the properties identities do.

To make this precise, this PR introduces a notion of 'heterogeneous identity', which is an identity morphism whose domain and codomain are propositionally equal (but not necessarily syntically equal).

For now, heterogeneous identities are used only to implement an improved functor equality (se PR #42), but my hope is that this construction is also useful in other circumstances.

**A note on naming.** The name "heterogeneous identity" suggests that some form of heterogeneous equality is at play here. There isn't. This is simply the transport of identity morphisms along homogeneous equalities. I realize, however, that this name might not sit easy with people because of the supposed association with heterogeneous equalities. I'm happy to change the name if we can agree on a better term. Two alternatives I have considered are "quasi-identities" and "transported identities".

### DIFF
--- a/Categories/Morphism/HeterogeneousIdentity.agda
+++ b/Categories/Morphism/HeterogeneousIdentity.agda
@@ -1,0 +1,80 @@
+{-# OPTIONS --without-K --safe #-}
+open import Categories.Category using (Category)
+
+-- 'Heterogeneous' identity morphism and some laws about them.
+
+module Categories.Morphism.HeterogeneousIdentity {o ℓ e} (C : Category o ℓ e) where
+
+open import Level
+open import Relation.Binary.PropositionalEquality
+
+import Categories.Morphism as Morphism
+
+open Category C
+open Morphism C
+
+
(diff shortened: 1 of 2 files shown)


JacquesCarette:
Great! Merging.

Issue #61, opened 2019-10-15
Cats is CartesianClosed (when all levels the same)


JacquesCarette:
I tried to prove this, got a fair ways in, and got stuck.  I don't think it is fundamental, just that I'm missing something that will likely turn out to be simple.  Also, my definitions might be off a bit, which certainly wouldn't help matters. 
I've pushed my attempts onto branch 'CatsCCC' (as I can't attach .agda files easily here). Holes 3 and 5 in particular.  I think the rest are ok, just tedious.

JacquesCarette:
So I think I know what the source of difficulty is: the definition of 'exponential object' used assumes that some products (as necessary) exist; but we already know all products exist. The problem is that the definition doesn't use the product just given.  But of course the it is CCC is with respect to that one product.  [This definition is inherited from the original port].  I think that if the definition was adjusted so that the exponential object was coherent, by construction, with the given product, lots of things would be considerably simpler.  And it would match the definitions in the literature too.

HuStmpHrrr:
actually the current behavior is correct.

consider a trivial example. usually we say a product of `A` and `B` is `A × B`. However, it doesn't have to be the *canonical* representation of the product. for example, `B × A` is also an legitimate representation. in fact, there isn't a preference between both; they are just equally good, and there are other equally good representations. the uniqueness part of products concludes all such best representations are isomorphic and that's good enough. that the exponential part does not make assumption of concrete representation of a product is consistent with this extensional view.

JacquesCarette:
The current behaviour is definitely not wrong. I agree with your reasoning.

I wonder if anyone has ever proved it in full generality though?  I really have no idea of how to proceed in those 2 holes. I suspect that existing proofs are either just sketches or 'for convenience' pick one product, since they are all "the same" anyways.

HuStmpHrrr:
I did it in https://github.com/agda/agda-categories/commit/1c7e2f44bb31e7f502b26d256e81cdc8f62ad5e3

it is not too difficult. the problem is that there are slow computation parts. the idea is to make use of the UMP of the products in the construction. assuming a canonical construction is morally correct, because if we only inspect objects and morphisms extensionally consistently, there shouldn't be a problem. this is seen in many other constructions where properties like associativity etc are ignored.

JacquesCarette:
[...]

PR #90, opened 2019-12-09
Refactor proof that Cats is a CCC to make its parts more reusable.

sstucki:
I decided to develop @HuStmpHrrr's insight from #61 a bit further.

**Recap.** The current definition of CCCs is very generic, unlike the 'typical' presentations where exponentials are defined in terms of a canonical choice of products. Thanks to the universal property of products, this does not matter. One can just repack products to work with whatever isomorphic presentation one desires.

**This PR.** This is of course not specific to `Cats`. Indeed, it is completely independent from the actual CCC one is working with, so the work of converting from/to the canonical products can be done once and for all, in a generic fashion. To this end, I added the module `Categories.Category.CartesianClosed.Canonical` which is closer to the 'typical' presentation of CCCs. It completely unfolds the definition of CCCs and defines exponentials w.r.t. to the canonical choice of products. Since this presentation is (provably) equivalent to the current one, we may prove that an instance (e.g. `Cats`) is 'canonically cartesian closed', and then convert the proof to the usual definition.

**Future work.** There are at least four equivalent definitions of CCCs that I can think of.
 1. the current, generic one (in terms of exponentials),
 2. the new, canonical one,
 3. one requiring a category to be a) cartesian, and b) monoidal-closed w.r.t. products,
 4. one in terms of an adjoint tripple of functors (diagonal, product, hom).

It would be nice to implement 3 and 4 as well, and prove them equivalent to the others.

### DIFF
--- a/Categories/Category/CartesianClosed/Canonical.agda
+++ b/Categories/Category/CartesianClosed/Canonical.agda
@@ -1,0 +1,161 @@
+{-# OPTIONS --without-K --safe #-}
+open import Categories.Category using (Category)
+
+-- A "canonical" presentation of cartesian closed categories.
+--
+-- This presentation is equivalent to the one in
+-- Categories.Category.CartesianClosed but it is easier to work with
+-- in some circumstances.
+--
+-- Here, exponentials are not defined in terms of arbitrary products,
+-- but in terms of a family of "canonical" products.  Since products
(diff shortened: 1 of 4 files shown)


HuStmpHrrr:
yeah this looks very good. thx.

PR #96, opened 2020-01-03
Refactor and complete enriched natural transformations.

sstucki:
This fixes some of @JacquesCarette's [comments on 4293cf22](https://github.com/sstucki/agda-categories/commit/4293cf22dd2a5b87a62e4ad9aa8acfb4a173a646#commitcomment-36642277), but not all. Most importantly, the following remain open:

> > 1. why `comp`?  Why not `η`?
> 
> Same reason as for functors: `η` should be the name of the natural transformation, not that of its first projection. (And also, `η` is used for so many other things). I picked `comp` for "component" but it's definitely not a great name. I'm open to other suggestions.
> 
> > 3. I am extremely dubious about the notation `_[_]` for NT. I fear it will not play well with other notation.  Why bother anyways?  `η X` usually looks fine.
> 
> I'm open to other suggestions. I was contemplating `_$_`, but thought it might be less readable, and `_⟨_⟩`, but that's harder to type (and could interfere with other things as well). As explained earlier, I'm not a fan of things like `η α X` and `α.η X`.

I'm not quite sure what to do with these. Rename or not? If yes, what names should be used instead? Those used for ordinary natural transformations? Or are there better ones?

### DIFF
--- a/Categories/Enriched/NaturalTransformation.agda
+++ b/Categories/Enriched/NaturalTransformation.agda
@@ -1,51 +1,226 @@
 {-# OPTIONS --without-K --safe #-}
 
-open import Level
-open import Categories.Category
+open import Categories.Category using () renaming (Category to Setoid-Category)
 open import Categories.Category.Monoidal
 
-module Categories.Enriched.NaturalTransformation {o ℓ e : Level} {V : Category o ℓ e} (M : Monoidal V) where
+module Categories.Enriched.NaturalTransformation
+  {o ℓ e} {V : Setoid-Category o ℓ e} (M : Monoidal V) where
 
-open import Categories.Enriched.Category renaming (Category to Enriched) hiding (_[_,_])
-open import Categories.Enriched.Functor M hiding (id)
+open import Level
 
-private
-  variable
-    v v′ : Level
-    C D E : Enriched M v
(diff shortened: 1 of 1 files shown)


JacquesCarette:
I've looked around and could not find any alternate notation for natural transformations - everyone seems to use "subscript the name" as the projector for the component in paper-math.

I then also did a survey of formalizations - most of them are hideous, in the sense that there is very little attention paid to notation, so that the results end up looking very far from math.

For all the record-based versions ou tthere, I have seen the following extractors:
- `η`
- `α`
- components_of
- nt_component
- ApNT
- transform (with notation transform[_] for extraction)
- natural_map (with a coercion added so that the name of the NT can be used as a function, which is nice but leads to coercion hell)
- indmor
- NatTransMap (with notation _$$_ for application)

So `comp` doesn't stand out at all from the crowd...  And the _[_] notation kind of has a precedent too.  Thus I'm thinking of trying this out as is.

sstucki:
Wow, thanks for that survey!

I contemplated the full `component` (rather than the abbreviated `comp`) but seemed a bit too verbose.

I didn't think of `transform`. That wouldn't be too bad either.

Let's discuss it further in a separate issue (which I will create soon, I promise!)

PR #97, opened 2020-01-03
Enriched functor categories.

sstucki:
Depends on PR #96.

Addresses @JacquesCarette's [comment on b72b3ce](https://github.com/sstucki/agda-categories/commit/b72b3ce8f836c94120edd532c4855f8e77307668#commitcomment-36642345).


JacquesCarette:
Interesting how `EnrichedFunctors` (as a 'normal' category) doesn't care at all what the level of the objects of the enriching category is.

sstucki:
I agree! That's because enriched functors don't either. All their fields are morphisms or equalities in `V`, no objects!

PR #107, opened 2020-01-06
The bicategory of V-enriched categories.

sstucki:
This completes the mechanization of Section 1.3 of [Kelly's book on enriched category theory](http://www.tac.mta.ca/tac/reprints/articles/10/tr10abs.html).

Type checking this module on my MacBook takes a ridiculous amounts of time (just over a 1min when I give it 6GM of memory).

I'm not sure what the problem is, though I suspect the culprits are the proofs of `⊚-assoc`, `unitˡ`, `unitʳ` and the coherence conditions, or rather, checking their types against the type signatures of the corresponding fields in the `Bicategory` record. Type checking time actually *went up* when I tried to first define functions corresponding to these fields separately and then use them to populate the record.

If someone has ideas on how to improve the performance, that would be great.

### DIFF
--- a/Categories/Bicategory/Instance/EnrichedCats.agda
+++ b/Categories/Bicategory/Instance/EnrichedCats.agda
@@ -1,0 +1,131 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Level
+open import Categories.Category using () renaming (Category to Setoid-Category)
+open import Categories.Category.Monoidal using (Monoidal)
+
+module Categories.Bicategory.Instance.EnrichedCats
+  {o ℓ e} {V : Setoid-Category o ℓ e} (M : Monoidal V) (v : Level) where
+
+-- The 2-category of V-enriched categories
+
+open import Data.Product as Prod using (_,_)
+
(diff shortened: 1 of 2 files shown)


JacquesCarette:
I think that 'combinator style' is slower than equational style.  I think it ends up generating a much larger system of equations to unify, some of which only appear "in the middle" and take more work to figure out. [I could, of course, be quite wrong; but I definitely had problems in other code with things that evaluated too much, and thus caused "in the middle" variables to disappear.]

sstucki:
I took your advice and switched from combinator style to equational style. Together with some other tweaks, I managed to reduce compile time by 50%. See #110 for details.

JacquesCarette:
So do you think it was due (mainly?) to the style switch?  That would be good to know!

sstucki:
I suspect that the lion share of the improvement stems from the use of `niHelper` since that reduces the number of proof terms to be checked for the associator and unitors a lot. But that cannot account for 100% of the speedup since the proofs of the coherence conditions does not involve `niHelper`. The rest might well be due to the style switch.

PR #108, opened 2020-01-06
Use 2-Category in the definition of 2-Functor...

sstucki:
(… instead of `Enriched.Category`.)

This is a trivial change. I'm not sure whether I should submit such things as PRs or just push to the main repo?


sstucki:
OK, I'll continue with the PRs then.

PR #110, opened 2020-01-09
Optimize proofs of EnrichedCats

sstucki:
I cleaned up and optimized the proofs in `EnrichedFunctors` and `EnrichedCats`.

With these optimizations, the compile time for the `EnrichedCats` module roughly halved on my MacBook (from ~1min to ~30s).

(~~Depends on #109.~~ Rebased on master; no further dependencies)

### DIFF
--- a/Categories/Bicategory/Instance/EnrichedCats.agda
+++ b/Categories/Bicategory/Instance/EnrichedCats.agda
@@ -13,19 +13,34 @@ open import Data.Product as Prod using (_,_)
 
 open import Categories.Bicategory using (Bicategory)
 open import Categories.Category.Construction.EnrichedFunctors M
-open import Categories.Category.Monoidal.Reasoning M
 open import Categories.Enriched.Category M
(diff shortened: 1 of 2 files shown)


JacquesCarette:
Hmm, I almost wished you'd done this in more stages, so that we could see what effect each individual optimization had.  Of course, it's also entirely possible that the various optimizations 'work together' to produce the results.  The main message here is, in the future, to try to isolate things even more [smaller steps], so that we can better understand what works & we can apply these same ideas elsewhere.

sstucki:
Indeed.

My initial plan was to do it in stages, but converting the proof terms was excruciating! 😖 Filling a single hole -- or displaying its normalized type -- took anywhere between 20s and several minutes.  Sometimes Agda wouldn't seem to terminate at all, and I had to kill and restart it.

With this kind of timing, gradual refactoring would have taken me hours (it still took about 3h overall), and when I was working on this it was already past midnight. So after converting 3 proofs, I gave up, introduced `niHelper` and did the rest in one go.

PR #111, opened 2020-01-10
The forgetful functor from enriched to ordinary categories

sstucki:
(Depends on #110.)

The "forgetful" functor from V-enriched categories to their underlying Setoid-categories.

This roughly corresponds to Section 1.3 of [Kelly's book](http://www.tac.mta.ca/tac/reprints/articles/10/tr10abs.html).

It's another compile-time-eating monster...


sstucki:
Ah, yes, you're right. That seems easy to fix, I'll try to do so some time during the weekend. (Unless you wanna go ahead and do it yourself.)

sstucki:
> I'll merge now, but I think that id-comm and id-comm-sym could be used to shorten many of the proofs (a little).

Addressed in #113.

PR #122, opened 2020-02-20
Implement Lawveres fixed point theorem

TOTBWF:
This PR implements [Lawvere's Fixed Point Theorem](https://ncatlab.org/nlab/show/Lawvere%27s+fixed+point+theorem).

On the organizational front, I'm not sure that this is the best place to put the theorem, but I couldn't think of anywhere better. I'm open to suggestions!

### DIFF
--- a/Categories/Category/CartesianClosed/Properties.agda
+++ b/Categories/Category/CartesianClosed/Properties.agda
@@ -13,11 +13,11 @@ module _ {o ℓ e} {𝒞 : Category o ℓ e} (𝓥 : CartesianClosed 𝒞) where
   open CartesianClosed 𝓥
   open HomReasoning
 
-  point-surjective : ∀ {A X Y : Obj} → (X ⇒ Y ^ A) → Set (ℓ ⊔ e)
-  point-surjective {A = A} {X = X} {Y = Y} ϕ =
+  PointSurjective : ∀ {A X Y : Obj} → (X ⇒ Y ^ A) → Set (ℓ ⊔ e)
+  PointSurjective {A = A} {X = X} {Y = Y} ϕ =
     ∀ (f : A ⇒ Y) → Σ[ x ∈ ⊤ ⇒ X ] (∀ (a : ⊤ ⇒ A) → eval′ ∘ first ϕ ∘ ⟨ x , a ⟩  ≈ f ∘ a)
 
-  lawvere-fixed-point : ∀ {A B : Obj} → (ϕ : A ⇒ B ^ A) → point-surjective ϕ → (f : B ⇒ B) → Σ[ s ∈ ⊤ ⇒ B ] f ∘ s ≈ s
+  lawvere-fixed-point : ∀ {A B : Obj} → (ϕ : A ⇒ B ^ A) → PointSurjective ϕ → (f : B ⇒ B) → Σ[ s ∈ ⊤ ⇒ B ] f ∘ s ≈ s
   lawvere-fixed-point {A = A} {B = B} ϕ surjective f = g ∘ x , g-fixed-point
     where
       g : A ⇒ B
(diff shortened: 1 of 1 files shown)


HuStmpHrrr:
thx for the PR. we usually use upper case for predicate, so `point-surjective` would become `PointSurjective` but otherwise I don't have other problems. probably @JacquesCarette would also want to take a look?

Categories/Category/CartesianClosed/Properties.agda
```diff
@@ -0,0 +1,45 @@
...
+      x : ⊤ ⇒ A
+      x = proj₁ (surjective  g)
+
+      g-surjective : eval′ ∘ first ϕ ∘ ⟨ x , x ⟩ ≈ g ∘ x
+      g-surjective = proj₂ (surjective g) x
+
+      lemma : ∀ {A B C D} → (f : B ⇒ C) → (g : B ⇒ D) → (h : A ⇒ B) → (f ⁂ g) ∘ ⟨ h , h ⟩ ≈ ⟨ f , g ⟩ ∘ h
+      lemma f g h = begin
+        (f ⁂ g) ∘ ⟨ h , h ⟩ ≈⟨  ⁂∘⟨⟩ ⟩
+        ⟨ f ∘ h , g ∘ h ⟩ ≈⟨ sym ⟨⟩∘ ⟩
```
JacquesCarette: You should use `≈˘` instead of an explicit call to `sym`.

Categories/Category/CartesianClosed/Properties.agda
```diff
@@ -0,0 +1,45 @@
...
+        ⟨ f , g ⟩ ∘ h ∎
+
+      g-fixed-point : f ∘ (g ∘ x) ≈ g ∘ x
+      g-fixed-point = begin
+        f ∘ g ∘ x ≈⟨  refl⟩∘⟨ sym g-surjective ⟩
+        f ∘ eval′ ∘ first ϕ ∘ ⟨ x , x ⟩  ≈⟨ refl⟩∘⟨ refl⟩∘⟨ lemma ϕ id x ⟩
+        f ∘ eval′ ∘ ⟨ ϕ , id ⟩ ∘ x ≈⟨ sym assoc ○ sym assoc ○ ∘-resp-≈ˡ assoc ⟩
+        (f ∘ eval′ ∘ ⟨ ϕ , id ⟩) ∘ x ≈⟨  refl ⟩∘⟨refl ⟩
```
JacquesCarette: I think this is just `refl`.

Categories/Category/CartesianClosed/Properties.agda
```diff
@@ -0,0 +1,45 @@
...
+      lemma f g h = begin
+        (f ⁂ g) ∘ ⟨ h , h ⟩ ≈⟨  ⁂∘⟨⟩ ⟩
+        ⟨ f ∘ h , g ∘ h ⟩ ≈⟨ sym ⟨⟩∘ ⟩
+        ⟨ f , g ⟩ ∘ h ∎
+
+      g-fixed-point : f ∘ (g ∘ x) ≈ g ∘ x
+      g-fixed-point = begin
+        f ∘ g ∘ x ≈⟨  refl⟩∘⟨ sym g-surjective ⟩
+        f ∘ eval′ ∘ first ϕ ∘ ⟨ x , x ⟩  ≈⟨ refl⟩∘⟨ refl⟩∘⟨ lemma ϕ id x ⟩
+        f ∘ eval′ ∘ ⟨ ϕ , id ⟩ ∘ x ≈⟨ sym assoc ○ sym assoc ○ ∘-resp-≈ˡ assoc ⟩
```
JacquesCarette: you can probably use a combination of `≈˘` and `sym-assoc` to shorten this, I think to
```agda
≈˘⟨ assoc ○ assoc ○ ∘-resp-≈ˡ sym-assoc ⟩
```
(or likely in the opposite order).

Categories/Category/CartesianClosed/Properties.agda
TOTBWF: Thanks, not sure how I missed that.

Issue #103, opened 2020-01-05
Rename record fields for functorial actions and components of natural transformations


sstucki:
#### TL;DR

I really, *really*, don't like the names of the fields `F₀`, `F₁` and `η` in the `Functor` and `NaturalTransformation` records.

Below, I explain why and suggest some alternatives. However, I realize that changing these names would affect just about every file in the library and completely break compatibility with older versions. So I think any such change should be thoroughly discussed first. That is the purpose of this issue.

#### Why I don't like the current names

In my opinion it's not a good idea to conflate the name of a projection (field name) with the typical name of an instance of the record it belongs to. There two reasons for this:

 1. Functors are typically called `F`, and their functorial actions `F₀` and `F₁`, so as long as there's only ever one functor, it makes sense to use this convention for the fields of `Functor`. Then we simply open `Functor F` and everything looks just like it is supposed to.

    But only until there are two or more functors. Then one gets unreadable code of the form `F₀ G` which just doesn't make sense, IMO. The same goes for the components of natural transformations. As long as there is only one, `η` is fine. But with two or more, we get letter soup like this excerpt from `Category.Construction.Functors`:

    ```Agda
      let open NaturalTransformation
          open Functor
      in begin
        (η β C ∘ η α C) ∘ F₁ F (g C.∘ f)  ≈⟨ refl ⟩∘⟨ homomorphism F ⟩
        (η β C ∘ η α C) ∘ F₁ F g ∘ F₁ F f ≈⟨ assoc ⟩
        η β C ∘ η α C ∘ F₁ F g ∘ F₁ F f   ≈⟨ refl ⟩∘⟨ pullˡ (commute α g) ⟩
        η β C ∘ (F₁ G g ∘ η α B) ∘ F₁ F f ≈⟨ refl ⟩∘⟨ assoc ⟩
        η β C ∘ F₁ G g ∘ η α B ∘ F₁ F f   ≈˘⟨ assoc ⟩
        (η β C ∘ F₁ G g) ∘ η α B ∘ F₁ F f ∎
    ```

    One can of course get around this by renaming things locally (e.g. `open Functor G renaming (F₀ to G₀; F₁ to G₁)`, and so on) but that just seems a like a hack to me (and it introduces lots of noise in the code).

[...]

PR #125, opened 2020-02-25
Implement Maclane's Comparison Functor To Eilenberg-Moore Categories

TOTBWF:
This PR implements the Eilenberg-Moore Comparison Functor, along with a few extra properties.

I was trying to also prove the uniqueness of the functor, but you end up in a very tricky position where you need to show that a bunch of equality proofs are equal.

### DIFF
--- a/Categories/Category/Construction/Properties/EilenbergMoore.agda
+++ b/Categories/Category/Construction/Properties/EilenbergMoore.agda
@@ -22,12 +22,10 @@ private
     o ℓ e : Level
     𝒞 𝒟 : Category o ℓ e
 
-open NaturalTransformation
-
-module _ {F : Functor 𝒞 𝒟} {G : Functor 𝒟 𝒞} (adjoint : Adjoint F G) where
+module _ {F : Functor 𝒞 𝒟} {G : Functor 𝒟 𝒞} (F⊣G : Adjoint F G) where
   private
     T : Monad 𝒞
-    T = adjoint⇒monad adjoint
+    T = adjoint⇒monad F⊣G
 
     𝒞ᵀ : Category _ _ _
     𝒞ᵀ = EilenbergMoore T
@@ -42,7 +40,8 @@ module _ {F : Functor 𝒞 𝒟} {G : Functor 𝒟 𝒞} (adjoint : Adjoint F G)
     module F = Functor F
     module G = Functor G
 
-    open Adjoint adjoint
+    open Adjoint F⊣G
+    open NaturalTransformation
 
   -- Maclane's Comparison Functor
   ComparisonF : Functor 𝒟 𝒞ᵀ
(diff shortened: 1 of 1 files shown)


JacquesCarette:
Generally looks good, but it might be another couple of days before I can do a full code review.

For uniqueness, we've managed to do that elsewhere by being quite careful with the definitions - @HuStmpHrrr is the real expert there.  Obvious question: are you sure you want definitional uniqueness?

HuStmpHrrr:
i will check on Saturday. I am buried by the reports I have to write.

Categories/Category/Construction/Properties/EilenbergMoore.agda
```diff
@@ -0,0 +1,103 @@
...
+
+private
+  variable
+    o ℓ e : Level
+    𝒞 𝒟 : Category o ℓ e
+
+open NaturalTransformation
+
+module _ {F : Functor 𝒞 𝒟} {G : Functor 𝒟 𝒞} (adjoint : Adjoint F G) where
+  T : Monad 𝒞
```
HuStmpHrrr: all these definitions and modules are now become public. probably they should be under a `private` block

Categories/Category/Construction/Properties/EilenbergMoore.agda
```diff
@@ -0,0 +1,103 @@
...
+  module 𝒞ᵀ = Category 𝒞ᵀ
+
+  module T = Monad T
+  module F = Functor F
+  module G = Functor G
+
+  open Adjoint adjoint
+
+  open Functor F
+  open Functor G renaming (F₀ to G₀; F₁ to G₁)
```
HuStmpHrrr: the usual style now is to avoid renaming, but rather use `module`. then the code is much cleaner and more predictable: https://github.com/agda/agda-categories/issues/103#issuecomment-572268282

e.g.

```agda
module G = Functor G
-- later on
G.F₀
```

Categories/Category/Construction/Properties/EilenbergMoore.agda
```diff
@@ -0,0 +1,103 @@
...
+  open Functor T.F renaming (F₀ to T₀; F₁ to T₁)
+
+  -- Maclane's Comparison Functor
+  K : Functor 𝒟 𝒞ᵀ
```
HuStmpHrrr: I am OK with this name but I guess if we are exporting this name it should be a bit longer. surely there can be a private short name. something like
```agda
ComparisonF : Functor X Y
private
  K = ComparisonF
```

Categories/Category/Construction/Properties/EilenbergMoore.agda
```diff
@@ -0,0 +1,103 @@
...
+        𝒞 [ G₁ (counit.η B) ∘ G₁ (F₁ (G₁ f)) ] ∎
+      }
+    ; identity = G.identity
+    ; homomorphism = G.homomorphism
+    ; F-resp-≈ = G.F-resp-≈
+    }
+    where
+      open 𝒞.HomReasoning
+
+  open Functor K renaming (F₀ to K₀; F₁ to K₁)
```
HuStmpHrrr: the same as above. renaming like this is not a sustainable style.

TOTBWF:
Thanks for the review! I'm going to be away from a computer for a few days, but I'll make the changes once I get back.

Issue #124, opened 2020-02-22
Compact regions effect on agda-categories


JacquesCarette:
We need to test the effect of this (now by-default) 'option' of agda 2.6.1 on agda-categories. Lots of relevant information on https://github.com/agda/agda/issues/4316. I'm hoping/guessing that the memory usage will go down considerably.

JacquesCarette:
Anecdotally: yes, memory consumption went down. It doesn't seem worth it to dig deeper, we won't learn anything particularly new or useful.

Issue #123, opened 2020-02-22
Ordering problem in 2.6.1


JacquesCarette:
(@nad reports - see comment https://github.com/agda/agda/issues/4316#issuecomment-589976621)

With version 1.2 of the standard library and Agda 2.6.1-556be02 I get the following error:

    […]/Categories/Category/Instance/Properties/Posets.agda:252,54-55
    Poset a a a !=< _C_1608 ⇒ _A_1603
    when checking that the expression f has type _C_1608 ⇒ _A_1603

My guess is that this problem was caused by a fix of an ordering bug in the generalisation machinery (https://github.com/agda/agda/issues/3779). When I use generalisable variables I bind implicit arguments by name, to avoid surprises if the order of the arguments is changed (primarily due to changes in the program, not changes in Agda).

gallais:
This was my fix for this very problem:

https://github.com/gallais/agda-categories/pull/1/commits/946ab84381e1c7b639137efa624d909651695169#diff-e0d3af684f34ef68eabbe213274f7329L252-R253

Cf. agda/agda#4481 for why I had to use a pattern-matching lambda.

JacquesCarette:
When I had to fix similar problems, I tended to remove the generalizable variables (for the minimal number of definitions) and replace them with a specific order. More future-proof that way, as it didn't seem that the underlying Agda problems were going to be fixed soon.

HuStmpHrrr:
is this issue occurring for 2.6.0.1?

here is what I got:

```agda
/home/hu/projects/agda-categories/src/Categories/Category/Instance/Properties/Posets.agda:252,24-253,53
Unexpected implicit argument
when checking that the clause
.extendedlambda0 {h = h} {f = f} {g = g}
  = ⟨,⟩-unique {f = f} {g = g} {h = h}
has type
{h.A f.B : Category.Obj (Posets a a a)}
{f : (Posets a a a Category.⇒ h.A) f.B}
{g.B : Category.Obj (Posets a a a)}
{h : (Posets a a a Category.⇒ h.A) (f.B × g.B)}
{g : (Posets a a a Category.⇒ h.A) g.B} →
(Posets a a a Category.≈ (Posets a a a Category.∘ π₁) h) f →
(Posets a a a Category.≈ (Posets a a a Category.∘ π₂) h) g →
(Posets a a a Category.≈ ⟨ f , g ⟩) h
```

HuStmpHrrr:
[...]

Issue #119, opened 2020-02-08
consider moving source code into a `src/` folder


HuStmpHrrr:
this is helpful if we need to have some helper tools like stdlib.

JacquesCarette:
It's certainly something to do before we hit 1.0, if it gets done at all.

Issue #20, opened 2019-07-12
More automation (Travis), including sanity checking


JacquesCarette:
As per comments on c968c74, we need to set up a Travis instance with sanity checks to enforce the agreed upon conventions.

gallais:
I got a [basic build](https://github.com/gallais/agda-categories/pull/1) running to generate HTML docs just like for the stdlib,
unfortunately the library requires too much memory for the machines travis has.

Trying to build against `master` now to see whether compact regions will help.
Worried about #91

JacquesCarette:
Thanks a lot for working on this, the help is most appreciated.

PR #126, opened 2020-02-25
[ doc ] produce & upload html doc using travis

gallais:
Looks like the new compact-regions option allows us to typecheck `Everything.agda`
using travis.

**To get the docs properly deployed, someone with push access to this repo needs to**

* create the `gh-pages` branch (ideally it should be empty)
* create a token [here](https://github.com/settings/tokens) with the `[ ] repo` option ticked
* activate the travis builds [here](https://travis-ci.org/agda/agda-categories)
* add a new environment variable [here](https://travis-ci.org/agda/agda-categories/settings) called `GH_TOKEN` and whose value is the one that was just generated

Once all of this is done, this PR can be merged into master and the html docs
should appear at http://agda.github.io/agda-categories/Everything.agda

Ping #20 #119 #123 #124


JacquesCarette:
It might be a couple of days before I can dig in to this.

gallais:
The 'files changed' count is scary but 99% of them are just renamings from X to src/X.

HuStmpHrrr:
thank you for moving everything to `src/`. this is a good move.

HuStmpHrrr:
I like this change but I will let Jacques to merge.

.travis.yml
```diff
@@ -0,0 +1,87 @@
...
+
+dist: xenial
+
+cache:
+  directories:
+    - $HOME/.cabsnap
+
+matrix:
+  include:
+    - env: TEST=MAIN GHC_VER=8.4.4 BUILD=CABAL CABAL_VER=2.2 STDLIB_VER=1.2
```
JacquesCarette: Why this particular version of ghc and cabal?

.travis.yml
```diff
@@ -0,0 +1,87 @@
...
+      addons:
+        apt:
+          packages:
+            - cabal-install-2.2
+            - ghc-8.4.4
+          sources:
+            - hvr-ghc
+
+before_install:
+  - export PATH=/opt/ghc/$GHC_VER/bin:/opt/cabal/$CABAL_VER/bin:/opt/alex/3.1.7/bin:/opt/happy/1.19.5/bin:~/.cabal/bin/:$PATH;
```
JacquesCarette: pinning down alex and happy indirectly like this also makes it seem problematic (for later maintenance).

.travis.yml
```diff
@@ -0,0 +1,87 @@
...
+
+install:
+  - cabal update
+  - sed -i 's/^jobs:/-- jobs:/' $HOME/.cabal/config
+  # installing agda dev
+  - git clone https://github.com/agda/agda --depth=1
+  - cd agda
+  # checking whether .ghc is still valid
+  - cabal install alex happy
```
JacquesCarette: especially as you don't pin down the versions here.

.travis.yml
gallais: You have to pick a version, don't you? This one works so far (it's what we
are using for the stdlib).

JacquesCarette: Definitely, a version needs to be picked. I was sort-of hoping to find a way to pick it consistently with the choice from stdlib, in a way that would not require constant maintenance.

Maybe a dummy 'build' project (aka under agda, i.e. `agda/build` parallel to `agda/agda`, etc) that would contain these choices in a central way. For example.

gallais: Ideally I would want nightly builds of Agda so that all software projects
could just pull whatever is the latest without having to pin down a commit
known to build (although Agda has gotten way better at not breaking master)
and re-build from scratch.

No time to look into this though.

HuStmpHrrr:
is there anything that blocks this? I think hardcoding the version is fine, as long as it works the same way as stdlib, so ultimately there is a reference for the future.

JacquesCarette:
Not from my end - I just have not had the time.  I think it's good to go.

HuStmpHrrr:
@gallais i think we can merge this. could you resolve the conflict so we can merge?

gallais:
Done

HuStmpHrrr:
thx. I merge now.

gallais:
You have not created the branch, the token, and setup travis accordingly so
the documentation is not getting deployed.

HuStmpHrrr:
@gallais is there a github page with full instructions?

JacquesCarette:
I meant to do that (branch, etc) this weekend, but ran out of time.  Hopefully @HuStmpHrrr can get to it soon.

HuStmpHrrr:
>You don't have sufficient rights to enable this repo on Travis.
>Please contact the admin to enable it or to receive admin rights yourself. 

is it because I am not a member of the agda group?

JacquesCarette:
@gallais I will have enough time to take care of this tonight. I've done it before, your instructions are great.

@HuStmpHrrr that might be the case.  I think you'll need to ask to be in the Agda group.

JacquesCarette:
Ok, I've done all 4 steps.

gallais:
Everything worked except for the push that failed. Dropping the `&>/dev/null`
from the last line of the travis file may help us to understand why.

JacquesCarette:
I think I know - the token was for just the gh-pages branch, not all branches. I've fixed that.  I've that doesn't fix it, then I'll try the above.

Issue #78, opened 2019-10-31
Implement Mac Lane comparison Functor?


JacquesCarette:
See the Coq code at https://github.com/ekiciburak/ComparisonTheorem-MacLane .  There's a paper too.  Looks like fun.

JacquesCarette:
@TOTBWF how much of this issue is dealt with by your PR #125  ?

TOTBWF:
I implemented the Eilenberg-Moore version of this, which is pretty similar in character. Doing the Kleisli version ought to be pretty straightforward 🙂 I suspect that proving uniqueness will be easier as well, as you don't need to show that two T-modules/T-actions are equivalent, only objects and arrows.

I'll take a look into finishing this off in the next couple of days.

Taneb:
Can this issue be closed now?

TOTBWF:
Yeah, Should be good to go.

JacquesCarette:
I haven't forgotten - I just want to double-check for myself that it's all done. Most likely is.

JacquesCarette:
Sorry to be so slow - it's all great. And the proof is readable (and short), unlike the Coq one.

Issue #58, opened 2019-10-01
Where to organize the 'with K' part of the categories library?


JacquesCarette:
As discussion on various issues make clear (#44 being just the latest), it would be quite useful to have parts of the library, especially that part which makes strong use of `≡` such as `StrictCats, also be able to use the K axiom.  But the whole library is currently built using `--without-K` (on purpose).

So this leads to a non-trivial organization question. For example, should there be a `WithK` hierarchy right below `Categories` (or even parallel to it?) for that? Or should they be intertwined, the way it is for stdlib? I see pros and cons to both, and am undecided which I prefer.

sstucki:
Interesting question.

I don't like the idea of mixing with/without K. (I wonder why it is done like this in the stdlib.)

Having a separate top-level `WithK` module seems like a nice approach. I'd suggest that modules nested in `WithK` should be allowed to refer to those outside, but not vice-versa, so that without-K-modules are not accidentally tainted. And everything that can, should of course still go in the without-K-part of the hierarchy. One could then have a `WithK` version of some modules that re-export functionality from outside (e.g. the weak Grothendieck construction) and add extra proofs/constructions (e.g. the strict Grothendieck construction).

JacquesCarette:
If we go with top-level `WithK`, I completely agree that it ought to be able to refer to anything outside, but not the other way around. And indeed, re-exports are possible.

This might not fully solve the "what is Strict" issue, but at least it'll be clear that everything under the `WithK` hierarchy are strongly truncated.

sstucki:
OK, I'll take care of this when I get some time (and after reworking #42).

HuStmpHrrr:
if `WithK` is a separate hierarchy, wouldn't it be mostly empty?

JacquesCarette:
The sparser the better!  But I'm fine with that.

gallais:
> I don't like the idea of mixing with/without K. (I wonder why it is done like this in the stdlib.)

In the stdlib we adopted `--without-K` and `--safe` as defaults. As a consequence
`--with-K` and "unsafe" are considered outliers and even sometimes evidence of the
compiler not being advanced enough to recognize K-freeness or safety.
Correspondingly modules tainted by these options are left behind, like an afterthought.

JacquesCarette:
[...]

PR #132, opened 2020-03-07
Implement the Kleisli Comparison Functor

TOTBWF:
This PR implements the Kleisli Comparison functor, finishing up #78.

I tried to prove uniqueness here as well, but ran into similar issues as before. It appears that the Coq development mentioned in the issue uses proof irrelevance/funext heavily, which would explain how they are able to prove it.

### DIFF
--- a/src/Categories/Category/Construction/Properties/EilenbergMoore.agda
+++ b/src/Categories/Category/Construction/Properties/EilenbergMoore.agda
@@ -83,8 +83,8 @@ module _ {F : Functor 𝒞 𝒟} {G : Functor 𝒟 𝒞} (F⊣G : Adjoint F G) w
       Module⇒.arr (𝒞ᵀ [ Fᵀ.F₁ f ∘ (hid 𝒞ᵀ ≡.refl) ])       ∎
     }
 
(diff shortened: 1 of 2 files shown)


TOTBWF:
Not sure why the CI is failing here, looks like it couldn't checkout agda properly?

JacquesCarette:
I'm going to guess that CI is failing because others were experimenting with various settings (lots of comments on a couple of related issues), and you submitted at a time when things were broken. I see nothing in your code that would cause a problem.

Rather than prove 'strict' uniqueness, you might want to show it up to natural isomorphism using a trivial NI (i.e. both natural transformations being the identity).

TOTBWF:
I tried the natural transformation route originally, but you still get mired down, as you have to use heterogeneous identity morphisms to establish the natural transformations.

TOTBWF:
After some work, I was able to get a proof of uniqueness up to NI, but it does require UIP. Perhaps it is time to start work on #58.

JacquesCarette:
Do you have the work checked-in to some branch?  I'd like to take a look.

That something **requires** UIP would be quite interesting, especially if it could in fact be shown to be equivalent to it.

Issue #135, opened 2020-03-08
look for a good definition of finite categories


HuStmpHrrr:
# Why is this interesting?

This allows a good use of topos. A topos, among other properties, has all finite limits. Though we currently have a definition of topos already, it is defined via equalizers. As a result, the existence of all finite limits needs to be proven as a property.

I've been thinking about this on and off but I don't see an obvious way to define this notion.

# Why is this nontrivial?

The reason why this is not immediate ultimately goes to how this library is implemented. In this library, we rarely talk about the identities of objects or morphisms; we only quantify the external behaviors of a category. This makes the definition of finite categories more challenging because classically finite categories counts the number of objects and morphisms and we cannot do that without considering the identities of them.

That said, I think it is still possible. Though we can't talk about identities of arbitrary objects, we should be able to talk about identities of known objects if we know what they are. For example, if a category has objects as `Fin n`, then we can discuss the identity of objects. In addition, if we require the morphisms between this category to have UIP and finite, we should obtain a special case of a finite category. Then the general notion of finite categories can be defined as an weak equivalence to this special family of finite categories.

I hope that makes sense and is viable. Timany and Jacobs [2016] has a notion of finite categories but they don't use setoids: https://drops.dagstuhl.de/opus/volltexte/2016/6000/

sstucki:
I have not read the paper you cite but I think you're idea of defining a finite category as one that has a finite shape is on the right track.

I think this can be extended into a rigorous argument as follows. Classically, a finite set is one that is in bijective correspondence with the set _{0,...,n}_ for _n_ a natural number (i.e. `Fin n`). If we are instead talking about setoids, then the bijection should be replaced by an isomorphism of setoids, with `Fin n` being a setoid modulo propositional equality (in this case, propositional equality is decidable, and UIP should hold definitionally). One could go further and replace setoids with groupoids and setoid isomorphism with groupoid equivalence, and so on for higher groupoids. But for our purposes setoids should be sufficient.

HuStmpHrrr:
@sstucki I have a WIP PR doing that. in fact, the PR pushes even further: it uses propositional equality and then we obtain both UIP for objects and morphisms. looks good to me. will do some proof of concept later.

JacquesCarette:
[...]

PR #136, opened 2020-03-09
finite categories

HuStmpHrrr:
#135


HuStmpHrrr:
yes, there needs to be some helper properties which are not available in stdlib.

HuStmpHrrr:
OK, I consider it done. the critical theorem is that in a finitely complete category, any finite shape determines a limit.

PR #137, opened 2020-03-10
Implement the selection functor

TOTBWF:
This PR implements the lifting of functions on objects of discreet categories into functors. I'm open to a better name than `Select`, I just couldn't come up with any better ones myself.

### DIFF
--- a/src/Categories/Functor/Construction/FromDiscrete.agda
+++ b/src/Categories/Functor/Construction/FromDiscrete.agda
@@ -3,7 +3,7 @@ open import Categories.Category
 
 -- You can transform functions out of discrete
 -- categories into functors.
-module Categories.Functor.Construction.Select {o ℓ e} (𝒞 : Category o ℓ e) where
+module Categories.Functor.Construction.FromDiscrete {o ℓ e} (𝒞 : Category o ℓ e) where
 
 open import Relation.Binary.PropositionalEquality as ≡ using (_≡_)
 
@@ -16,8 +16,8 @@ private
 open Category 𝒞
 open 𝒞.HomReasoning
 
-Select : ∀ {o} {A : Set o} → (A → Obj) → Functor (Discrete A) 𝒞
-Select {o} {A = A} select = record
+FromDiscrete : ∀ {o} {A : Set o} → (A → Obj) → Functor (Discrete A) 𝒞
+FromDiscrete {o} {A = A} select = record
   { F₀ = select
   ; F₁ = λ { ≡.refl → id }
   ; identity = refl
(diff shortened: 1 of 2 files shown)


sstucki:
I think this is the action of the `Discrete` functor: https://github.com/agda/agda-categories/blob/master/src/Categories/Adjoint/Instance/Discrete.agda

So maybe one could call it `Discrete₁`?

I think it's a good idea to factor the components of adjoint functors into separate modules -- not least because they can often be made more universe-polymorphic that way. But we should probably try to come up with a system to consistently name and organize the corresponding modules.

sstucki:
> I think this is the action of the Discrete functor

...whoops, I spoke too soon. This is a different (more general?) construction?

JacquesCarette:
`FromDiscrete` ?  Especially if done via a qualified import, it could become `Functor.FromDiscrete`, which isn't too bad.  Variants on that idea might be even better, that's the one that came to mind now.

JacquesCarette:
Do you have references for where this construction occurs?  That might help scrounge up a name too.

TOTBWF:
I mostly need it for constructing (co)limits. For example, when viewing a (co)product as a (co)limit, you need a functor from the 2 element  discrete category to select your two objects.

`FromDiscrete` seems like it could be a good choice!

JacquesCarette:
That application seems to have non-trivial overlap with what @HuStmpHrrr is doing with finite categories -- have you looked at that?

sstucki:
> I mostly need it for constructing (co)limits. For example, when viewing a (co)product as a (co)limit, you need a functor from the 2 element discrete category to select your two objects.

But that seems very specific to (co)products, no? If you want to define (co)limits over diagrams with arrows (e.g. pullbacks/pushouts) a functor out of a discrete category won't work, right?

Still, I think it might be a useful abstraction for other purposes.

HuStmpHrrr:
I don't think you can generate products from discrete categories. the category generating products as limits is the following

```
a <------- b -------> c
```

which is not discrete.

TOTBWF:
@sstucki Yes, it doesn't work for limits over diagrams with non-identity arrows. However, I found myself implementing this functor for `Discrete Bool`, and realized that I could generalize it!

@HuStmpHrrr You do use a discrete category, see https://ncatlab.org/nlab/show/limits+and+colimits+by+example#product for more details.

There definitely seems to be some overlap with the finite category stuff. Maybe we could have a notion of a discrete shape for any finite type?

HuStmpHrrr:
yes, you are right. I was not thinking.

JacquesCarette:
How are we doing on finding a better name?

TOTBWF:
I think `FromDiscreet` is the best one so far. Will make the changes later today

HuStmpHrrr:
FYI, I am about to merge the finite category PR which I have prepared forever. from there, I proved a theorem which shows a limit can be derived from a finite shape if a category is finitely complete. special limits like products are easy corollaries.

WolframKahl:
> I  think `FromDiscreet` is the best one so far. 

`FromDiscrete` ?

TOTBWF:
>>    I think FromDiscreet is the best one so far.

> FromDiscrete ?

Looks like auto-correct has it's own opinions on what this should be named!

Anyways, I've renamed it to `FromDiscrete`, so this should be good to go.

PR #161, opened 2020-04-02
evaluation functor is continuous

HuStmpHrrr:


src/Categories/Category/Construction/Functors.agda
```diff
@@ -76,6 +76,9 @@ eval {C = C} {D = D} = record
         open MR D
         open HomReasoning
 
+evalF : ∀ (C : Category o ℓ e) (D : Category o′ ℓ′ e′) → Category.Obj C → Functor (Functors C D) D
```
JacquesCarette, line 79: Why are `C` and `D` explicit arguments?

HuStmpHrrr: `C` is almost always not inferrable. `D` can only be inferred from the return type which might not always be available.

Issue #140, opened 2020-03-18
travis should use 1.3 for stdlib


HuStmpHrrr:

gallais:
It should hopefully be as simple as changing [this line](https://github.com/agda/agda-categories/blob/master/.travis.yml#L19)

Issue #131, opened 2020-03-07
use 2.6.0.1 in travis


HuStmpHrrr:
we shouldn't use master branch of agda in travis. instead, we should use the latest release. that means we will have to update the version every release but hopefully it's fine.

ideally, it should be agda's continuous integration to compile this library to make sure they don't break anything.

HuStmpHrrr:
c.f. #123

HuStmpHrrr:
https://github.com/agda/agda-categories/commit/294091f6c5b3e5668cafeaa89187add95f0d72a7

gallais:
Broke the build.

I ended up using master precisely because travis did not have enough memory to
compile agda-categories unless we were using the newer version of Agda using
compact regions.

Issue #91, opened 2019-12-14
Test against agda HEAD?


JacquesCarette:
I'm seeing a lot of issues coming up against current HEAD, which is slated to be released soon. I'm concerned that only some of the bugs will be found, and that this library will stop type-checking. Best bet to prevent that is to test against HEAD and report any/all bugs so found. [I'm sure they would then be fixed]

Unfortunately, I don't have the bandwidth to do that right now - anyone else?

JacquesCarette:
This is being worked on - I've got 2.6.1 installed on a VM (though I need to reinstall), and it looks like @asr is also trying [thanks!]

HuStmpHrrr:
I have some older version of agda working (but still newer than 2.6.0.1 thx to cubical), I can only see some warnings from stdlib though. I don't think the library will receive much shock modulo the part coming from stdlib.

HuStmpHrrr:
btw you can run multiple versions on the same machine at the same time. the way I do it is I pick a `stack.yaml` file and do `stack install`, which will install the binaries to `~/.local/bin`. I just need to rename the binaries to `*-2.6.1` and I can choose different versions of agda from emacs. very handy.

asr:
Using `cabal-install`, you can use `cabal install --program-suffix=-2.6.1`. I guess there is something similar in `stack`.

JacquesCarette:
I'll try to set that up on my servers.  It does look like, modulo a couple of small fixes, it works.  Unfortunately, it also looks like the memory requirements have gone up. [I'm working on splitting up the worst offender; it's also kind of fun].

sstucki:
This discussion reminds me that it would be nice to set up continuous integration for the library, e.g. using Travis.

Maybe we could adopt the Travis configuration from the standard library?

JacquesCarette:
Yes, absolutely!  I just haven't had any spare cycles to do that.

HuStmpHrrr:
I doubt Travis provides sufficient memory though. I remember it's 6g?

sstucki:
Yea, that might be an issue. I wonder how the standard library gets around this.

Their config suggests they get by with under 4G:

```
  - export RTS_OPTIONS="+RTS -M3.5G -H3.5G -A128M -RTS"
```
(From https://github.com/agda/agda-stdlib/blob/master/.travis.yml)

Do they just have less memory-hungry code? There must be some rather complex bits in the stdlib, no?

MatthewDaggitt:
[...]

PR #164, opened 2020-04-11
Added a helper for Monoidal creation

MirceaS:
Because 3 of the fields can be proved from the other

### DIFF
--- a/src/Categories/Category/Monoidal/Core.agda
+++ b/src/Categories/Category/Monoidal/Core.agda
@@ -107,3 +107,78 @@ record Monoidal : Set (o ⊔ ℓ ⊔ e) where
                            ≈ α⇒               ⇒⟨ (X ⊗₀ Y) ⊗₀ Z ⊗₀ W ⟩
                              α⇒
                            ⟩
+
+private
+  record Monoidal′ : Set (o ⊔ ℓ ⊔ e) where
+    infixr 10 _⊗₀_ _⊗₁_
+
+    field
+      ⊗  : Bifunctor C C C
+      unit : Obj
+
+    open Functor ⊗
+
+    _⊗₀_ : Obj → Obj → Obj
+    _⊗₀_ = curry′ F₀
+
+    _⊗₁_ : X ⇒ Y → Z ⇒ W → X ⊗₀ Z ⇒ Y ⊗₀ W
+    f ⊗₁ g = F₁ (f , g)
+
+    field
+      unitorˡ    : unit ⊗₀ X ≅ X
+      unitorʳ    : X ⊗₀ unit ≅ X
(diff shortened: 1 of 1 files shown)


MirceaS:
> Isn't there a problem where this cannot be used, as `Monoidal'` is private, thus not externally visible? Or is it ok because `monoidalHelper` is always meant to be called on an explicit record?

I followed the 'helper' pattern present in `Categories.Category.Monoidal.Symmetric`. The simplified record is private there. I think this is because, ideally, you want to only call monoidalHelper on an explicit record as the new record is only used as an injection into the main record.

I was under the impression that `Categories.Category.Core` did the same thing, but there the primed record is not made private. So we might need to tweak that so that `Category` and Symmetric are consistent with eachother.

I'm happy to change my implementation to the style in `Core.agda` and make the primed record public if you feel like that would be better.

JacquesCarette:
Right - I'm not sure we've evolved a style yet.  @HuStmpHrrr opinion?

HuStmpHrrr:
`private` is OK because it actually prevents the definition from being (easily) referred to externally. the helper can still be used because ultimately it just wants a record, which can be defined anyways.

JacquesCarette:
Ok, I'll merge as is then.

MirceaS:
Should we also change the definition of `Category'` to be private as well then?

JacquesCarette:
Probably, yes.  [I just don't want to do the 15 minute wait for `make` to ensure that it doesn't break anything right now, I've got a bunch of other things I should be working on.]

MirceaS:
> Probably, yes. [I just don't want to do the 15 minute wait for `make` to ensure that it doesn't break anything right now, I've got a bunch of other things I should be working on.]

Can't we have GitHub do that automatically, as it does for agda-stdlib?
There, any PR is checked using Travis-CI

JacquesCarette:
Yes - we even have an open issue for that. No one has yet tackled it.  If you'd like to, we'd be quite happy to merge that in!

MirceaS:
> Yes - we even have an open issue for that. No one has yet tackled it. If you'd like to, we'd be quite happy to merge that in!

I look at the open issues and I didn't see anything about that...
I also noted that you have a `.travis.yml` file in the repo already? Were there any previous attempts to get Travis working in the past that didn't go through?

JacquesCarette:
There are various discussion in issues #20 #91 #126 #131 and #140 .  Looks like Travis is currently only used for producing documentation, but not for automatic testing.  The problem, until 2.6.1, was that the library took too much memory to check.  Now it's ok.  So I think it's worth trying again.

HuStmpHrrr:
when it generates the doc, it will need to compile the files first. that already suffices. usually a cycle takes around 40 minutes.

PR #165, opened 2020-04-26
Fix typo: catesian -> cartesian

JDemler:


JDemler:
I just realized, that this is probably a breaking change. Is that relevant in this case?

JacquesCarette:
It's only relevant / a problem if it actually does break something in the library.  I'll wait until CI comes back green.  If you do 'make' locally, you'll be able to see if it breaks anything.  If it does, add those fixes to this PR please.

HuStmpHrrr:
it's fine. a typo shouldn't really be a permanent thing.

PR #168, opened 2020-04-30
Monoid objects

Taneb:
This adds a definition of monoid objects in a monoidal cateogry, and the constructed category of monoids.

This is the first time I'm contributing to an agda project, so please do give me feedback


src/Categories/Category/Construction/Monoids.agda
```diff
@@ -0,0 +1,74 @@
...
+
+Monoids : Category (o ⊔ ℓ ⊔ e) (ℓ ⊔ e) e
+Monoids = record
+  { Obj = Monoid
+  ; _⇒_ = Monoid⇒
+  ; _≈_ = λ f g → arr f ≈ arr g
+  ; id = λ {A} → record
+    { arr = id
+    ; preserves-μ = begin
+      id ∘ μ A
```
JacquesCarette, line 26: We prefer to use side-by-side alignment whenever possible.

src/Categories/Category/Construction/Monoids.agda
```diff
@@ -0,0 +1,74 @@
...
+        ≈⟨ identityˡ ⟩
+      μ A
+        ≈⟨ sym identityʳ ⟩
```
JacquesCarette, line 29: I think these 2 lines can be replaced with id-comm (or id-comm-sym).

src/Categories/Category/Construction/Monoids.agda
```diff
@@ -0,0 +1,74 @@
...
+      μ A ∘ id
+        ≈⟨ (refl ⟩∘⟨ sym (Functor.identity ⊗)) ⟩
```
JacquesCarette, line 31: we tend to prefer putting the `sym` on the `≈` instead.

src/Categories/Category/Construction/Monoids.agda
```diff
@@ -0,0 +1,74 @@
...
+      μ A ∘ id ⊗₁ id
+        ∎
+    ; preserves-η = identityˡ
+    }
+  ; _∘_ = λ {A B C} f g → record
+    { arr = arr f ∘ arr g
+    ; preserves-μ = begin
+      (arr f ∘ arr g) ∘ μ A
+        ≈⟨ assoc ⟩
```
JacquesCarette, line 40: look at the push/pull combinators, as they embody this exact idea of assoc then operate on one side.

src/Categories/Category/Construction/Monoids.agda
```diff
@@ -0,0 +1,74 @@
...
+        ≈⟨ preserves-η f ⟩
+      η C
+        ∎
+    }
+  ; assoc = assoc
+  ; sym-assoc = sym-assoc
+  ; identityˡ = identityˡ
+  ; identityʳ = identityʳ
+  ; identity² = identity²
+  ; equiv = record
```
JacquesCarette, line 68: does `equiv = equiv` not work? [if not, a comment should say so]

src/Categories/Category/Construction/Monoids.agda
```diff
@@ -0,0 +1,74 @@
+{-# OPTIONS --without-K --safe #-}
+open import Categories.Category.Core
+open import Categories.Category.Monoidal.Core
+
+module Categories.Category.Construction.Monoids {o ℓ e} {𝒞 : Category o ℓ e} (C : Monoidal 𝒞) where
```
JacquesCarette, line 5: Perhaps this should be called ...`Construction.InternalMonoids` ?  The point is that the category of monoids (in `Setoid`, as they exist in stdlib) would also be expected to have that name. The file can still define `Monoids` as you have it, i.e. without changing the name of the category.

src/Categories/Category/Construction/Monoids.agda
Taneb: It doesn't. This is basically the same as `Cones`.

src/Categories/Category/Construction/Monoids.agda
Taneb: I ended up using `introʳ` to replace the second of these, along with the following line, which simplifies it about as much.

src/Categories/Category/Construction/Monoids.agda
JacquesCarette: That's what I expected, but figured I needed to ask, as the next person reading the code will too, and so on.

src/Categories/Category/Construction/Monoids.agda
Taneb: Hmm, I'm not entirely convinced. I'd expect that to be in `Categories.Category.Instance.Monoids`. Perhaps I could add a note of explanation, and if/when that gets added, the two modules can refer to each other?

JacquesCarette: Good point!  `Construction` vs `Instance` is indeed an appropriate differentiator.  A note would be splendid.

src/Categories/Category/Construction/Monoids.agda
Taneb: Honestly, I'm not entirely sure why it doesn't work. The error message suggests it's because the `_⇒_` of Monoids has a different level to the `_⇒_` of `𝒞`

src/Categories/Category/Construction/Monoids.agda
Taneb: Thank you for the tip! I've shortened pretty much all the proofs now.

src/Categories/Category/Construction/Monoids.agda
Taneb: Have updated it to use this style.

src/Categories/Category/Construction/Monoids.agda
Taneb: I eliminated this particular instance, but I've changed another place to use `≈˘⟨_⟩` instead, where it makes it a little clearer.

PR #177, opened 2020-06-10
Functor instance for List

glittershark:
Add proof of a functor instance for List from the agda stdlib.

Thinking of putting functor instances all in one module per-category like this
since they're pretty short, let me know if I should do something else.

----

#

### DIFF
--- a/src/Categories/Functor/Instance/Sets.agda
+++ b/src/Categories/Functor/Instance/Sets.agda
@@ -1,0 +1,23 @@
+{-# OPTIONS --safe --without-K #-}
+module Categories.Functor.Instance.Sets where
+
+open import Level
+open import Categories.Functor using (Endofunctor)
+open import Categories.Category.Instance.Sets
+open import Relation.Binary.PropositionalEquality as ≡ using (_≡_)
+
+open import Function using (id; λ-; _$-)
+open import Categories.Category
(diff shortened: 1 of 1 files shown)


src/Categories/Functor/Instance/Sets.agda
```diff
@@ -0,0 +1,23 @@
...
+import Data.List as List
+import Data.List.Properties
+
+ListFunctor : ∀ {o} → Endofunctor (Sets o)
+ListFunctor = record
+  { F₀ = List.List
+  ; F₁ = List.map
+  ; identity = λ {_} {x} → map-id x
+  ; homomorphism = λ {_} {_} {_} {_} {_} {x} → map-compose x
+  ; F-resp-≈ = λ {_} {_} {_} {_} f≈g {x} → map-cong (λ x → f≈g {x}) x
```
glittershark, line 20: is there a utility function to convert pointwise equality from explicit to implicit and vice versa? this feels really ugly

gallais: [`_$-`](http://agda.github.io/agda-stdlib/Function.Base.html#2399) perhaps?

glittershark:
https://github.com/agda/categories-examples/pull/2

PR #178, opened 2020-06-11
Pullback squares

glittershark:
Add a definition of a pullback square, mostly inspired by the
definition[1] in copumpkin's library, with some slighly simplified
notation and a little extra documentation.

I'm still working on showing that this definition is equivalent to a limit of a
cospan-shaped diagram, but figured I'd submit this in the meantime


glittershark:
of course as soon as I open this I realized that there's already a Pullback in `Categories.Diagram`. Ah well - this was a reasonable exercise

PR #179, opened 2020-06-13
prove setoids are locally CCC

HuStmpHrrr:


HuStmpHrrr:
@JacquesCarette OK I am done!

JacquesCarette:
This is really great. Way harder than expected.

PR #182, opened 2020-07-20
Some work on the MultiCategory definition

laMudri:
This PR contains a few updates to the definition of `MultiCategory`, as summarised in the commit messages. Longer descriptions are as follows:

* Previously, the type of `_∘_` didn't ensure any compatibility between the morphisms. Now, morphisms compose along `aₙ` only.
* `∘-resp-≈` has been implemented as suggested. I just did the obvious thing, so I'm not sure why it wasn't added before. It looks okay to me, though.
* Usually, definitions of structures like this have explicit levels, as generally they are not checkable from anywhere (even if they are usually synthesisable from explicit instances). In any case, I don't think it's possible to infer `ı` from anywhere. I have made `MultiCategory` follow the explicit level convention.

I intend to add an instance (commutative monoids and multilinear maps) as a sanity check and example. This should be ready pretty soon.


JacquesCarette:
Looks good. I'll wait until Travis finishes though.

PR #184, opened 2020-08-01
refactoring terminal/initial to have the predicate form

HuStmpHrrr:
Closes #172


HuStmpHrrr:
unfortunately the continuity of evalF requires some non-trivial effort. should be dealt with later.

PR #189, opened 2020-08-07
Rigid category

DreamLinuxer:


HuStmpHrrr:
thx. looks very good to me. we should probably emphasize that it is a definition for left rigid

DreamLinuxer:
> thx. looks very good to me. we should probably emphasize that it is a definition for left rigid

It requires the category to be symmetric, so the left or right rigidity means the same thing.

PR #191, opened 2020-08-13
Feature/coend

JacquesCarette:
Starting to work on the Coend calculus, formalizing everything in Fosco Loregian's book bit by bit.


src/Categories/NaturalTransformation/Extranatural.agda
```diff
@@ -0,0 +1,52 @@
+{-# OPTIONS --without-K --safe #-}
+module Categories.NaturalTransformation.Extranatural where
+
+-- Although there is a notion of Extranatural in Categories.NaturalTransformation.Dinatural,
```
HuStmpHrrr, line 4: it looks somewhat close to dinaturality. I think dinatural is a special base of this definition. do you have a source of this definition?

HuStmpHrrr:
is it the coend calculus book you are following?

src/Categories/Diagram/End/Properties.agda
```diff
@@ -22,6 +26,33 @@ private
     o ℓ e : Level
     C D E : Category o ℓ e
 
+module _ {o ℓ e o′ ℓ′ e′} {C : Category o ℓ e} {D : Category o′ ℓ′ e′}
+  (F : Bifunctor (Category.op C) C D) where
+  open Wedges F
+
+  -- Being an End is the same as being a Terminal object in the category of Wedges
```
HuStmpHrrr, line 33: I think we can obtain most of the conclusions here via duality

src/Categories/NaturalTransformation/Extranatural.agda
JacquesCarette: I am following the coend calculus book, page-by-page.  Yes, dinaturality is a special case, but the proof is a bit complicated (I skipped that).

src/Categories/Diagram/End/Properties.agda
JacquesCarette: From the Coend case [which I readily believe], or from some other results?  I'm quite fine rewriting some of the proofs to go via duality, once we fix a convention of what results should be done 'by hand' and which should be via duality. We should be consistent which of the 'side' we (arbitrarily!) chose to do first.

HuStmpHrrr: yeah it looks derivable from the coend case. I think the idea is just to save effort from dual proofs once we have one side already. I don't think one side is better than another, so this choice can be a bit arbitrary.

src/Categories/Category/Equivalence/Preserves.agda
```diff
@@ -0,0 +1,79 @@
+{-# OPTIONS --without-K --safe #-}
+
+-- Categorical equivalences preserve various structures
```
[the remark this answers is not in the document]
JacquesCarette: [Thanks for continuing to review this.]  Yes, it would make sense to have that.

src/Categories/Diagram/Duality.agda
```diff
@@ -38,6 +38,20 @@ private
     A B : Obj
     f g : A ⇒ B
 
+terminal⇒coInitial : IsTerminal C A → IsInitial (Category.op C) A
```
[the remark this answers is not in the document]
JacquesCarette: Oops, missed that, I will fix that, thanks!

HuStmpHrrr:
this PR has been hanging for quite a while. should we merge it?

JacquesCarette:
I still want to fix a couple of things first. Then we can merge. I should finally have time again soon.

src/Categories/Category/Equivalence/Preserves.agda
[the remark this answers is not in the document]
JacquesCarette: I've created issue #237 for this bit.

PR #192, opened 2020-08-23
Add some extra equalities to Pullback

TOTBWF:
This PR adds some extra equality proofs to Pullback, mirroring similar proofs that can be found in [Categories.Diagram.Equalizer](https://github.com/agda/agda-categories/blob/master/src/Categories/Diagram/Equalizer.agda#L38). This also adds proofs that Equalizers and Pullbacks are unique up to isomorphism.


TOTBWF:
Added some proofs that Equalizers and Pullbacks are unique up to isomorphism as well.

HuStmpHrrr:
yeah that one looks an overkill to me too. otherwise looks good.

TOTBWF:
I've removed `equalize-resp-≈′`, are there any other little helper proofs we might want to add while we are here?

HuStmpHrrr:
i think it's good to go

Issue #106, opened 2020-01-06
Monoidal Groupoid


JacquesCarette:
In my 'negative thinking' exploration, it has become clear that I need to enrich over a monoidal groupoid too [this is how to get Setoid, since otherwise one only gets Poset].  The question here is how to 'encode' monoidal groupoid.  For using them for enriching, there are two obvious ways:
1. `{C : Category} (G : IsGroupoid C) (M : Monoidal C)`
2. `{C : Category} (MG : MonoidalGroupoid C)`
where `MonoidalGroupoid` is the obvious record with 2 fields.

They are both 'equivalent' in some ways, so the question is one of usability / expectations.  Opinions?

sstucki:
A third option would be:

  3. `(MG : MonoidalGroupoid)` where everything is in one record (i.e. `MonoidalGroupoid` "extends" `Groupoid` with another field).

I'm tempted to also add a `MonoidalCategory` record. That would simplify the parametrization of the various enriched modules quite a bit and eliminate duplicate code that just opens and imports the same fields from the two records over and over.

JacquesCarette:
Indeed.  Sometimes it's useful to be able to say you have 2 monoidal structures on the same category (which is what option 1 makes easiest), but sometimes you have a unique one, where option 3 is likely simplest.

HuStmpHrrr:
I'd prefer option 2.  in option 1, G and M are disjoint so option 2 makes no harm. in addition, it's clear to write. we can add another definition describing option 3, so option 2 seems the right solution.

JacquesCarette:
So option 2 is basically along the lines of "a category C along with extra structure that makes it into a Monoidal Groupoid" while option 3 is "a Monoidal Groupoid".  Option 2 is crucial when you want to talk about 2 different structures on the same category, while option 3 makes sense when you really are not all that interested in the category itself, but all of the structure.

I think we should provide both!  And that, given our current naming conventions, option 2 should be called `IsMonoidalGroupoid` and option 3 `MonoidalGroupoid`. And both ought to be defined in the same file.

sstucki:
> I think we should provide both! And that, given our current naming conventions, option 2 should be called `IsMonoidalGroupoid` and option 3 `MonoidalGroupoid`. And both ought to be defined in the same file.

Sounds good to me!

PR #195, opened 2020-08-30
fix the definition of cartesian functor

### LINKED ISSUE #194: is the definition of product preserving functor correct?
https://github.com/agda/agda-categories/blob/master/src/Categories/Functor/Cartesian.agda

I find it's not quite the same as what I would expect. for example, 
https://github.com/agda/agda-categories/blob/dfa6dd47a23992e6b9108e09e42d0bf5d881d719/src/Categories/Functor/Cartesian.agda#L26
this doesn't look right to me as this morphism alone is not enough to show `Functor.F₀ F CA.⊤` to be a terminal object. consider a category with global elements, then in this category, this morphism just exists. this is not sufficient to prove this object is also a terminal object. the right definition should ensure that `Functor.F₀ F CA.⊤` is a terminal object.

the same goes the products, a natural transformation doesn't look enough to me.

HuStmpHrrr:
Closes #194

### DIFF
--- a/src/Categories/Category/Cartesian/Structure.agda
+++ b/src/Categories/Category/Cartesian/Structure.agda
@@ -10,11 +10,11 @@ open import Categories.Category.Cartesian
 
 record CartesianCategory o ℓ e : Set (suc (o ⊔ ℓ ⊔ e)) where
   field
-    𝒞 : Category o ℓ e
-    cartesian : Cartesian 𝒞
+    U : Category o ℓ e          -- U for underlying
+    cartesian : Cartesian U
 
-  module 𝒞 = Category 𝒞
+  module U = Category U
   module cartesian = Cartesian cartesian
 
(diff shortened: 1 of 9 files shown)


src/Categories/Category/Cartesian/Structure.agda
```diff
@@ -0,0 +1,20 @@
+{-# OPTIONS --without-K --safe #-}
+
+-- here we define the structure version of a cartesian category
+module Categories.Category.Cartesian.Structure where
```
Taneb, line 4: Is this a pattern we'd like to see elsewhere in the library? For example, for monoidal categories?

HuStmpHrrr: there was some debate going on, but I think it's good to have this pattern around: https://github.com/agda/agda-categories/issues/106

PR #201, opened 2020-09-07
Monoids and groups regarded as categories

Taneb:


JacquesCarette:
Even though these are quite small files, and they are given often as examples, I'm still not sure about also having them in the library. If they are going to be re-used for sure, then definitely. Otherwise, I'm a little tempted to have them in categories-examples instead.

The above is my current (weak) opinion.

sstucki:
I think they should stay for the same reason we have `Categories.Category.Construction.{Thin,0-Groupoid}`: they are useful when one wants to treat structures such as posets or monoids as degenerate categories. For example, graded monads are monoidal functors where the domain is a pre/partially ordered monoid seen as a monoidal category.

src/Categories/Category/Construction/MonoidAsCategory.agda
```diff
@@ -0,0 +1,23 @@
...
+
+open import Data.Unit
+open import Level
+
+open import Categories.Category.Core
+
+open Monoid M
+
+-- A monoid is a category with one object
+MonoidAsCategory : Category zero c ℓ
```
HuStmpHrrr, line 13: I would parameterize the first level.

JacquesCarette:
Perfect, that's 2 people advocating for (@Taneb as author, obviously, and @sstucki too). My "I'm not sure" was quite weak, so I'm now convinced.

Taneb:
So, what remains to be done here?

PR #202, opened 2020-09-08
Add bundled forms of Epi and Monomorphisms, as well as some useful notation

TOTBWF:
This PR adds a bundled forms of Epi and Monomorphisms, which can make some type signatures a little bit nicer. It also adds some notation similar to that found in `Categories.Category` for Epi, Mono, and Isomorphisms, which can make things nicer in contexts where you have multiple categories.

## Notes
I'm not super set on the name of `Categories.Morphism.Notation`, but I didn't want to do something like `Categories.Morphism.Core`, as doing so might break existing imports of the form `open import Categories.Morphism C`.
However, maybe it is best to just make the (slight) breaking change.

### DIFF
--- a/src/Categories/Morphism.agda
+++ b/src/Categories/Morphism.agda
@@ -27,8 +27,8 @@ Mono {A = A} f = ∀ {C} → (g₁ g₂ : C ⇒ A) → f ∘ g₁ ≈ f ∘ g₂
 
 record _↣_ (A B : Obj) : Set (o ⊔ ℓ ⊔ e) where
   field
-    mor : A ⇒ B
-    mono  : Mono mor
+    mor  : A ⇒ B
+    mono : Mono mor
 
 Epi : ∀ (f : A ⇒ B) → Set (o ⊔ ℓ ⊔ e)
 Epi {B = B} f = ∀ {C} → (g₁ g₂ : B ⇒ C) → g₁ ∘ f ≈ g₂ ∘ f → g₁ ≈ g₂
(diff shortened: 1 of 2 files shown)


JacquesCarette:
I like this.  I'm also fine with `.Notation`. I'll wait for others to opine.

src/Categories/Morphism.agda
```diff
@@ -25,9 +25,19 @@ private
 Mono : ∀ (f : A ⇒ B) → Set (o ⊔ ℓ ⊔ e)
 Mono {A = A} f = ∀ {C} → (g₁ g₂ : C ⇒ A) → f ∘ g₁ ≈ f ∘ g₂ → g₁ ≈ g₂
 
+record _↣_ (A B : Obj) : Set (o ⊔ ℓ ⊔ e) where
+  field
+    mor : A ⇒ B
+    mono  : Mono mor
```
sstucki, line 31: Spacing is a bit odd here, no?

src/Categories/Morphism.agda
```diff
@@ -25,9 +25,19 @@ private
...
+
 Epi : ∀ (f : A ⇒ B) → Set (o ⊔ ℓ ⊔ e)
 Epi {B = B} f = ∀ {C} → (g₁ g₂ : B ⇒ C) → g₁ ∘ f ≈ g₂ ∘ f → g₁ ≈ g₂
 
+record _↠_ (A B : Obj) : Set (o ⊔ ℓ ⊔ e) where
+  field
+    mor : A ⇒ B
+    epi  : Epi mor
+
```
sstucki, line 40: Same here.

sstucki (commented):
I like this too.

src/Categories/Morphism.agda
TOTBWF: Whoops! Will fix

TOTBWF:
Is this good to go?

PR #204, opened 2020-09-10
Implement Subobjects, as well as the subobject presheaf

### LINKED ISSUE #203: Impelement Subobjects
See https://ncatlab.org/nlab/show/subobject, or CWM 2nd Edition, V.7 for more context.

Subobjects seem to pop up quite often, so it would be handy to have a definition. The definition using the Over Category on nlab seems especially appealing.

The one tricky bit seems to be the isomorphism classes, but I imagine we ought to be able to construct a `Poset` without _too_ much trouble.

TOTBWF:
This PR defines a notion of subobject, and also implements the subobject presheaf. This PR builds off of #202, so that should get merged before this. This should also close #203.

## Notes
Interestingly, we don't need to say that 𝒞 is well-powered for this to work! I'm a bit of a novice on size issues in general, but I suspect that the notion of "well-poweredness" does not apply for this library.

## References
https://ncatlab.org/nlab/show/subobject
Categories For the Working Mathematician, V.7
Sheaves in Geometry and Logic, I.3

### DIFF
--- a/Everything.agda
+++ b/Everything.agda
@@ -296,6 +296,8 @@ import Categories.Object.Product.Indexed
 import Categories.Object.Product.Indexed.Properties
 import Categories.Object.Product.Limit
 import Categories.Object.Product.Morphisms
+import Categories.Object.Subobject
+import Categories.Object.Subobject.Properties
 import Categories.Object.Terminal
 import Categories.Object.Terminal.Limit
 import Categories.Object.Zero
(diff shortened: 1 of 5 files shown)


src/Categories/Category/Construction/Comma.agda
```diff
@@ -139,3 +139,11 @@ module _ {C : Category o₁ ℓ₁ e₁} {D : Category o₂ ℓ₂ e₂} where
 
   _↘_ : (S : Functor D C) (X : C.Obj) → Category _ _ _
   S ↘ X = S ↓ const! X
+
+module _ (C : Category o₁ ℓ₁ e₁) where
+  private
+    module C = Category C
+
+  -- The slice, or over category
+  _/_ : (X : C.Obj) → Category _ _ _
```
HuStmpHrrr, line 148: there is a standalone slice category. is this preferable? where is it used?

TOTBWF: Ah, I was looking in `Categories.Category.Construction` for this. We should probably move `Slice` in there at some point. Either way, I've updated the code to use `Slice` instead.

TOTBWF:
Rebased on top of current master.

PR #211, opened 2020-10-02
Implement Natural Number Objects, and prove some properties

TOTBWF:
This PR implements [natural number objects](https://ncatlab.org/nlab/show/natural+numbers+object), and also shows that they are Initial F-Algebras for the Maybe Endofunctor.

### DIFF
--- a/src/Categories/Object/NaturalNumber.agda
+++ b/src/Categories/Object/NaturalNumber.agda
@@ -1,0 +1,59 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Categories.Category.Core
+open import Categories.Object.Terminal hiding (up-to-iso)
+
+module Categories.Object.NaturalNumber {o ℓ e} (𝒞 : Category o ℓ e) (𝒞-Terminal : Terminal 𝒞) where
+
+open import Level
+
+open import Categories.Morphism 𝒞
+open import Categories.Morphism.Reasoning 𝒞
+
+open Category 𝒞
+open HomReasoning
+open Equiv
+
+open Terminal 𝒞-Terminal
+
+private
+  variable
+    A B C D X Y Z : Obj
+    h i j : A ⇒ B
+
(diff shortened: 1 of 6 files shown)


src/Categories/Object/NaturalNumber.agda
```diff
@@ -0,0 +1,59 @@
...
+open Equiv
+
+open Terminal 𝒞-Terminal
+
+private
+  variable
+    A B C D X Y Z : Obj
+    h i j : A ⇒ B
+
+record NaturalNumber : Set (o ⊔ ℓ ⊔ e) where
```
HuStmpHrrr, line 24: not sure if would make sense to provide a predicate version, like `IsNaturalNumber`?

TOTBWF: Good point! We should probably do this for all of our `Object`s moving forward 🙂 Perhaps we should add something to the design doc about this?

HuStmpHrrr: let's put some thought into it. in fact, all definitions of objects and diagrams are in predicate style.

TOTBWF: I can think of a few off the top of my head that aren't, but definitely should be (`Pullback` and `Coequalizer` are ones that comes to mind immediately)

HuStmpHrrr: yeah probably `Terminal` is not. `Product` is though. sort of inconsistent indeed.

JacquesCarette:
So should I wait for further commits, or is this ready as is, and will be augmented later?

TOTBWF:
Sorry, life has been getting in the way of Agda recently! I'll add a predicate form of this, and then we ought to be good to go.

JacquesCarette:
I completely understand - my backlog of stuff on here shows the same pattern!

TOTBWF:
Finally got around to tidying this up! Should be good to merge.

JacquesCarette (approved):

PR #213, opened 2020-10-09
SubCategory functors

conal:
Faithful and full functors from subcategories.

I also added `private variable I : Set i` and used it throughout.

### DIFF
--- a/src/Categories/Category/SubCategory.agda
+++ b/src/Categories/Category/SubCategory.agda
@@ -81,14 +81,13 @@ FullSubCategory {I = I} U = record
 
 SubCategoryFunctor : ∀ {ℓ′} (Sub : SubCat {i} {ℓ′} I)
                    → Functor (SubCategory Sub) C
-SubCategoryFunctor Sub = record
+SubCategoryFunctor (record {U = U}) = record
   { F₀ = U
   ; F₁ = proj₁
   ; identity = refl
   ; homomorphism = refl
   ; F-resp-≈ = id→
-  } where
-      open SubCat Sub using (U)
+  }
 
 SubCategoryFaithful : ∀ {ℓ′} (Sub : SubCat {i} {ℓ′} I)
                     → Faithful (SubCategoryFunctor Sub)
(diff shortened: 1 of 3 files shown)


conal:
Maybe some of the explicit `Sub` and `U` should be implicit instead.

JacquesCarette:
We've been trying to split these things differently. Specific functors go under `Functor/Construction` (when they have inputs) and `Functor/Instance` otherwise. Properties of these go in a `/Properties` sub-directory.  This allows shorter names as there is no need to encode the whole context in the name.

Otherwise, this all looks good.

conal:
> We’ve been trying to split these things differently. *\[…\]*

Thanks for the tip. New changes:

  - The functors have a `SubCat` or `I → Obj` input, so I moved the new functionality to a new module `Categories.Functor.Construction.SubCategory`.
  - Shorter names:
      - “`SubCategoryFunctor`” ⟶ “`Sub`”
      - “`SubCategoryFaithful`” ⟶ “`SubFaithful`”
      - “`FullSubCategoryFunctor`” ⟶ “`FullSub`”
      - “`FullSubCategoryFaithful`” ⟶ “`FullSubFaithful`”
      - “`FullSubCategoryFull`” ⟶ “`FullSubFull`”
    I’m not attached to these names, and I guess the remaining “`Sub`” repeats information in the module name. Suggestions welcome.
  - The `U`-valued parameters (to `SubFaithful`, `FullSub`, `FullSubFaithful`, and `FullSubFull`) are now implicit, because I think they can be inferred from the result. The `SubCat`-valued parameters (to `Sub` and `SubFaithful`) are still explicit, because I think they cannot be inferred from the result.
  - The `private variable` lists in `Categories.Category.SubCategory` and `Categories.Functor.Construction.SubCategory` now include `ℓ′ : Level`, allowing removal a few more explicit implicit parameters from signatures.

JacquesCarette (approved):
Sorry this review has taken so long.  Thanks.  Merging.

PR #214, opened 2020-10-19
Use Data.Unit.Polymorphic instead, and drop explicit liftings

conal:
A bit of code simplification/reuse, though you may prefer the current explicit version.


JacquesCarette:
Just to be clear: I authored `Data.Unit.Polymorphic` and `Data.Empty.Polymorphic` with explicitly this in mind. I just had not gotten around to doing this, because I was waiting for stdlib to release.  So thanks!

PR #216, opened 2020-10-30
Rels and dagger categories

Taneb:


JacquesCarette (approved):
Great, thanks.

JacquesCarette:
You marked this WIP - what else are you planning to do ?

Taneb:
I've got a proof that Rels is Cartesian lying around, I didn't know where to put it. I've sinced realised that the equivalent for Sets is in Monoidal.Instances so I'll put it there

That said, I have no issues with this going in as-is

Taneb:
Other things I want to add:

* Discrete categories are dagger
* All groupoids are dagger
* Vector spaces are dagger (this requires a definition of fields and vector spaces in stdlib that doesn't exist yet, so it can probably wait)
* Symmetric monoidal dagger categories and dagger compact categories
* Category of spans (I don't think this has been defined here yet) is dagger

I'd like to build on this a categorical quantum mechanics library but that shouldn't be in this library.

conal:
> * Vector spaces are dagger (this requires a definition of fields and vector spaces in stdlib that doesn't exist yet, so it can probably wait)

Maybe use rings and modules instead. Or semirings and left semimodules.

JacquesCarette:
Sounds great. This does seem ready to go in (though we really ought to start a CHANGELOG to note all the new stuff in new releases!).  You can add the rest incrementally.

Issue #215, opened 2020-10-28
Compiling Categories.Category.Cartesian is very slow


jwiegley:
Today I was looking at https://github.com/conal/agda-cat-linear with @conal , and I was working on getting the Nix-based build working. It uses the current master versions of the `standard-library` and `agda-categories`.

This caused me to do something now, which is to actually compile the code. :-) Nearly all the dependencies compiled fairly quickly, but when it reached `MAlonzo.Code.Categories.Category.Cartesian`, this file alone took more than 30 minutes to compile.

It is fairly easy to reproduce this, it requires a directory with all three projects Git cloned, and then running `nix-build --pure` inside `agda-cat-linear`. Or, enter `nix-shell`, cd into `src`, and type `agda --compile AlgebraicCats.agda`. You'll notice it pause when it reaches that file, and GHC will take 100% of a CPU core for as much of the time as I was willing to watch.

Taneb:
I've taken a little look at this, and discovered that Agda is generating a 162093 line long Haskell file when it compiles `Categories.Category.Cartesian`.

There's a few identifiers that get compiled to over 10000 lines of Haskell each:

* `Categories.Category.Cartesian.Cartesian.monoidal` is 24015 lines

and all of 

* `Categories.Category.Cartesian.Cartesian.monoidal.pentagon`
* `Categories.Category.Cartesian.Cartesian._.Symmetric.pentagon`
* `Categories.Category.Cartesian.Cartesian._.Symmetric.braided.pentagon`
* `Categories.Category.Cartesian.Cartesian.symmetric.pentagon`
* `Categories.Category.Cartesian.Cartesian.symmetric.braided.pentagon`

are 19669 lines. These six identifiers account for over three quarters of the compiled Haskell!

Taneb:
`Categories.Category.Cartesian` isn't the only problem module. Here's all the modules in current master that compile to more than 100000 lines of Haskell

```
  1539839 src/MAlonzo/Code/Categories/Theory/Lawvere.hs
   954948 src/MAlonzo/Code/Categories/Category/CartesianClosed.hs
   856929 src/MAlonzo/Code/Categories/Category/Instance/Properties/Posets.hs
   693640 src/MAlonzo/Code/Categories/Category/Instance/Properties/Cats.hs
   523664 src/MAlonzo/Code/Categories/Category/CartesianClosed/Canonical.hs
   398934 src/MAlonzo/Code/Categories/Category/Cocomplete/Finitely.hs
[...]

PR #218, opened 2020-11-08
Move monoidal properties of cartesian categories into separate modules

sstucki:
This makes some progress towards addressing #215. By separating the monoidal properties of (co-)cartesian categories into separate modules, the overall size of the generated Haskell code seems to be reduced. In particular, there are fewer "copies" of the large proof term of the pentagon identity for cartesian categories that get included in other modules.

Questions that remain open:
 - Why is the Haskell code generated for the proof term of the pentagon identity so big in the first place?
 - Why did it get duplicated (and not just referenced) when the `monoidal` field in the `Cartesian` record was re-exported?

### DIFF
--- a/src/Categories/Category/Cartesian.agda
+++ b/src/Categories/Category/Cartesian.agda
@@ -262,6 +262,11 @@ record Cartesian : Set (levelOfTerm 𝒞) where
   power A 1 = A
   power A (suc (suc n)) = A × power A (suc n)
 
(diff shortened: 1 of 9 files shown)


JacquesCarette (approved):

JacquesCarette:
I think part of the fundamental issue is that Agda doesn't have a true module system. Worse, some of its implementation works by copying, which has serious side-effects. It is well-known to lose a lot of sharing, but I think this shows that it might have other negative side effects.

PR #219, opened 2020-11-09
an n-uniform hypergraph

bblfish:
This is a simple n-uniform hypergraph inspired from other similar categories from the finite sets categories. This should cover any small category with 2 objects and n arrows (including 0) from the first to the second. 

If this is useful I guess it would need splitting into different files, and perhaps more mappings to other categories to test it, ...

One may be able to show that a functor from Hypgergraph to set is equivalent to a functor from Span to Set with the middle node fixed on an object in Set with n elements.

Let me know what improvements would be needed, I'll be happy to work on them.

### DIFF
--- a/src/Categories/Category/Finite/Fin/Instance/HGraph.agda
+++ b/src/Categories/Category/Finite/Fin/Instance/HGraph.agda
@@ -1,0 +1,149 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Category.Finite.Fin.Instance.HGraph where
+
+open import Data.Nat.Base using (ℕ)
+open import Data.Fin.Base using (Fin)
+open import Data.Fin.Patterns
+open import Relation.Binary.PropositionalEquality.Core using (_≡_ ; refl)
+
+open import Categories.Category.Finite.Fin
+open import Categories.Category
+open import Categories.Functor using (Functor; _∘F_) renaming (id to idF)
+open import Categories.Functor.Presheaf
+
+open import Categories.NaturalTransformation using (ntHelper)
+open import Categories.NaturalTransformation.NaturalIsomorphism using (_≃_)
+open import Categories.Adjoint.Equivalence using (⊣Equivalence)
+open import Categories.Adjoint.TwoSided using (_⊣⊢_; withZig)
+import Categories.Morphism as Mor
+
+
+private
+  variable
+    a b c d : Fin 2
+
+x : Fin 3
+x = {!   !}
+
+-- The diagram is the following:
+-- 0 ---n--> 1
+-- with n arrows between 0 and 1
+
+HyperGraphShape : ℕ → FinCatShape
+HyperGraphShape n = shapeHelper record
+  { size      = 2
+  ; ∣_⇒_∣     = morph
+  ; id        = id
+  ; _∘_       = _∘_
+  ; assoc     = assoc
+  ; identityˡ = identityˡ
+  ; identityʳ = identityʳ
+  }
+  where morph : Fin 2 → Fin 2 → ℕ
+        morph 0F 0F = 1
+        morph 0F 1F = n
+        morph 1F 0F = 0
+        morph 1F 1F = 1
+        
+        id : Fin (morph a a)
+        id {0F} = 0F -- 0F is the only inhabitant of Fin 1
+        id {1F} = 0F 
+
+        _∘_ : ∀ {a b c} → Fin (morph b c) → Fin (morph a b) → Fin (morph a c)
+        _∘_ {0F} {0F} {0F} 1₀ 1₀' = 1₀
+        _∘_ {0F} {0F} {1F} g 1₀   = g
+        _∘_ {1F} {1F} {1F} 1₁ 1₁' = 1₁
+        _∘_ {0F} {1F} {1F} 1₁ f   = f
+   
+        assoc : ∀ {f : Fin (morph a b)} {g : Fin (morph b c)} {h : Fin (morph c d)} →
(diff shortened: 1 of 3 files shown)


src/Categories/Category/Finite/Fin/Instance/HGraph.agda
```diff
@@ -0,0 +1,146 @@
...
+        morph 0F 1F = n
+        morph 1F 0F = 0
+        morph 1F 1F = 1
+        
+        id : Fin (morph a a)
+        id {0F} = 0F -- 0F is the only inhabitant of Fin 1
+        id {1F} = 0F 
+
+        _∘_ : ∀ {a b c} → Fin (morph b c) → Fin (morph a b) → Fin (morph a c)
+        _∘_ {0F} {0F} {0F} 1₀ 1₀' = 1₀
```
JacquesCarette, line 51: we tend to use `_` for unused arguments instead of giving them names.

src/Categories/Category/Finite/Fin/Instance/HGraph.agda
```diff
@@ -0,0 +1,146 @@
...
+        _∘_ {0F} {0F} {1F} g 1₀   = g
+        _∘_ {1F} {1F} {1F} 1₁ 1₁' = 1₁
+        _∘_ {0F} {1F} {1F} 1₁ f   = f
+   
+        assoc : ∀ {f : Fin (morph a b)} {g : Fin (morph b c)} {h : Fin (morph c d)} →
```
JacquesCarette, line 56: using private `variables` is great, unless you need to explicitly given them a value (as you do here). Then it is much better to actually give them explicitly (with different names, just to be sure) because of *ordering*. Between large versions of Agda, the algorithm that choose the ordering of insertion of implicit variable changes, causing all such code to break.  So here it is best to make them explicit.

src/Categories/Category/Finite/Fin/Instance/HGraph.agda
```diff
@@ -0,0 +1,146 @@
...
+        identityˡ {0F} {0F} {0F} = refl
+        identityˡ {0F} {1F} {f} = refl
+        identityˡ {1F} {1F} {0F} = refl
+
+        identityʳ : ∀ {f : Fin (morph a b)} → (f ∘ id) ≡ f
+        identityʳ {0F} {0F} {f} = refl
+        identityʳ {0F} {1F} {f} = refl
+        identityʳ {1F} {1F} {0F} = refl
+
+open import Level
```
JacquesCarette, line 74: please put all `open import` at the top of the file.

src/Categories/Category/Finite/Fin/Instance/HGraph.agda
```diff
@@ -0,0 +1,146 @@
...
+
+Δₙ : ℕ → Category 0ℓ 0ℓ 0ℓ
+Δₙ n = FinCategory (HyperGraphShape n)
+
+----- Some Categories
+----- convential names should be found 
+
+open import Categories.Category.Instance.Sets
+
+2GraphSet : (o : Level) → Set (suc o)
```
JacquesCarette, line 84: consider putting these instances in a separate file.

src/Categories/Category/Finite/Fin/Instance/HGraph.agda
```diff
@@ -0,0 +1,146 @@
...
+nGraphSet : ℕ → (o : Level) → Set (suc o)
+nGraphSet n o = Presheaf (Δₙ n) (Sets o)
+
+----- Mapping to other constructions
+
+----- Map Δₙ 2 to Parallel 
+
+open import Categories.Category.Finite.Fin.Instance.Parallel as P
+
+Δ₂ : Category 0ℓ 0ℓ 0ℓ
```
JacquesCarette, line 102: From here below, definitely belongs in separate file(s).

JacquesCarette (changes requested):
Nice, thanks.  There are a few changes needed, but quite straightforward.

src/Categories/Category/Finite/Fin/Instance/HGraph.agda
bblfish: Ah yes. Here I thought it could help the reader to know that they were representing the identity morphism on 0. It gets quite confusing, with objects and morphisms being named by numbers.
But I see that one can argue the other way too :-)

bblfish:
Thanks for the review. I wrote out Δ₂⇒∥ and should probably write out the other direction and to show that Δ₂ and ∥  equivalent. I did not do that, as I was not sure it would really be useful. Perhaps there is something more important to prove. But it's a good exercise anyway.

JacquesCarette:
Are you intending to continue with this PR @bblfish ?

bblfish:
> Are you intending to continue with this PR @bblfish ?

yes, definitely. 

I am taken up a bit work on [Solid](https://solidproject.org/) (see my [github issues](https://github.com/search?q=is%3Aopen+is%3Aissue+author%3Abblfish+archived%3Afalse)) and also got stuck trying to create delta sigma pi functors between delta2 and delta3 to Set functor categories see [agda.zulipchat](https://agda.zulipchat.com/#narrow/stream/259644-newcomers/topic/Help.20proving/near/217746644).

JacquesCarette:
"soon" was now ~3 months ago...  Do you still intend to come back to this?

bblfish:
Oh sorry. I got a funded project from the EU in January to [work on Solid Access Control](https://nlnet.nl/project/SolidControl/). That is taking up most of my time right now. Perhaps better close it, and I can reopen later when I get time again, if someone else did not get around to it.

PR #220, opened 2020-11-10
Add support for monoidal natural transformations and isomorphisms

sstucki:
… and fix the definitions in `Category.Instance.Monoidals` to use the appropriate type of natural iso.

### DIFF
--- a/src/Categories/Category/Instance/Monoidals.agda
+++ b/src/Categories/Category/Instance/Monoidals.agda
@@ -9,7 +9,7 @@ open import Categories.Category.Helper
 open import Categories.Category.Monoidal
 open import Categories.Functor.Monoidal
 open import Categories.Functor.Monoidal.Properties
-open import Categories.NaturalTransformation.NaturalIsomorphism
+open import Categories.NaturalTransformation.NaturalIsomorphism.Monoidal
 
 module _ o ℓ e where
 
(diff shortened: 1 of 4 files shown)


src/Categories/Category/Instance/Monoidals.agda
```diff
@@ -9,44 +9,38 @@ open import Categories.Category.Helper
...
     ; _∘_       = ∘-StrongMonoidal
-    ; assoc     = λ {_ _ _ _ F G H} → associator (M.F F) (M.F G) (M.F H)
-    ; identityˡ = unitorˡ
-    ; identityʳ = unitorʳ
-    ; equiv     = record
-      { refl  = ≃.refl
-      ; sym   = ≃.sym
-      ; trans = ≃.trans
-      }
+    -- NOTE: these η-expanded versions typecheck much faster...
```
JacquesCarette, line 39: I think it would be worth finding ways to document this issue and report it as an issue on the main agda repo. I'm not sure the principal devs are aware of this.  There might be nothing that can be done but without awareness, there is definitely no chance of a fix!

src/Categories/NaturalTransformation/Monoidal.agda
```diff
@@ -0,0 +1,279 @@
...
+  -- To shorten some definitions
+  private _⇛_ = MonoidalNaturalTransformation
+
+  -- Identity and compositions
+
+  module _ {o ℓ e o′ ℓ′ e′}
+           {C : MonoidalCategory o ℓ e} {D : MonoidalCategory o′ ℓ′ e′} where
+
+    private
+      module C = MonoidalCategory C
```
JacquesCarette, line 79: Note that for faster checking (and smaller file sizes), I've been inserting `using` statements on such private module definitions, so that only the needed pieces are copied.

JacquesCarette (approved):
Neither of these comments are need-to-fix, this is good to go.

src/Categories/Category/Instance/Monoidals.agda
sstucki: Yep, I'll open an issue in the Agda repo.

I suspect, though, that the devs will tell me to craft an minimal example, which I won't be able to do (this only seems to happen when the development gets sufficiently complex).

src/Categories/NaturalTransformation/Monoidal.agda
sstucki: What really bugs me: why would that change anything?

Surely, for private definitions, the compiler should be able figure out the dependencies and do this automatically? But I agree that it seems to help sometimes. Is this behavior documented somewhere?

While working on #215 I tried writing a minimal example that exhibits this "bug" (copying terms from a privately re-exported module) but I couldn't see any superfluous terms in the output. So if there's a bug, it doesn't seem to be an obvious one.

So I stopped performing this 'optimization' again. Without being able to trigger the issue explicitly, or finding some documentation of it, it kinda feels like a superstitious ritual...

src/Categories/Category/Instance/Monoidals.agda
sstucki: Reported as  agda/agda#5060.

sstucki: FYI, @andreasabel responded to the issue I opened at agda/agda#5060 a while ago. The issue has been closed now, so I just wanted to point it out in case you hadn't seen the response.

PR #231, opened 2020-11-17
Add a variant of Kelly's monoidal coherence lemmas for bicategories…

sstucki:
… and some other lemmas and support for working with bicategories that was missing (shorthands, reasoning combinators, etc.)

The PR also fixes some inconsistently named lemmas (e.g. `λ-∘ᵥ-◁` was about the right unitor, while `ρ-∘ᵥ-▷` was about the left unitor) and updates dependencies.


JacquesCarette (approved):
Very nice!

JacquesCarette:
I'm going to let Travis finish first.  [Like Agda, we should probably switch to Github Actions, as CI has gone dead slow].

PR #235, opened 2020-12-08
Add a module for Star-Autonomous Categories.

bolt12:
## Summary

This PR adds a module for Star-Autonomous Categories to `agda-categories`.

## Intro

I've been working on this and asking help on Agda Zulip. It seemed a nice, simple addition to the library and a good excuse to learn about its ins&outs. From my understanding Star-Autonomous Categories can have 2 equivalent definitions:

## Definition

 - One where a Star-Autonomous Category is a Symmetric Monoidal Closed Category with a ⊥ object (also called global dualizing object) such that for every object A and ⊥, there exists a morphism  A → ( A ⊸⊥) ⊸⊥ that is an isomorphism for all A.

- One where a Star-Autonomous Category is a Symmetric Monoidal Category equipped with a full and faithful functor (_)*: C^op ⟶ C, such that there is a natural isomorphism C(A⊗B,D*)≃C(A,(B⊗D)*). Here nLab and Wikipedia differ a little. Wikipedia says that there's also for every object A a Natural Isomorphism A ≅ A**, and that the previous Natural Isomorphism is actually a Natural Bijection.

## Code

I want to understand what changes/improvements do I need to make in order for this PR to be accepted. If you think that there's no need for Star-Autonomous Categories in `agda-categories` that's also fine by me, but I'd appreciate if you could give me any feedback, since I'm just learning and would like to improve!

Although both definitions are equivalent, I decided to go for the "weaker" definition (the one that only required a Symmetric Monoidal Category) because it seemed to be the simpler to define in Agda. However, I struggled a lot in making the conversion from the mathematical definitions seen on nLab and Wikipedia to the actual Agda code, specially wrt to the Natural Isomorphism between the Hom Functors. Luckily everyone over at Zulip was able to guide me through what could be a possible solution!

Regarding the Natural Isomorphism between the Hom Functors, I composed them with `Hom[-,-]` since Wikipedia asks for a Natural Bijection. I can give more explanations about the code but I'd rather wait for someone's opinion and suggestion first!

# References

- https://ncatlab.org/nlab/show/star-autonomous+category
- https://en.wikipedia.org/wiki/*-autonomous_category


src/Categories/Category/Monoidal/Star-Autonomous.agda
```diff
@@ -0,0 +1,35 @@
...
+open Category C renaming (op to Cᵒᵖ) hiding (id)
+open Monoidal M
+open Functor ⊗ renaming (op to ⊗ₒₚ)
+
+open import Relation.Binary.Structures
+
+record Star-Autonomous : Set (levelOfTerm M) where
+  field
+    symmetric : Symmetric M
+    F⁻¹ : Functor Cᵒᵖ C
```
laMudri, line 28: I would call `F⁻¹` something like `star` or `dual` (with capitalisation dependent on agda-categories conventions for functors).

src/Categories/Category/Monoidal/Star-Autonomous.agda
```diff
@@ -0,0 +1,35 @@
...
+
+  open Functor F⁻¹ renaming (op to F⁻¹ₒₚ ; F₀ to F⁻¹₀)
+
+  field
+    FFF⁻¹ : FullyFaithful F⁻¹
+    A≃A⁻¹⁻¹ : id ≃ (F⁻¹ ∘F F⁻¹ₒₚ)
```
laMudri, line 34: I think this would make more sense being called `A≃A**`, matching conventional notation. The star is a dual rather than an inverse. Similar comment for `𝒞[A⊗B,C⁻¹]≃𝒞[A,B⊗C⁻¹]`.

src/Categories/Category/Monoidal/Star-Autonomous.agda
bolt12: That makes more sense! Thanks!

src/Categories/Category/Monoidal/Star-Autonomous.agda
bolt12: @laMudri is there any way to write superscript `*` in Agda?

laMudri: Oh, I forgot that there are fonts where ASCII `*` doesn't appear as a superscript already.

bolt12: I've used normal `*` in the code. What do you think?

laMudri: It always looked fine for me. The question is whether anyone else would object. I want to say “eh, it'll be fine”, but I can't really judge that while I always use fonts that put `*` in the superscript position.

src/Categories/Category/Monoidal/Star-Autonomous.agda
```diff
@@ -0,0 +1,35 @@
...
+open Functor ⊗ renaming (op to ⊗ₒₚ)
+
+open import Relation.Binary.Structures
+
+record Star-Autonomous : Set (levelOfTerm M) where
+  field
+    symmetric : Symmetric M
+    Star : Functor Cᵒᵖ C
+
+  open Functor Star renaming (op to Starₒₚ ; F₀ to Star₀)
```
laMudri, line 30: Some stuff like this should be opened `public`, or somehow available outside without manually opening `Functor` modules and stuff. This might be the point where you rename `F₀` to `_*₀` and `F₁` to `_*₁`, but all the properties should be exported somehow too.

laMudri:
A natural proof to put inside the record would be that (A ⊗ B*)* is an internal hom, so a *-autonomous category is monoidally closed. I guess together with this goes the fact that I* is the dualising object (IIRC). To be honest, we probably eventually want both of the definitions you mentioned at the start, and coercions between them.

src/Categories/Category/Monoidal/Star-Autonomous.agda
bolt12: Okay! Let's see what other people say :)

src/Categories/Category/Monoidal/Star-Autonomous.agda
bolt12: I see! I realized the second renaming is not used so I've deleted it. Also I'm not sure how to manually open  `Star` functor outside since it is defined.

bolt12:
@laMudri yes that would be interesting to have. What do you think is the best way to do that? Have both in the same module? Or have another `Star-Autonomous` module under `Category.Monoidal.Closed` ?

src/Categories/Category/Monoidal/Star-Autonomous.agda
JacquesCarette: It's fine with me.

JacquesCarette:
As far as the 2 definitions go: I would somehow be tempted to call the first one `Has-⊥-Object`, and then prove that an SMCC with that extra property is *-autonomous (and vice versa).

src/Categories/Category/Monoidal/Star-Autonomous.agda
```diff
@@ -0,0 +1,35 @@
...
+
+open Category C renaming (op to Cᵒᵖ) hiding (id)
+open Monoidal M
+open Functor ⊗ renaming (op to ⊗ₒₚ)
+
+open import Relation.Binary.Structures
+
+record Star-Autonomous : Set (levelOfTerm M) where
+  field
+    symmetric : Symmetric M
```
JacquesCarette, line 27: I would think that you'd want this `Symmetric` to be part of the module parameters too? Are there cases where a single monoidal category has different `Symmetric` witnesses that lead to different Star-Autonomous structures?

src/Categories/Category/Monoidal/Star-Autonomous.agda
```diff
@@ -0,0 +1,35 @@
...
+    Star : Functor Cᵒᵖ C
+
+  open Functor Star renaming (op to Starₒₚ ; F₀ to Star₀)
+
+  field
+    FF-Star : FullyFaithful Star
+    A≃A**  : id ≃ (Star ∘F Starₒₚ)
```
JacquesCarette, line 34: For chaining of reasoning, it's usually best to have things the other way around, i.e. `A**≃A` instead and similarly swapping the rhs.

JacquesCarette (changes requested):
almost ready!

laMudri:
By the way, we were quite careful on Zulip about stating the natural isomorphisms, but `Categories.Category.Monoidal.Core` seems to do something different. In fact, it's not clear to me where naturality of `unitorˡ`, `unitorʳ`, and `associator` come from. Is there a clever design trick there which would be applicable to the natural isomorphisms we need here?

JacquesCarette:
I think some of those things were 'unbundled' in `Categories.Category.Monoidal.Core`. I think some of those things are then proved in `Categories.Category.Monoidal.Properties`.

src/Categories/Category/Monoidal/Star-Autonomous.agda
bolt12: Great! :smile:

bolt12:
> As far as the 2 definitions go: I would somehow be tempted to call the first one `Has-⊥-Object`, and then prove that an SMCC with that extra property is *-autonomous (and vice versa).

I like that approach!

src/Categories/Category/Monoidal/Star-Autonomous.agda
bolt12: I'm not sure, I just opted to do things this way because after looking a bit for inspiration I found that Compact Closed Categories did have a field called `symmetric`. I'll gladly take your suggestion if you think that's the best approach :smile:

src/Categories/Category/Monoidal/Star-Autonomous.agda
bolt12: Got it! :+1:

src/Categories/Category/Monoidal/Star-Autonomous.agda
bolt12: Also would that change imply I move the module to under `Monoidal.Symmetric` ?

laMudri:
Aah, that's what `unitorˡ-commute-from` &c are about, with things bundled together in `Categories.Category.Monoidal.Utilities`. It looks like you eventually have to do the same amount of work, so maybe it doesn't matter? Naïvely, if I were trying to build an instance, I think I'd prefer the bundled version.

src/Categories/Category/Monoidal/Star-Autonomous.agda
JacquesCarette: "best" is definitely open for debate! But for Star-Autonomous always assume SMCC, then yes to both.

JacquesCarette:
@HuStmpHrrr did this particular bit of unbundling, so he'd be the best one to recall the finer details of why it was done that way.

Sometimes, doing the reasoning "at the lower level", which is what happens to clients of this interface, ends up being quite a bit easier **in Agda**. In some sense, it is not as 'categorical'.  If I'm going to speculate a bit, I'd say that it is because in agda-categories, the tools for doing reasoning at the level of NaturalIsomorphism (and simply for Object Isos) are not as fully developed and fully mature. There is nothing wrong/bad with the tools, but it hasn't been put through enough use to fully develop.  That could be a fun project.

bolt12:
I see. I'm not sure which way I prefer since I'm still learning how to use the library but I'd rather leave it as it is for now and then unbundle if asked/required.

src/Categories/Category/Monoidal/Star-Autonomous.agda
JacquesCarette: Yes, I think so.

bolt12:
@JacquesCarette is there anything that I should change?

bolt12:
I think I should update Everything.agda

JacquesCarette (approved):

JacquesCarette:
Sorry, this slipped under my radar. Yes, please add to `Everything.agda`, but otherwise it's ready to go.

bolt12:
@JacquesCarette Done! Thank you so much for all the help!

HuStmpHrrr:
sorry to jam in so late. yes, we split the definition of naturality apart precisely because we want to avoid associativity:
```
Hom[-,-] ∘F (⊗ₒₚ ⁂ Star) ≃ Hom[-,-] ∘F (id ⁂ (Star ∘F ⊗ₒₚ)) ∘F assocˡ _ _ _
```

there are two reasons to avoid it:

1. it's not reflected in classical settings.
2. once you try to prove it, your prove obligation will involve composing `id`, which is quite annoying because it's some unnecessary book keeping to do.

I also feel skeptical about

```
{o ℓ e} {C : Category o ℓ e} {M : Monoidal C} (S : Symmetric M)
```

one thing to think about this predicate form is that, do we have a chance where we have the same `Monoidal` structure, but different `Symmetric`? I think it's probably very unlikely, so it would probably the best if `Symmetric` is part of the definition, so that the chain of this predicates can be shorter.

bolt12:
Hi @HuStmpHrrr, thanks for your input!

> we split the definition of naturality apart precisely because we want to avoid associativity

Could you please point me where to look to improve this definition without using associativity? I'd be more than happy to change it to something more straightforward!

Regarding the `Symmetric` issue, unfortunately I do not have a strong opinion since I'm not experienced enough to know the full implications of each approach. Once again, I'm more than happy to make your suggestion

HuStmpHrrr:
monoidal category is an example:

https://github.com/agda/agda-categories/blob/master/src/Categories/Category/Monoidal/Core.agda

consider the left unitor. we define the unitor in two parts:

1. we first require an isomorphism:
https://github.com/agda/agda-categories/blob/f29629f90c373e41b9589f39a00e69fad549394b/src/Categories/Category/Monoidal/Core.agda#L71
this is the isomorphic part.
2. then we require two naturality squares:
https://github.com/agda/agda-categories/blob/f29629f90c373e41b9589f39a00e69fad549394b/src/Categories/Category/Monoidal/Core.agda#L94-L95
this is the natural part.

so natural isomorphisms sometimes are split into these two parts. sometimes it's not quite necessary while sometimes it makes sense. in the monoidal case, we do it because we write `1 ⊗ X ≃ X` on paper, so this separation in some sense is more intuitive. on other cases, it is more helpful to give the natural isomorphisms as a whole, e.g. https://github.com/agda/agda-categories/blob/master/src/Categories/Adjoint/TwoSided.agda.

bolt12:
I see, thanks for all the references! Wrt to Star-Autonomous Categories, I've seen the natural transformation written in terms of the action on the objects (like in the monoidal case), but I also have seen it written in terms of the NI as a whole.

I agree that the associativity bit can be avoided, so I'll refactor it into two parts and see how it turns out :)

bolt12:
Having the first half right, can you help me getting the commutative square for setoids right? @laMudri @JacquesCarette  @HuStmpHrrr 

```agda
record Star-Autonomous : Set (levelOfTerm S) where
  field
    Star : Functor Cᵒᵖ C

  open Functor Star renaming (op to Starₒₚ; F₀ to Star₀; F₁ to Star₁) public
  open Functor Hom[-,-] renaming (F₁ to Hom[-,-]₁)

  field
    FF-Star : FullyFaithful Star
    A**≃A  : (Star ∘F Starₒₚ) ≃ id
    -- 𝒞[A⊗B,C*]≃𝒞[A,B⊗C*] : Hom[-,-] ∘F (⊗ₒₚ ⁂ Star) ≃ Hom[-,-] ∘F (id ⁂ (Star ∘F ⊗ₒₚ)) ∘F assocˡ _ _ _
    𝒞[A⊗B,C*]≃𝒞[A,B⊗C*] : ∀ {A B C} -> Hom[ A ⊗₀ B , Star₀ C ] ≅ Hom[ A , Star₀ (B ⊗₀ C) ]

  module 𝒞[A⊗B,C*]≃𝒞[A,B⊗C*] {A} {B} {C} = _≅_ (𝒞[A⊗B,C*]≃𝒞[A,B⊗C*] {A} {B} {C})

  private
    η⇒ = 𝒞[A⊗B,C*]≃𝒞[A,B⊗C*].from
    η⇐ = 𝒞[A⊗B,C*]≃𝒞[A,B⊗C*].to

  field
    𝒞[A⊗B,C*]≃𝒞[A,B⊗C*]-commute-from : ∀ {f g h} -> {! ? ≈ ? !} -- Commutative Square with Setoid?
```

I'm struggling getting the intuition right. I tried mimicking the monoidal categories approach but without success.

HuStmpHrrr:
I guess let's address the predicate part so that PR can get merged. i don't find the isomorphism part is that important.

bolt12:
> I guess let's address the predicate part so that PR can get merged. i don't find the isomorphism part is that important.

What do you mean with "the predicate part" ?

HuStmpHrrr:
For the record, I think this diagram says what the naturality condition should be:

IMG_20201222_120944

however, looking at it, I think there is a more concise and elegant formulation.

since this isomorphism is natural is all A, B, and C. I found that the naturality of A and C is easily achieved, if we formulate this isomorphism as an adjunction. so we have `L = - ⊗ B` and `R = (B ⊗ -)*`. the isomorphism and naturality follow.

the naturality of `B` is achieved by `Mate`.: https://github.com/agda/agda-categories/blob/master/src/Categories/Adjoint/Mate.agda. so this pattern turns out to be very similar to closed monoidal categories: https://github.com/agda/agda-categories/blob/master/src/Categories/Category/Monoidal/Closed.agda

HuStmpHrrr:
> 
> 
> > I guess let's address the predicate part so that PR can get merged. i don't find the isomorphism part is that important.
> 
> What do you mean with "the predicate part" ?

the predicate part means that `Symmetric` should be part of the definition of `*-Autonomous`, instead of in the predicate position.

bolt12:
I see! Thank you for your insights! I'll resolve the predicate issue straight away!

PR #236, opened 2020-12-29
Define Kernels and Normal Monomorphisms

TOTBWF:
## Patch Description
This PR implements Kernels and Normal Monomorphisms, which moves us closer to being able to use `agda-categories` to do some group theory!

## Notes
I took this opportunity to split up equalizers into a bundled and predicate form, as this is going to be useful when we start working with regular morphisms (of which normal morphisms are a subset).

## References:
https://ncatlab.org/nlab/show/kernel
https://ncatlab.org/nlab/show/normal+monomorphism


JacquesCarette (approved):
Very nice - I really like the style.

Note that I would have preferred 4 PRs: one for splitting Equalizer, one for Kernel (and its properties), one for Normal Morphism, and one for the additions to Object.Zero.

TOTBWF:
Thanks, will do in the future!

PR #239, opened 2021-01-03
Show that we can build Equivalences of Categories from Pointwise Isomorphisms

TOTBWF:
# Patch Description

This PR adds 2 helper functions for building various incarnations of natural isomorphisms from pointwise equivalences.
This is sorely needed for my proofs of the various monadicity theorems, as otherwise we get bogged down in the details incredibly quickly.

# Notes
I'm not super happy with the sigma types present in the definition of `pointwise-iso-equivalence`. Perhaps we ought to add something to `Categories.Morphism` that captures the notion of a morphism being one half of some isomorphism. It seems like this situation is often referred to as "f is an isomorphim" which suggests this type ought to be called `IsIso`, but that conflicts with our naming scheme for predicates.

### DIFF
--- a/src/Categories/Category/Equivalence/Properties.agda
+++ b/src/Categories/Category/Equivalence/Properties.agda
@@ -6,10 +6,14 @@ module Categories.Category.Equivalence.Properties where
 
 open import Level
 
+open import Data.Product using (Σ-syntax; _,_; proj₁)
+
 open import Categories.Adjoint.Equivalence using (⊣Equivalence)
+open import Categories.Adjoint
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):
Some these lines are getting a tad long. In general, it would be better to lay out those statements over 2 lines (or more!) for easier scanning.

TOTBWF:
Yeah, I should probably do a formatting pass over the Crude Monadicity Stuff as well. I tried to compress as best I could but some of them are just gnarly

PR #241, opened 2021-01-20
Inverse category

DreamLinuxer:


JacquesCarette (commented):
This is certainly more compact, but is it a good idea? This isn't really weaker either? The old one reads like there's a distinguished inverse, but that's also what the new one ends up saying, no?

DreamLinuxer:
> This is certainly more compact, but is it a good idea? This isn't really weaker either? The old one reads like there's a distinguished inverse, but that's also what the new one ends up saying, no?

The old one misses the uniqueness requirement. The new one fixes this.

JacquesCarette:
All of the other similar structures have a separate field, `unique` for that. This change would make this one 'stick out' by bundling everything together into a single field, instead of having 4. I definitely prefer the more-fields approach.

JacquesCarette (approved):

PR #246, opened 2021-02-10
The Bicategory of Spans

TOTBWF:
This PR implements the Bicategory of Spans, which gives us an idea of what building generalized multicategories will look like, as the construction relies on building the bicategory of T-Spans.

## Notes
This one was pretty brutal. I couldn't figure out any clever tricks for getting out of the pullback soup that you find yourselves in for the associativity proofs.

## References
* https://ncatlab.org/nlab/show/span#categories_of_spans
* https://ncatlab.org/nlab/show/generalized+multicategory

### DIFF
--- a/src/Categories/Bicategory/Construction/Spans.agda
+++ b/src/Categories/Bicategory/Construction/Spans.agda
@@ -3,7 +3,8 @@
 open import Categories.Category
 open import Categories.Diagram.Pullback
 
-module Categories.Bicategory.Construction.Spans {o ℓ e} (𝒞 : Category o ℓ e) (has-pullbacks : ∀ {X Y Z} (f : 𝒞 [ X , Z ]) (g : 𝒞 [ Y , Z ]) → Pullback 𝒞 f g) where
+module Categories.Bicategory.Construction.Spans {o ℓ e} {𝒞 : Category o ℓ e}
+                                                (_×ₚ_ : ∀ {X Y Z} → (f : 𝒞 [ X , Z ]) (g : 𝒞 [ Y , Z ]) → Pullback 𝒞 f g) where
 
 open import Level
 open import Function using (_$_)
@@ -13,6 +14,7 @@ open import Data.Product using (Σ; _×_; _,_; proj₁; proj₂; uncurry)
 open import Categories.Bicategory
 open import Categories.NaturalTransformation.NaturalIsomorphism hiding (refl; sym)
 
+open import Categories.Category.Diagram.Span 𝒞
 open import Categories.Category.Construction.Spans 𝒞 renaming (Spans to Spans₁)
 
 open import Categories.Morphism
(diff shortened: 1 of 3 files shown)


src/Categories/Bicategory/Construction/Spans.agda
```diff
@@ -0,0 +1,406 @@
...
+open Pullback
+
+private
+  variable
+    A B C D E : Obj
+
+  --------------------------------------------------------------------------------
+  -- Various Spans, Span Maps, and their compositions
+  
+  id-span : Span A A
```
JacquesCarette, line 39: Shouldn't the next 3 constructions be defined in `Categories.Category.Construction.Spans` ?  Actually, let me correct that: in `Categories.Category.Diagram.Span`.

src/Categories/Bicategory/Construction/Spans.agda
```diff
@@ -0,0 +1,406 @@
...
+        dom g′ ∘ arr β ∘ p₁ pullback                    ≈⟨ pullˡ (commute-dom β) ⟩
+        dom g ∘ p₁ pullback                             ∎
+      ; commute-cod = begin
+        (cod f′ ∘ p₂ pullback′) ∘ universal pullback′ _ ≈⟨ pullʳ (p₂∘universal≈h₂ pullback′) ⟩
+        cod f′ ∘ arr α ∘ p₂ pullback                    ≈⟨ pullˡ (commute-cod α) ⟩
+        cod f ∘ p₂ pullback                             ∎
+      }
+  
+  --------------------------------------------------------------------------------
+  -- Proofs about span compositions
```
JacquesCarette, line 76: There proofs should probably be defined as properties of `Span` rather than 'buried' here.

src/Categories/Category/Construction/Spans.agda
```diff
@@ -0,0 +1,64 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Categories.Category
+
+-- The 1-Category of Spans
+module Categories.Category.Construction.Spans {o ℓ e} (𝒞 : Category o ℓ e) where
```
JacquesCarette, line 6: So there is something related at `Categories.Category.Instance.Span`. It's weird because it is really a "span diagram" that is defined there, whereas here is the more usual construction of a span in a category. That's definitely a bit of a mess (but I'm not going to require you to clean that part up as part of this!)

However, your `Span` here is a proper categorical construction / diagram that should exist in `Diagram.Span` and then `Spans` should indeed be defined here.

JacquesCarette (changes requested):
Very nice. Just a few things to tweak.

Taneb:
I'm excited to see this. The category of spans is something I was looking at for my work on implementing dagger category theory in this library, the category of spans on a category with pullbacks is a good example of a dagger category. I'd been trying to implement this category without much success.

Once this is merged I'll prove that it is indeed a dagger category.

src/Categories/Bicategory/Construction/Spans.agda
TOTBWF: I could see `id-span` belonging there, but the other two constructions require the existence of pullbacks, so I think it may make more sense to have them live alongside the bicategory

src/Categories/Bicategory/Construction/Spans.agda
TOTBWF: Similar to the two composition primitives, all of these properties rely on the existence of pullbacks, which makes them a bit awkward to use separate from the bicategory instance

src/Categories/Category/Construction/Spans.agda
TOTBWF: Definitely agree that the naming is confusing! Will restructure things 🙂

src/Categories/Bicategory/Construction/Spans.agda
JacquesCarette: Well, you're right that this would all be conditional on some pullbacks existing - but that is part and parcel of spans composing, isn't it? So anyone wanting to compose spans will assume that those pullbacks exist. And then may end up rewriting the exact same functions.

In other words, I still believe they belong in `Categories.Category.Diagram.Span`, but in a sub-module with a `has-pullbacks` exactly like the one here.

src/Categories/Category/Construction/Spans.agda
JacquesCarette: Thanks. To be clear: the 1-category you define does belong exactly in this file, it's the definition of `Span` and `Span⇒` that doesn't. And while you're at it, may as well give names to the identity and composition.

Also, now that I'm looking even more deeply: why `dom` and `cod` ? I know that makes sense in a lot of contexts where spans are used, but in general? [I'm not immediately wanting to change them, rather to be convinced that these are good names.]

TOTBWF: My mental model and main motivation for these things is that "Categories are Monads in Span(C)", so those names were chosen accordingly. I do agree that these names are a bit tied to that example, and may not suit other applications. If anyone has better names I'd love to hear them 🙂

TOTBWF:
Is this good to go? I have some stuff that is waiting on this (Categories as Monads in Span(Set))

JacquesCarette (approved):
Good to go.

PR #248, opened 2021-02-17
Show that Monads in Span(Setoid) are categories

TOTBWF:
## Patch Description

This PR proves that Monads in Span(Setoid) are categories. To do so, we also need to define Monads in a bicategory, which has been done here as well.

I've also defined a nicer implementation of pullbacks in Setoid that are easier to work with than those obtained via completeness.

### DIFF
--- a/src/Categories/Bicategory.agda
+++ b/src/Categories/Bicategory.agda
@@ -63,6 +63,29 @@ record Bicategory o ℓ e t : Set (suc (o ⊔ ℓ ⊔ e ⊔ t)) where
   _▷_ : {A B C : Obj} {f g : A ⇒₁ B} (h : B ⇒₁ C) (α : f ⇒₂ g) → h ∘ₕ f ⇒₂ h ∘ₕ g
   _ ▷ α = id₂ ⊚₁ α
 
+  -- Coherence morphisms
+  unitˡ⇒ : {A B : Obj} {f : A ⇒₁ B} → id₁ ⊚₀ f ⇒₂ f
+  unitˡ⇒ {_} {_} {f} = NaturalIsomorphism.⇒.η unitˡ (_ , f)
+
+  unitˡ⇐ : {A B : Obj} {f : A ⇒₁ B} → f ⇒₂ id₁ ⊚₀ f
+  unitˡ⇐ {_} {_} {f} = NaturalIsomorphism.⇐.η unitˡ (_ , f)
+
+  unitʳ⇒ : {A B : Obj} {f : A ⇒₁ B} → f ⊚₀ id₁ ⇒₂ f
+  unitʳ⇒ {_} {_} {f} = NaturalIsomorphism.⇒.η unitʳ (f , _)
+
+  unitʳ⇐ : {A B : Obj} {f : A ⇒₁ B} → f ⇒₂ f ⊚₀ id₁
+  unitʳ⇐ {_} {_} {f} = NaturalIsomorphism.⇐.η unitʳ (f , _)
+
+  assoc⇒ : {A B C D : Obj} {f : D ⇒₁ B} {g : C ⇒₁ D} {h : A ⇒₁ C} →
+            ((f ⊚₀ g) ⊚₀ h) ⇒₂ (f ⊚₀ (g ⊚₀ h))
+  assoc⇒ {_} {_} {_} {_} {f} {g} {h} = NaturalIsomorphism.⇒.η ⊚-assoc ((f , g) , h)
+
+  assoc⇐ : {A B C D : Obj} {f : D ⇒₁ B} {g : C ⇒₁ D} {h : A ⇒₁ C} →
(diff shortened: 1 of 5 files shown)


src/Categories/Bicategory.agda
```diff
@@ -63,6 +63,29 @@ record Bicategory o ℓ e t : Set (suc (o ⊔ ℓ ⊔ e ⊔ t)) where
   _▷_ : {A B C : Obj} {f g : A ⇒₁ B} (h : B ⇒₁ C) (α : f ⇒₂ g) → h ∘ₕ f ⇒₂ h ∘ₕ g
   _ ▷ α = id₂ ⊚₁ α
 
+  -- Coherence morphisms
```
sstucki, line 66: There are variants of these in [Bicategory.Extras](https://github.com/agda/agda-categories/blob/1ba82d98faddc750fe4c0363af0d1e21292a9085/src/Categories/Bicategory/Extras.agda#L38), including shorthands. Probably not a good idea to duplicate them.

src/Categories/Bicategory/Monad.agda
```diff
@@ -0,0 +1,26 @@
...
+module Categories.Bicategory.Monad where
+
+open import Level
+open import Data.Product using (_,_)
+
+open import Categories.Bicategory
+open import Categories.NaturalTransformation.NaturalIsomorphism using (NaturalIsomorphism)
+
+
+record Monad {o ℓ e t} (𝒞 : Bicategory o ℓ e t) : Set (o ⊔ ℓ ⊔ e ⊔ t) where
```
sstucki, line 14: Excellent!
Maybe we should show that the definition in [Categories.Monad](https://github.com/agda/agda-categories/blob/1ba82d98faddc750fe4c0363af0d1e21292a9085/src/Categories/Monad.agda#L13) is an instance in Cat? Or even replace the latter? (Not asking you to do this, just proposing it as a future addition.)

sstucki (commented):
Cool!

src/Categories/Bicategory/Monad.agda
TOTBWF: I think showing an equivalence is a good call, the more elementary version of monad is far easier to work with, so we probably don't want to replace it :)

TOTBWF: Turns out this was pretty easy! I've added a pair of proofs going both directions to the PR.

src/Categories/Category/Instance/Properties/Setoids/Limits/Canonical.agda
```diff
@@ -0,0 +1,61 @@
+{-# OPTIONS --without-K --safe #-}
+
+-- A "canonical" presentation of limits in Setoid.
+--
+-- These limits are obviously isomorphic to those created by
```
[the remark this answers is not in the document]
sstucki: Splendid!

src/Categories/Bicategory/Monad.agda
sstucki: Brilliant, thanks!

src/Categories/Category/Instance/Properties/Setoids/Limits/Canonical.agda
```diff
@@ -0,0 +1,62 @@
...
+open import Categories.Diagram.Pullback
+
+open import Categories.Category.Instance.Setoids
+
+open Setoid renaming (_≈_ to [_][_≈_])
+
+--------------------------------------------------------------------------------
+-- Pullbacks
+
+record FiberProduct {o ℓ} {X Y Z : Setoid o ℓ} (f : X ⟶ Z) (g : Y ⟶ Z) : Set (o ⊔ ℓ) where
```
JacquesCarette, line 28: In a subsequent PR, I think it would make sense to promote `FiberProduct` to be something less buried. I'm needing it elsewhere, and this seems like a odd place to dig it out of!

src/Categories/Category/Instance/Properties/Setoids/Limits/Canonical.agda
```diff
@@ -0,0 +1,62 @@
...
+  field
+    elem₁ : Carrier X
+    elem₂ : Carrier Y
+    commute : [ Z ][ f ⟨$⟩ elem₁ ≈ g ⟨$⟩ elem₂ ]
+
+open FiberProduct
+
+pullback : ∀ (o ℓ : Level) {X Y Z : Setoid (o ⊔ ℓ) ℓ} → (f : X ⟶ Z) → (g : Y ⟶ Z) → Pullback (Setoids (o ⊔ ℓ) ℓ) f g
```
JacquesCarette, line 36: strictly speaking, this property can be extracted from the fact that `Setoids` is Complete. However, I do think the explicit construction is worthwhile having on its own too.

JacquesCarette (approved):
The elegance of this code is striking. The fact that Monad in a bicategory is so easily defined in a way that looks quite natural, is very comforting. It says that other parts really do seem to compose well.

PR #252, opened 2021-02-23
Refactor the 'Categories.Object.Zero' module

TOTBWF:
As per @HuStmpHrrr's suggestion on #249, I've reworked the definition of zero objects to use `IsTerminal` and `IsInitial`. I also made a couple of fixes to some rather silly mistakes I made when writing this module:
- I've added a predicate version of `Zero`
- I've renamed `zero` to `𝟘`, to avoid conflicts with `zero` from `Level`, `Data.Nat`, and many others!
- I've flipped `¡` and `!` so that their types now make sense:
  `¡ : 𝟘 ⇒ A`
  `! : A ⇒ 𝟘`

While I was at it, I added some autoduality proofs about zero objects that I need for cokernels.

### DIFF
--- a/src/Categories/Object/Duality.agda
+++ b/src/Categories/Object/Duality.agda
@@ -17,6 +17,7 @@ open import Categories.Object.Initial C
 open import Categories.Object.Product op
 open import Categories.Object.Coproduct C
 
+open import Categories.Object.Zero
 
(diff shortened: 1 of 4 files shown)


HuStmpHrrr:
Wonderful. That's very cool.

JacquesCarette (approved):

themar7777 (approved):

PR #253, opened 2021-02-25
Define Cokernels

TOTBWF:
This patch defines Cokernels, and also proves some facts about duality with kernels. I did a minor refactoring in the `Categories.Object.Duality` module, by pulling some implicits into variables.


JacquesCarette (approved):
Perfect.

PR #254, opened 2021-02-27
Implement a basic category tactic

TOTBWF:
All of this work on abelian categories/biproducts has me drowning in some really horrible reassociations, so I figured I'd bite the bullet and just port the stdlibs monoid solver to work with categories. Right now it should handle arbitrary reassociations/identity morphisms, but doesn't handle rewriting equalities anywhere.

Here's an example of it's use:
```
{-# OPTIONS --without-K --safe #-}

module Experiments.Tactics {o ℓ e} {𝒞 : Category o ℓ e} where

open import Categories.Tactic.Category using (solve; solve-macro)

open Category 𝒞
open HomReasoning

private
  variable
    A B C : Obj
    f g   : A ⇒ B

test : ((id ∘ f) ∘ id) ∘ (id ∘ g) ≈ f ∘ g
test {f = f} {g = g} = solve 𝒞
```

It also works inside of equational reasoning blocks as well.

### DIFF
--- a/src/Categories/Tactic/Category.agda
+++ b/src/Categories/Tactic/Category.agda
@@ -23,14 +23,11 @@ open import Reflection.Argument
 open import Reflection.Term using (getName; _⋯⟅∷⟆_)
 open import Reflection.TypeChecking.Monad.Syntax
 
-import Categories.Morphism.Reasoning as MR
-
 module _ {o ℓ e} (𝒞 : Category o ℓ e) where
 
   open Category 𝒞
   open HomReasoning
   open Equiv
-  open MR 𝒞
 
   private
     variable
@@ -70,7 +67,8 @@ module _ {o ℓ e} (𝒞 : Category o ℓ e) where
   preserves-≈′ (f ∘′ g) h = begin
     embed (f ∘′ g) id ∘ h         ≡⟨⟩
     embed f (embed g id) ∘ h      ≈˘⟨ preserves-≈′ f (embed g id) ⟩∘⟨refl ⟩
-    (embed f id ∘ embed g id) ∘ h ≈⟨ pullʳ (preserves-≈′ g h) ⟩
+    (embed f id ∘ embed g id) ∘ h ≈⟨ assoc  ⟩
+    embed f id ∘ embed g id ∘ h   ≈⟨ refl⟩∘⟨ preserves-≈′ g h ⟩
     embed f id ∘ embed g h        ≈⟨ preserves-≈′ f (embed g h) ⟩
     embed (f ∘′ g) h              ∎
   
--- a/src/Categories/Tactic/Category.agda
+++ b/src/Categories/Tactic/Category.agda
@@ -1,0 +1,186 @@
+{-# OPTIONS --without-K --safe #-}
+
+--------------------------------------------------------------------------------
+-- A simple reflection based solver for categories.
+--
+-- Based off 'Tactic.MonoidSolver' from 'agda-stdlib'
+--------------------------------------------------------------------------------
+
+open import Categories.Category
+
+module Categories.Tactic.Category where
+
+open import Level
+open import Function using (_⟨_⟩_)
+
+open import Data.Bool    as Bool    using (Bool; _∨_; if_then_else_)
+open import Data.Maybe   as Maybe   using (Maybe; just; nothing; maybe)
+open import Data.List    as List    using (List; _∷_; [])
+open import Data.Product as Product using (_×_; _,_)
+
+open import Agda.Builtin.Reflection
+open import Reflection.Argument
+open import Reflection.Term using (getName; _⋯⟅∷⟆_)
+open import Reflection.TypeChecking.Monad.Syntax
+
+import Categories.Morphism.Reasoning as MR
+
+module _ {o ℓ e} (𝒞 : Category o ℓ e) where
+
+  open Category 𝒞
+  open HomReasoning
+  open Equiv
+  open MR 𝒞
+
+  private
+    variable
+      A B C : Obj
+      f g   : A ⇒ B
+
+  --------------------------------------------------------------------------------
+  -- An 'Expr' reifies the parentheses/identity morphisms of some series of 
+  -- compositions of morphisms into a data structure. In fact, this is also
+  -- a category!
+  --------------------------------------------------------------------------------
+  data Expr : Obj → Obj → Set (o ⊔ ℓ) where
+    _∘′_ : ∀ {A B C} → Expr B C → Expr A B → Expr A C 
+    id′  : ∀ {A} → Expr A A
+    [_↑] : ∀ {A B} → A ⇒ B → Expr A B
+  
+  -- Embed a morphism in 'Expr' back into '𝒞' without normalizing.
+  [_↓] : Expr A B → A ⇒ B 
+  [ f ∘′ g ↓] = [ f ↓] ∘ [ g ↓]
+  [ id′ ↓]    = id
+  [ [ f ↑] ↓] = f
+  
+  -- Convert an 'Expr' back into a morphism, while normalizing
+  --
+  -- This actually embeds the morphism into the category of copresheaves
+  -- on 𝒞, which obeys the category laws up to beta-eta equality.
+  -- This lets us normalize away all the associations/identity morphisms.
+  embed : Expr B C → A ⇒ B → A ⇒ C
+  embed (f ∘′ g) h  = embed f (embed g h)
+  embed id′ h       = h
+  embed [ f ↑] h    = f ∘ h
+
+  
+  preserves-≈′ : ∀ (f : Expr B C) → (h : A ⇒ B) → embed f id ∘ h ≈ embed f h
+  preserves-≈′ id′ f      = identityˡ
+  preserves-≈′ [ x ↑] f   = ∘-resp-≈ˡ identityʳ
+  preserves-≈′ (f ∘′ g) h = begin
+    embed (f ∘′ g) id ∘ h         ≡⟨⟩
+    embed f (embed g id) ∘ h      ≈˘⟨ preserves-≈′ f (embed g id) ⟩∘⟨refl ⟩
+    (embed f id ∘ embed g id) ∘ h ≈⟨ pullʳ (preserves-≈′ g h) ⟩
+    embed f id ∘ embed g h        ≈⟨ preserves-≈′ f (embed g h) ⟩
+    embed (f ∘′ g) h              ∎
+  
+  preserves-≈ : ∀ (f : Expr A B) → embed f id ≈ [ f ↓]
+  preserves-≈ id′      = refl
+  preserves-≈ [ x ↑]   = identityʳ
+  preserves-≈ (f ∘′ g) = begin
+    embed (f ∘′ g) id       ≈˘⟨ preserves-≈′ f (embed g id) ⟩
+    embed f id ∘ embed g id ≈⟨ preserves-≈ f ⟩∘⟨ preserves-≈ g ⟩
+    [ f ↓] ∘ [ g ↓]         ≡⟨⟩
+    [ f ∘′ g ↓]             ∎
+
+--------------------------------------------------------------------------------
+-- Reflection Helpers
+--------------------------------------------------------------------------------
+
+_==_ = primQNameEquality
+{-# INLINE _==_ #-}
+
+getArgs : Term → Maybe (Term × Term)
+getArgs (def _ xs) = go xs
+  where
+  go : List (Arg Term) → Maybe (Term × Term)
+  go (vArg x ∷ vArg y ∷ []) = just (x , y)
+  go (x ∷ xs)               = go xs
+  go _                      = nothing
+getArgs _ = nothing
+
+--------------------------------------------------------------------------------
+-- Getting Category Names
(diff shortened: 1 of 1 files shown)


sstucki:
This is amazing!

Have you tried using it to simplify some of the bigger proofs in the library? Does it have a noticable impact? There are some big ones in the `Enriched` modules. It might be nice to also compare the compilation times for those. I don't know how efficient tactics are, compared to big equational reasoning chains using the DSL from the stdlib.

Now, what I really want though, is a solver for monoidal categories! 😀 If you think the proofs we have in the library now are bad, try proving some basic facts about monoidal categories, like coherence laws for the interchange. I've spent hours on this and haven't come half way. Not because it's a hard proof, but because it involves reassociating in 2D (using the associator and associativity). It's horrible. Especially when you know that a string-diagrammatic proof is basically immediate. I've been scratching my head on how one would even represent the embedding of the free monoidal category (essentially string diagrams) as a human-readable DSL in Agda, but it never occurred to me that one could use reflection/tactics to avoid that issue altogether! So this gives me hope that a monoidal solver might in fact be possible -- exciting! 🤩

TOTBWF:
I haven't tried yet, but I've started using it on some of the biproducts stuff and it does make life much easier, especially when you have 7+ morphisms that you need to reassociate. I think the should be reasonably fast, but perhaps some benchmarking is in order.

A monoidal solver would be really cool! I haven't thought about it too much, but for a closed SMC V we could probably do something similar, using the enriched yoneda lemma rather than the regular one.

JacquesCarette:
So I hold two simultaneous opinions on this:
1. it's very cool,
2. I'm deathly afraid of the speed hit, opacity of proofs and dependency-graph mess this might cause.

I figured I should warn you about this asap. A priori, including the infrastructure in the library is not a problem at all.

What I'm going to be quite a bit more reluctant about is to use it in the library itself. There, I'll need a fair bit more evidence than usual. [What would be best is if the solver produced equational proofs as output; then these could be used, instead of having the solver itself called each time the library needs to be typechecked. In other words, I would definitely be fine with the solver being used as a meta-program, rather than as a program.]

JacquesCarette:
Hmm, so do I have the ability to offer some code for this branch? I don't like the weird mucking around with `id` needed because `embed` has 2 arguments. I think this can be avoided, by having both a 1-argument and a 2-argument embedding.

Also, I need to dig into some of Ulf's reflection code, I think I've seen some nice stuff there that would make some of the code here a little less opaque (even though it is already quite straightforward and clean).

sstucki:
AFAIK @jespercockx wrote a tactic library that allows one to fill goals using a tactic so that the resulting term can be safed in the source file, but doesn't have to be written by hand. One can also re-generate proofs if there are dependencies by rerunning the tactic. I don't remember the details (maybe Jesper could help me out here) but IIRC the library is this one:
https://github.com/jespercockx/ataca/blob/master/src/Ataca/Demo.agda

TOTBWF:
I 100% agree with you on all fronts. I've done my best to keep the dependency footprint minimal. The only import from `agda-categories` is `Categories.Category`, and really that could be `Categories.Category.Core`.

The reason I prefer `agda-categories` over other libraries is the lack of magic, which is why this tactic is as unmagical as possible! It only handles reassocations + identities, which are (a) extremely obvious when you look at the steps in an eq-reasoning block and (b) extremely tedious when you have a non-trivial reassocation. Here's an example from some of the stuff I've been working on.

```
  π₁∘i₂-coconstant : f ∘ π₁ ∘ i₂ ≈ g ∘ π₁ ∘ i₂
  π₁∘i₂-coconstant {f = f} {g = g} = begin
    f ∘ π₁ ∘ i₂                                    ≈⟨ pushˡ (⟺ inject₁) ⟩
    [ f , g ∘ π₁ ∘ i₂ ] ∘ i₁ ∘ π₁ ∘ i₂             ≈⟨ introʳ π₂∘i₂≈id ⟩
    ([ f , g ∘ π₁ ∘ i₂ ] ∘ i₁ ∘ π₁ ∘ i₂) ∘ π₂ ∘ i₂ ≈⟨ solve 𝒞 ⟩
    [ f , g ∘ π₁ ∘ i₂ ] ∘ (i₁ ∘ π₁ ∘ i₂ ∘ π₂) ∘ i₂ ≈⟨ refl⟩∘⟨ permute ⟩∘⟨refl ⟩
    [ f , g ∘ π₁ ∘ i₂ ] ∘ (i₂ ∘ π₂ ∘ i₁ ∘ π₁) ∘ i₂ ≈⟨ pullˡ (pullˡ inject₂) ⟩
    ((g ∘ π₁ ∘ i₂) ∘ π₂ ∘ i₁ ∘ π₁) ∘ i₂            ≈⟨ solve 𝒞 ⟩
    g ∘ π₁ ∘ (i₂ ∘ π₂ ∘ i₁ ∘ π₁) ∘ i₂              ≈˘⟨ refl⟩∘⟨ refl⟩∘⟨ permute ⟩∘⟨refl ⟩
    g ∘ π₁ ∘ (i₁ ∘ π₁ ∘ i₂ ∘ π₂) ∘ i₂              ≈⟨ refl⟩∘⟨ pullˡ (cancelˡ π₁∘i₁≈id) ⟩
    g ∘ (π₁ ∘ i₂ ∘ π₂) ∘ i₂                        ≈⟨ refl⟩∘⟨ pullʳ (cancelʳ π₂∘i₂≈id) ⟩
    g ∘ π₁ ∘ i₂ ∎
```

Without `solve`, each of these associations become monsters

```
  π₁∘i₂-coconstant : f ∘ π₁ ∘ i₂ ≈ g ∘ π₁ ∘ i₂
  π₁∘i₂-coconstant {f = f} {g = g} = begin
    f ∘ π₁ ∘ i₂                                    ≈⟨ pushˡ (⟺ inject₁) ⟩
    [ f , g ∘ π₁ ∘ i₂ ] ∘ i₁ ∘ π₁ ∘ i₂             ≈⟨ introʳ π₂∘i₂≈id ⟩
    ([ f , g ∘ π₁ ∘ i₂ ] ∘ i₁ ∘ π₁ ∘ i₂) ∘ π₂ ∘ i₂ ≈⟨ pullʳ (pullˡ (pullʳ assoc)) ⟩
    [ f , g ∘ π₁ ∘ i₂ ] ∘ (i₁ ∘ π₁ ∘ i₂ ∘ π₂) ∘ i₂ ≈⟨ refl⟩∘⟨ permute ⟩∘⟨refl ⟩
    [ f , g ∘ π₁ ∘ i₂ ] ∘ (i₂ ∘ π₂ ∘ i₁ ∘ π₁) ∘ i₂ ≈⟨ pullˡ (pullˡ inject₂) ⟩
    ((g ∘ π₁ ∘ i₂) ∘ π₂ ∘ i₁ ∘ π₁) ∘ i₂            ≈⟨ assoc ○ extendʳ (assoc ○ pushʳ assoc) ○ assoc ⟩
    g ∘ π₁ ∘ (i₂ ∘ π₂ ∘ i₁ ∘ π₁) ∘ i₂              ≈˘⟨ refl⟩∘⟨ refl⟩∘⟨ permute ⟩∘⟨refl ⟩
    g ∘ π₁ ∘ (i₁ ∘ π₁ ∘ i₂ ∘ π₂) ∘ i₂              ≈⟨ refl⟩∘⟨ pullˡ (cancelˡ π₁∘i₁≈id) ⟩
    g ∘ (π₁ ∘ i₂ ∘ π₂) ∘ i₂                        ≈⟨ refl⟩∘⟨ pullʳ (cancelʳ π₂∘i₂≈id) ⟩
    g ∘ π₁ ∘ i₂ ∎
```

I suspect that reflection may actually be faster here, as we don't need to typecheck nearly as many things (that 2nd solve call transitively uses 6 `assoc` calls!)

TOTBWF:
The 2 arguments to `embed` are what make the whole thing work, sadly. By turning our reflected morphism into an action on morphisms, we get beta-eta equality for free. I could add a helper function
```
  [_⇓] : Expr A B → A ⇒ B
  [ f ⇓] = embed f id
```
if that helps clarify things.

src/Categories/Tactic/Category.agda
```diff
@@ -0,0 +1,184 @@
...
+    embed f id ∘ embed g id ≈⟨ preserves-≈ f ⟩∘⟨ preserves-≈ g ⟩
+    [ f ↓] ∘ [ g ↓]         ≡⟨⟩
+    [ f ∘′ g ↓]             ∎
+
+--------------------------------------------------------------------------------
+-- Reflection Helpers
+--------------------------------------------------------------------------------
+
+_==_ = primQNameEquality
+{-# INLINE _==_ #-}
```
HuStmpHrrr, line 89: I am wondering whether stdlib has provided basic definitions like this?

MatthewDaggitt: Apparently [not](https://github.com/agda/agda-stdlib/search?q=primQNameEquality). Might be good to add it to `Reflection.Name`?

HuStmpHrrr: that would be reasonable.

TOTBWF: Yeah, that would be a good call. I stole this part shamelessly from the `MonoidSolver` in the stdlib, so if that gets added, we ought to fix that as well.

TOTBWF:
I did some preliminary benchmarking of this with agda-bench, using the following test file:
```

{-# OPTIONS --without-K --safe #-}

module Experiments.Tactics {o ℓ e} {𝒞 : Category o ℓ e} where

open import Categories.Tactic.Category using (solve; solve-macro)

open import Categories.Morphism.Reasoning 𝒞

open Category 𝒞
open HomReasoning


private
  variable
    A B C : Obj
    f g h i j k : A ⇒ B

bench-tactic : ((id ∘ f) ∘ id) ∘ (id ∘ (g ∘ h)) ≈ f ∘ (g ∘ id ∘ h)
bench-tactic = solve 𝒞

bench-reasoning : ((id ∘ f) ∘ id) ∘ (id ∘ (g ∘ h)) ≈ f ∘ (g ∘ id ∘ h)
bench-reasoning {f = f} {g = g} {h = h} = begin
  ((id ∘ f) ∘ id) ∘ id ∘ g ∘ h ≈⟨ identityʳ ⟩∘⟨ identityˡ ⟩
  (id ∘ f) ∘ g ∘ h             ≈⟨ identityˡ ⟩∘⟨refl ⟩
  f ∘ g ∘ h                    ≈⟨ refl⟩∘⟨ pushˡ (⟺ identityʳ) ⟩
  f ∘ g ∘ id ∘ h               ∎

bench-tactic-assoc : f ∘ g ∘ h ∘ i ∘ j ∘ k ≈ ((((f ∘ g) ∘ h) ∘ i) ∘ j) ∘ k
bench-tactic-assoc = solve 𝒞

bench-reasoning-assoc : ∀ {A} {f g h i j k : A ⇒ A} → f ∘ g ∘ h ∘ i ∘ j ∘ k ≈ ((((f ∘ g) ∘ h) ∘ i) ∘ j) ∘ k
bench-reasoning-assoc {f = f} {g = g} {h = h} {i = i} {j = j} {k = k} = begin
  f ∘ g ∘ h ∘ i ∘ j ∘ k       ≈˘⟨ assoc  ⟩
  (f ∘ g) ∘ h ∘ i ∘ j ∘ k     ≈˘⟨ assoc ⟩
  ((f ∘ g) ∘ h) ∘ i ∘ j ∘ k   ≈˘⟨ assoc ⟩
  (((f ∘ g) ∘ h) ∘ i) ∘ j ∘ k ≈˘⟨ assoc ⟩
  ((((f ∘ g) ∘ h) ∘ i) ∘ j) ∘ k ∎
```

The results were as follows:
```
benchmarking bench-tactic
time                 2.280 μs   (2.187 μs .. 2.374 μs)
                     0.988 R²   (0.981 R² .. 0.994 R²)
mean                 2.218 μs   (2.160 μs .. 2.297 μs)
std dev              232.6 ns   (180.7 ns .. 311.6 ns)
variance introduced by outliers: 89% (severely inflated)

benchmarking bench-reasoning
time                 2.055 μs   (2.030 μs .. 2.085 μs)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 2.058 μs   (2.037 μs .. 2.091 μs)
std dev              86.75 ns   (62.61 ns .. 117.9 ns)
variance introduced by outliers: 56% (severely inflated)

benchmarking bench-tactic-assoc
time                 2.058 μs   (2.025 μs .. 2.089 μs)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 2.069 μs   (2.046 μs .. 2.094 μs)
std dev              74.67 ns   (64.45 ns .. 103.1 ns)
variance introduced by outliers: 48% (moderately inflated)

benchmarking bench-reasoning-assoc
time                 2.228 μs   (2.145 μs .. 2.345 μs)
                     0.979 R²   (0.966 R² .. 0.988 R²)
mean                 2.528 μs   (2.368 μs .. 2.732 μs)
std dev              661.8 ns   (514.0 ns .. 826.8 ns)
variance introduced by outliers: 98% (severely inflated)
```

JacquesCarette:
Just to be sure: this is ready for merge?  [I'm fine with it now]

TOTBWF:
Yes, this ought to be good to go 🙂

PR #259, opened 2021-03-23
Kernel Pair

JacquesCarette:
simple synonym for Kernel Pair. Useful in defining pieces that will eventually (!) lead to Pretopos.

### DIFF
--- a/src/Categories/Diagram/KernelPair.agda
+++ b/src/Categories/Diagram/KernelPair.agda
@@ -1,0 +1,21 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Categories.Category.Core
+
+-- Kernel Pair - a Pullback of a morphism along itself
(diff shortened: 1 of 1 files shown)


src/Categories/Diagram/KernelPair.agda
```diff
@@ -0,0 +1,21 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Categories.Category.Core
+
+-- Kernel Pair - a Pullback of a morphism along itself
+-- https://ncatlab.org/nlab/show/reflexive+coequalizer
```
HuStmpHrrr, line 6: is this link accurate? https://ncatlab.org/nlab/show/kernel+pair

JacquesCarette: Oops! I must have pasted from the wrong window. Nice catch.

PR #260, opened 2021-03-23
Regular

JacquesCarette:
Regular Category, Epi and Mono morphisms.  Needed for Pretopos.

### DIFF
--- a/src/Categories/Morphism/Regular.agda
+++ b/src/Categories/Morphism/Regular.agda
@@ -1,0 +1,34 @@
+{-# OPTIONS --without-K --safe #-}
+
+{-
+  Properties regarding Morphisms of a category:
+  - Regular Monomorphism
+  - Regular Epimorphism
+
+  https://ncatlab.org/nlab/show/regular+epimorphism
+
+  These are defined here rather than in Morphism, as this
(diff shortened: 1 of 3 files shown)


TOTBWF:
Should we maybe pull the `Regular` morphisms into records? Seems like it might be a bit easier to work with long term.

JacquesCarette:
Good point. [And exactly why I'm now switching to doing PRs myself, to get this kind of feedback.]

src/Categories/Morphism/Regular.agda
```diff
@@ -0,0 +1,34 @@
...
+open import Categories.Diagram.Equalizer 𝒞
+open import Categories.Diagram.Coequalizer 𝒞
+
+open Category 𝒞
+
+private
+  variable
+    A B : Obj
+
+RegularMono : (f : A ⇒ B) → Set (o ⊔ ℓ ⊔ e)
```
HuStmpHrrr, line 30: would it be better if you define them as records?

JacquesCarette: Yes, that was @TOTBWF comment too. As soon as I get a minute...

PR #262, opened 2021-04-04
Quiver

JacquesCarette:
This is a rather large refactor of what used to be Graph into Quiver (as that's really the more appropriate name for this context). It's not necessarily fully done, but I want feedback before I continue.


JacquesCarette:
Note that one of the things I'm considering is submitting Quiver (and some other pieces) to stdlib, and maybe move a certain amount of the categorical stuff to categories-examples, as this stuff really doesn't need to be in agda-categories per se.

src/Categories/Category/Construction/Quivers.agda
```diff
@@ -0,0 +1,191 @@
...
+        F₁ G f ∘ hid (eq₀ F≈G x)   ≈˘⟨ hid-subst-dom (eq₀ F≈G x) (F₁ G f) ⟩
+        eq₀ F≈G x ◂ F₁ G f         ∎
+    }
+  }
+  where
+  open NaturalTransformation
+  open NaturalIsomorphism
+  open _≡F_
+
+-- define these ahead of time
```
sstucki, line 155: These feel out of place here. Should they go into `Data.Quiver.Morphism`?

src/Categories/Category/Construction/Quivers.agda
```diff
@@ -0,0 +1,191 @@
...
+-- 3. If you want a Functor Categories Graph, then the object/hom levels are straightforward,
+--   but if you use "Categories" with NaturalIsomorphism, then the 2-point Groupoid is contractible
+--   (thus equivalent to the 1-point Category), and yet those two Graphs shouldn't be considered to
+--   be the same!  So _≡F_ must be used so that we have no wiggle room when mapping Graphs around.
+--   Interestingly, this only shows up when trying to prove that the Underlying Functor respects
+--   _≈_, which is not traditionally a requirement of a Functor! This requirement is usually left
+--   implicit.  If we were to allow Functor to not respect _≈_, things would be fine. But of course
+--   we couldn't do any reasonable equational reasoning anymore.  Of course, if _≈_ were the global
+--   _≡_, this isn't an issue, but that is definitely not wanted here.
+-- 4. A lot of weird little lemmas about _≡_ and subst/transport are needed, because we're in a mixed
```
sstucki, line 20: Not sure this comment applies any longer after this reorganization.

src/Categories/Functor/Construction/FreeCategory.agda
```diff
@@ -0,0 +1,65 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Functor.Construction.FreeCategory where
```
sstucki, line 3: Here it's called `FreeCategory` but the construction of the category itself is called `PathCategory`. These should be made consistent.

src/Data/Quiver/Morphism.agda
```diff
@@ -0,0 +1,56 @@
...
+_∘_ : {G₁ G₂ G₃ : Quiver o ℓ e} (m₁ : Morphism G₂ G₃) (m₂ : Morphism G₁ G₂) → Morphism G₁ G₃
+m₁ ∘ m₂ = record
+  { F₀ = F₀ m₁ ⊚ F₀ m₂
+  ; F₁ = F₁ m₁ ⊚ F₁ m₂
+  ; F-resp-≈ = F-resp-≈ m₁ ⊚ F-resp-≈ m₂
+  }
+  where open Morphism
+
+-- Define Morphism equivalence here as well.
+-- (Proof that it's an equivalence relation straightforward, but not needed?)
```
sstucki, line 41: It's needed in `Categories.Category.Construction.Quivers` (see my comment there).

src/Data/Quiver/Morphism.agda
```diff
@@ -0,0 +1,56 @@
...
+  open Quiver G′ using (_≈_)
+  private
+    module M  = Morphism M
+    module M′ = Morphism M′
+  open Shorthands (Quiver._⇒_ G′)
+
+  -- Pick a presentation of equivalence for graph morphisms that works
+  -- well with functor equality.
+
+  field
```
sstucki, line 54: I think I was complicit in this definition. I remember now why we picked it: it's close to (strict) functor equality, which helps with some of the proofs. Ideally, we'd like something weaker, akin to natural isomorphism, but since there's no notion of composition and identity for edges, that doesn't work (there are not commuting squares in quivers). So we have to be strict here, unless we also introduce a notion of equality on nodes. Alas, that would make the definition of quivers rather different than that of categories.  All this should be documented here!

Digression: if "setoid-enriched quivers" are not the natural generalization of categories in the setoid-enriched setting, then what is?

sstucki (changes requested):
I wrote this "review" in one go, so it documents my confusion at various points, even if some of it subsequently got resolved. The result is a rather long review, sorry about that. I hope it's helpful in showing where the documentation or module structure could be improved. If not, feel free to ignore the comments that turned out to be addressed somewhere else.

Other than that, most of my comments are about documenting or renaming things.
I didn't spot any technical issues.

There are some more philosophical points too. It still feels like quivers don't quite fit into this library, but I'm not sure how this could be solved or if there's a replacement that would feel more natural. Food for thought.

src/Categories/Category/Construction/Quivers.agda
```diff
@@ -0,0 +1,191 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Category.Construction.Quivers where
+
+-- The "Underlying Graph" <-> "Free Category on a Graph" Adjunction.
```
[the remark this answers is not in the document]
sstucki: UPDATE: Oh, actually, these definitions were moved somewhere else too, so maybe move those comments too then.

src/Categories/Category/Construction/FreeQuiver.agda
```diff
@@ -0,0 +1,42 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Level
+open import Data.Quiver using (Quiver)
+
+-- The Category of (free) paths over a Quiver
+module Categories.Category.Construction.FreeQuiver {o ℓ e}
```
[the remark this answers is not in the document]
HuStmpHrrr: this is really the notion of free category. I suggest we call this `Free`.

src/Categories/Functor/Construction/FreeCategory.agda
HuStmpHrrr: this name is somewhat misleading. it's a functor, not a category, so `FreeCategory` is not quite accurate. not sure what's a better name though. what about `FreeCatF`? at least it makes it clear it's a functor.

src/Categories/Adjoint/Instance/QuiverCategory.agda
```diff
@@ -0,0 +1,91 @@
...
+import Categories.Category.Construction.FreeQuiver as FQ
+open import Categories.Category.Construction.Quivers
+open import Categories.Functor using (Functor)
+open import Categories.Functor.Construction.FreeCategory
+open import Categories.NaturalTransformation hiding (id)
+import Categories.Morphism.Reasoning as MR
+
+module _ (o ℓ e : Level) where
+
+  GM : (X : Quiver o (o ⊔ ℓ) (o ⊔ ℓ ⊔ e)) → Morphism X (Underlying₀ (FQ.PathCategory X))
```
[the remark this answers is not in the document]
JacquesCarette: Right, `unit` is correct - done.

src/Categories/Category/Construction/FreeQuiver.agda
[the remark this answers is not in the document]
JacquesCarette: After thinking about this for a while, I decided that `Free` is too overloaded and that I wanted something more indicative, so settled on `PathCategory`.

src/Categories/Category/Construction/FreeQuiver.agda
```diff
@@ -0,0 +1,42 @@
...
+open import Relation.Binary.Construct.Closure.ReflexiveTransitive.Properties hiding (trans)
+open import Data.Quiver.Paths
+
+open import Categories.Category.Core using (Category)
+
+open Quiver G
+private module P = Paths G
+open P
+
+PathCategory : Category o (o ⊔ ℓ) (o ⊔ ℓ ⊔ e)
```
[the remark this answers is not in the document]
JacquesCarette: Agreed and done.

src/Categories/Category/Construction/Quivers.agda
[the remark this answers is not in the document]
JacquesCarette: I've moved all the comments to where the definitions are.

src/Categories/Category/Construction/Quivers.agda
```diff
@@ -0,0 +1,191 @@
...
+--   [It's actually a Multidigraph, or Quiver.  Use the latter.]
+-- Lots of surprises here, of various level of surprisingness
+-- 1. The Free Category rises universe levels of arrows (and equivalences); the Underlying Graph does not
+--   (due to the Transitive Closure "Star" construction)
+-- 2. As a consequence of (1), the eventual Adjunction will be for all Levels being the same
+-- 3. If you want a Functor Categories Graph, then the object/hom levels are straightforward,
```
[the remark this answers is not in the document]
JacquesCarette: Yes, this was badly named - now `Functor Cats Quivers`.  I agree that the commentary about equivalence was off, I've rewritten it. I'll indeed move the comments to the appropriate places.

src/Categories/Category/Construction/Quivers.agda
```diff
@@ -0,0 +1,191 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Category.Construction.Quivers where
```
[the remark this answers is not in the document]
JacquesCarette: You were right - moved.

src/Categories/Category/Construction/Quivers.agda
```diff
@@ -0,0 +1,191 @@
...
+  ; _⇒_       = Morphism
+  ; _≈_       = _≃_
+  ; id        = QM.id
+  ; _∘_       = QM._∘_
+  ; assoc     = λ {_ _ _ G} → record { F₀≡ = refl ; F₁≡ = Equiv.refl G }
+  ; sym-assoc = λ {_ _ _ G} → record { F₀≡ = refl ; F₁≡ = Equiv.refl G }
+  ; identityˡ = λ {_ G}     → record { F₀≡ = refl ; F₁≡ = Equiv.refl G }
+  ; identityʳ = λ {_ G}     → record { F₀≡ = refl ; F₁≡ = Equiv.refl G }
+  ; identity² = λ {G}       → record { F₀≡ = refl ; F₁≡ = Equiv.refl G }
+  ; equiv     = λ {_ G} → record
```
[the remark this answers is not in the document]
JacquesCarette: Good point - done.

src/Categories/Category/Construction/Quivers.agda
JacquesCarette: In the construction of the category, no. But in the constructor of the functor, it does, so I've adjusted the comments there.

src/Data/Quiver/Morphism.agda
JacquesCarette: Comments added.

I've asked some questions around that Digression on the category theory Zulip, and didn't get any decent leads. Fundamentally, I don't think there is any!  Graphs, even as Quivers with setoids of edges, are still strict in some sense. The fundamental thing is that the naive "underlying graph" doesn't respect category equivalence. For example, the 2-point Groupoid is equivalent to the 1-point Category, but one tends to see those graphs as different.

src/Categories/Category/Construction/Quivers.agda
JacquesCarette: Done.

src/Categories/Functor/Construction/FreeCategory.agda
JacquesCarette: After a lot of thinking, I went with `PathsOf` as the name of the functor.

JacquesCarette:
Ok, everything taken care of!  This is much nicer now, thanks.

src/Categories/Category/Slice/Properties.agda
```diff
@@ -2,6 +2,9 @@
 
 open import Categories.Category
 
+-- Various properties of Slice Categories
```
sstucki, line 5: Is this changeset supposed to be part of this PR?

src/Categories/Morphism/Reasoning.agda
```diff
@@ -6,5 +6,18 @@ open import Categories.Category
...
 
+-- some items are defined in sub-modules
 open import Categories.Morphism.Reasoning.Core C public
 open import Categories.Morphism.Reasoning.Iso C public
+
+open Category C
+open Definitions C
+open HomReasoning
+
+-- create a commutative square from an equivalence
```
sstucki, line 17: Again, I'm wondering why this is in the PR.

sstucki (approved):
LGTM.
There are two changesets where I'm not sure whether they are supposed to be in the PR, though.

src/Categories/Category/Slice/Properties.agda
JacquesCarette: It's a little sloppy to have this here, but it's just comments, so it seemed ok to throw them in.

src/Categories/Morphism/Reasoning.agda
JacquesCarette: I think because `toSquare` was buried in some Quiver-related file, and it made most sense to move it here at the same time.

Issue #256, opened 2021-03-03
It would be nice for (1 -> C ᵒᵖ) to be the same as (1 -> C) ᵒᵖ


JasonGross:
> We've actually gone out of our way in agda-categories to make as many of these things hold judgementally.  For `1 -> C` vs `C`, that is hopeless, as they are fundamentally different even if equivalent (though there might be odd things when `C` is not inhabited).  Certainly  `(1 -> C ᵒᵖ)` should be the same as `(1 -> C) ᵒᵖ`, and it something that can be fixed if it's not.  For Kan extension, it's possible we've erred, and we'd be glad to see a concrete bug report.

_Originally posted by @JacquesCarette in https://github.com/agda/agda/issues/5255#issuecomment-790045655_

```agda
{-# OPTIONS --without-K --safe #-}
module Categories.test where
open import Level
open import Relation.Binary.PropositionalEquality as ≡ using (_≡_; refl)
open import Categories.Functor
open import Categories.Category.Core
open import Categories.Category.Instance.One
open import Categories.Category.Construction.Functors

private
  variable
    o ℓ e o′ ℓ′ e′ : Level

test : ∀ {C : Category o ℓ e} → Category.op (Functors (One {o′} {ℓ′} {e′}) C) ≡ Functors (One {o′} {ℓ′} {e′}) (Category.op C)
test = {refl }0
{- x != x₁ of type Category.Obj C
when checking that the expression refl has type
Category.op (Functors One C) ≡ Functors One (Category.op C) -}
```

I believe if you make `Functor` a nested sigma type rather than a record, this equality will go through judgmentally.  I'm not 100% sure, though.

JacquesCarette:
Oh, you're asking for something much stronger than I expected.  But it's still an interesting question. I'm investigating now.

JasonGross:
Great!  I'm curious to know the results of your investigation.  Btw, the ~equivalent PR for the Coq HoTT/HoTT library is https://github.com/HoTT/HoTT/pull/849 (abandoned because it resulted in a moderate slowdown for a couple files, and also did not buy us anything without judgmental eta for the unit type)

JacquesCarette:
Are you sure you're asking the right question?  The objects aren't the same:
```agda
  module _ {C : Category o ℓ e} {o′ ℓ′ e′ : Level} where
    open Category
    open Functor

    X Y : Category _ _ _
    X = Category.op (Functors (One {o′} {ℓ′} {e′}) C)
[...]

PR #264, opened 2021-04-08
Kan Complexes and Weak Kan Complexes

TOTBWF:
## Patch Description
This PR implements Kan Complexes and Weak Kan Complexes, as well as adding some machinery surrounding Simplicial Sets that is required to do so. Of note are definitions for face and degeneracy maps, as well as boundaries and horns.
Particularly, this implements _Algebraic_ Kan Complexes, as those have less fiddly issues with choice.

## Notes
The combinatorics here were not as bad as I initially thought they would be! Everything works out rather well. However, there are a few things that I'm not 100% sure of.
- The inequality in the definition of `Horn` seems like it could become a bit awkward down the line?
- The proof obligations in `WeakKanComplex` seem pretty painful. Perhaps there is a more clever way of doing this?
- The unfolded equality in `∂Δ[_]` and `Λ[_,_]` does better than just lifting the equality from `Δ`, but it still falls on it's face from time to time (See `KanComplex⇒WeakKanComplex`)

## References
- https://ncatlab.org/nlab/show/Kan+complex
- https://ncatlab.org/nlab/show/weak+Kan+complex
- https://ncatlab.org/nlab/show/algebraic+Kan+complex

### DIFF
--- a/src/Categories/Category/Instance/Simplex.agda
+++ b/src/Categories/Category/Instance/Simplex.agda
@@ -7,10 +7,10 @@ module Categories.Category.Instance.Simplex where
 open import Level
 open import Data.Product
 open import Data.Fin.Base using (Fin; _≤_)
-open import Data.Nat.Base using (ℕ)
+open import Data.Nat.Base using (ℕ; z≤n; s≤s)
 open import Function renaming (id to idF; _∘_ to _∙_)
 
-open import Relation.Binary
+open import Relation.Binary using (_=[_]⇒_)
 open import Relation.Binary.PropositionalEquality
 
 Δ : Category 0ℓ 0ℓ 0ℓ
@@ -32,3 +32,38 @@ open import Relation.Binary.PropositionalEquality
     }
   ; ∘-resp-≈  = λ {_ _ _ f g h i} eq₁ eq₂ x → trans (cong (λ t → proj₁ f t) (eq₂ x)) (eq₁ (proj₁ i x))
   }
+
+open Category Δ
+
+--------------------------------------------------------------------------------
+-- Face + Degeneracy Maps
+
+face-map : ∀ {n} → Fin (ℕ.suc n) → Fin n → Fin (ℕ.suc n)
+face-map Fin.zero k = Fin.suc k
+face-map (Fin.suc i) Fin.zero = Fin.zero
+face-map (Fin.suc i) (Fin.suc k) = Fin.suc (face-map i k)
+
+face-mono : ∀ {n} → (i : Fin (ℕ.suc n)) → _≤_ =[ face-map i ]⇒ _≤_
+face-mono Fin.zero {x} {y} le = s≤s le
+face-mono (Fin.suc i) {Fin.zero} {y} le = z≤n
+face-mono (Fin.suc i) {Fin.suc x} {Fin.suc y} (s≤s le) = s≤s (face-mono i le)
+
+face : ∀ {n} → Fin (ℕ.suc n) → n ⇒ ℕ.suc n
+face i = face-map i , face-mono i
+
+degeneracy-map : ∀ {n} → Fin n → Fin (ℕ.suc n) → Fin n
+degeneracy-map Fin.zero Fin.zero = Fin.zero
+degeneracy-map Fin.zero (Fin.suc Fin.zero) = Fin.zero
+degeneracy-map Fin.zero (Fin.suc (Fin.suc k)) = Fin.suc k
+degeneracy-map (Fin.suc i) Fin.zero = Fin.zero
+degeneracy-map (Fin.suc i) (Fin.suc k) = Fin.suc (degeneracy-map i k)
+
+degeneracy-mono : ∀ {n} → (i : Fin n) → _≤_ =[ degeneracy-map i ]⇒ _≤_
+degeneracy-mono Fin.zero {Fin.zero} {y} le = z≤n
+degeneracy-mono Fin.zero {Fin.suc (Fin.zero)} {Fin.suc y} le = z≤n
+degeneracy-mono Fin.zero {Fin.suc (Fin.suc x)} {Fin.suc (Fin.suc y)} (s≤s le) = le
+degeneracy-mono (Fin.suc i) {Fin.zero} {y} le = z≤n
+degeneracy-mono (Fin.suc i) {Fin.suc x} {Fin.suc y} (s≤s le) = s≤s (degeneracy-mono i le)
+
+degeneracy : ∀ {n} → Fin n → ℕ.suc n ⇒ n
+degeneracy i = degeneracy-map i , degeneracy-mono i
--- a/src/Categories/Category/Instance/Simplex.agda
+++ b/src/Categories/Category/Instance/Simplex.agda
@@ -39,31 +39,29 @@ open Category Δ
 -- Face + Degeneracy Maps
 
 face-map : ∀ {n} → Fin (ℕ.suc n) → Fin n → Fin (ℕ.suc n)
-face-map Fin.zero k = Fin.suc k
-face-map (Fin.suc i) Fin.zero = Fin.zero
+face-map Fin.zero    k           = Fin.suc k
+face-map (Fin.suc i) Fin.zero    = Fin.zero
 face-map (Fin.suc i) (Fin.suc k) = Fin.suc (face-map i k)
 
 face-mono : ∀ {n} → (i : Fin (ℕ.suc n)) → _≤_ =[ face-map i ]⇒ _≤_
-face-mono Fin.zero {x} {y} le = s≤s le
-face-mono (Fin.suc i) {Fin.zero} {y} le = z≤n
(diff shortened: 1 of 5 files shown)


sstucki:
I like the notes in the PR description. Can we add them to the README.md or somewhere else in the documentation? Maybe a README/Kan complexes.md?

TOTBWF:
That's a great idea! I've made a `README` directory and added some notes to it. Perhaps there is a larger organizational discussion to be had about how we should keep track of various bits and bobs that we come across while implementing :)

TOTBWF:
Oh man those Kan Complex equalities do _not_ make Agda happy! For some reason this passed locally without any yellow, but upon further inspection this seems to have unsolved metas anyways.

To make matters even _more_ confusing, Agda seems to be giving up randomly. For instance, the following proof is broken, and results in a huge pile of unsolved metas:
```agda
    ; Rid = λ { {X , X-Kan} i → begin
        id ∘ filler X-Kan i   ≈⟨ identityˡ {f = filler X-Kan i} ⟩
        -- What???
        filler X-Kan i        ≈˘⟨ filler-cong X-Kan (⟺ (identityˡ {f = i})) ⟩
        filler X-Kan (id ∘ i) ∎
      }
```
However, if we move that symmetry out it works fine?
```agda
    ; Rid = λ { {X , X-Kan} i → begin
        id ∘ filler X-Kan i   ≈⟨ identityˡ {f = filler X-Kan i} ⟩
        -- What???
        filler X-Kan i        ≈˘⟨ filler-cong X-Kan (identityˡ {f = i}) ⟩
        filler X-Kan (id ∘ i) ∎
      }
```

TOTBWF:
Ok, I think I've got it compiling, but whatever I've created has totally trashed implicit resolution for reasons I don't quite understand. For some reason it looks like agda keeps eta-expanding records, and then filling a bunch of fields in with totally trash metavariables that never get solved. Perhaps we should think harder about what is going on with the equality here, because it's almost impossible to work with.

sstucki:
> For some reason it looks like agda keeps eta-expanding records, and then filling a bunch of fields in with totally trash metavariables that never get solved.

Indeed, this is a curse or a blessing depending on context. It's a pain when you want to reason about e.g. "derived" notions because Agda first computes everything to weak-head normal forms before unification. Along the way, it often looses context that it then replaces by metas. But it's a blessing if you want to identify structures that are "extensionally" equal, i.e. all the record fields agree, but not the original expressions. There's a related discussion in #256.

sstucki:
In particular, [this comment](https://github.com/agda/agda-categories/issues/256#issuecomment-790885946) has an example.

TOTBWF:
I've started a Zulip thread for investigating how we can fix the metavariable issues in `filler-cong`

JacquesCarette:
So should this PR wait until changes to stdlib, or would you like this in as soon as it's ready, and it can be improved later?

src/Categories/Category/Instance/Simplex.agda
```diff
@@ -32,3 +32,38 @@ open import Relation.Binary.PropositionalEquality
...
+face-mono Fin.zero {x} {y} le = s≤s le
+face-mono (Fin.suc i) {Fin.zero} {y} le = z≤n
+face-mono (Fin.suc i) {Fin.suc x} {Fin.suc y} (s≤s le) = s≤s (face-mono i le)
+
+face : ∀ {n} → Fin (ℕ.suc n) → n ⇒ ℕ.suc n
+face i = face-map i , face-mono i
+
+degeneracy-map : ∀ {n} → Fin n → Fin (ℕ.suc n) → Fin n
+degeneracy-map Fin.zero Fin.zero = Fin.zero
+degeneracy-map Fin.zero (Fin.suc Fin.zero) = Fin.zero
```
JacquesCarette, line 56: Do you need to split these two (line 56 and 57) ? Looks like you don't.

src/Categories/Category/Instance/SimplicialSet/Properties.agda
```diff
@@ -0,0 +1,212 @@
...
+
+open import Data.Empty.Polymorphic
+open import Data.Nat using (ℕ)
+open import Data.Fin using (Fin; _<_; fromℕ)
+open import Data.Fin.Patterns
+open import Data.Product using (proj₁)
+
+import Relation.Binary.PropositionalEquality as Eq
+
+open import Categories.Category
```
JacquesCarette, line 16: If you could try to be more precise on the imports, especially when you need few things?

src/Categories/Category/Instance/SimplicialSet/Properties.agda
```diff
@@ -0,0 +1,212 @@
...
+Δ[_] : ℕ → ΔSet
+Δ[_] n = LiftSetoids o ℓ ∘F Y.F₀ n
+
+--------------------------------------------------------------------------------
+-- Boundaries of the Standard Simplicies
+--
+-- The basic idea here is that we will build up boundary of 'Δ[ n ]' by considering
+-- all of the morphisms 'Δ[ m , n ]' that factor through some face map 'face i : Δ[ n - 1 , n ]'
+
+-- The indexing here is a bit tricky, but this is the price we pay to avoid 'pred'
```
JacquesCarette, line 45: Excellent comments.

src/Categories/Category/Instance/SimplicialSet/Properties.agda
```diff
@@ -0,0 +1,212 @@
...
+-- A 'Boundary m n' represents some set of maps into 'Δ[ ℕ.suc n ]' that factor through
+-- a face map. To make this indexing more obvious, we use the suggestively named variable 'n-1'.
+record Boundary (m n-1 : ℕ) : Set where
+  field
```
JacquesCarette, line 49: maybe use
```agda
private
  n : ℕ
  n = ℕ.suc n-1
```
so as to be less confusing?

src/Categories/Category/Instance/SimplicialSet/Properties.agda
```diff
@@ -0,0 +1,212 @@
...
+  ; factor-face = λ x → factor-face b (proj₁ f x)
+  }
+  where
+    open Category Δ
+    open Boundary
+
+-- The boundary of an n-simplex
+∂Δ[_] : ℕ → ΔSet 
+∂Δ[_] ℕ.zero = const record
+  { Carrier = ⊥
```
JacquesCarette, line 70: I'm pretty sure `record { Carrier = ⊥ }` would work - Agda's pretty good at auto-filling the rest in this case.

src/Categories/Category/Instance/SimplicialSet/Properties.agda
```diff
@@ -0,0 +1,212 @@
...
+    { η = λ X → record
+      { _⟨$⟩_ = λ (lift h) → lift (hom h)
+      ; cong = λ eq → lift (λ x → lower (eq x))
+      }
+    ; commute = λ f eq → lift (λ x → lower (eq (proj₁ f x)))
+    }
+    where
+      open Horn
+  
+  record IsKanComplex (X : ΔSet) : Set (o ⊔ ℓ) where
```
JacquesCarette, line 192: This is not the right file in which to define this. An actual `KanComplex` file would make sense. The top-level is getting a little crowded though. `Categories.Constructions.KanComplex` ?

JacquesCarette (changes requested):

HuStmpHrrr:
better keep the description in the source code, so it can get updated as the code itself evolves.

HuStmpHrrr:
overall these PRs you submit recently are very great contributions!

TOTBWF:
Thanks! I really appreciate it 🙂 I've been pretty busy these past few weeks, but I hope to get back to update this PR soon!

src/Categories/Category/Instance/SimplicialSet/Properties.agda
TOTBWF: Good call! Moved

src/Categories/Category/Instance/SimplicialSet/Properties.agda
TOTBWF: For some reason this doesn't seem to work? I suspect there are some normalization shenanigans here

src/Categories/Category/Instance/SimplicialSet/Properties.agda
TOTBWF: Good idea! Done

src/Categories/Category/Instance/Simplex.agda
TOTBWF: Looks like I can, nice catch :)

TOTBWF:
Ok, I bit the bullet and updated `IsWeakKanComplex` so that we avoid the (really annoying) proof obligations for internal horns. The indexing does get a bit crazy, but this is the price we pay!

src/Categories/Category/Instance/SimplicialSet/Properties.agda
JacquesCarette: Oh well.

JacquesCarette (approved):

PR #269, opened 2021-04-13
Products of monoidal categories are again monoidal

sstucki:
... and similarly for braided and symmetric monoidal categories.

Depends on #268.


JacquesCarette (approved):
Nice, thanks. Annoying how tedious `Monoidal.Construction.Product` is. This is where the automation in other systems does seem to have advantages.

PR #274, opened 2021-04-25
Rework the definition of the Simplex Category

TOTBWF:
## Patch Description

This PR reworks the definition of how we define Δ. The definition we have now is workable for simple stuff, but you run into trouble very fast when you start trying to define certain simplicial sets. Most of the literature defines simplicial sets by how they act on the face/degeneracy maps, and for good reason! This is much simpler to think about than defining their action on some random monotonic map. However, the way we have Δ set up now is not amenable to this style of definition.

To fix this, I've defined the morphisms of Δ to be formal compositions of face and degeneracy maps. However, equality is still defined via their semantics as maps between finite ordinals, which lets us avoid having to deal with the simplicial identities.

## Notes

I've tried something a bit different with `Categories.Category.Instance.Simplex` this time around. Instead of having a giant note at the beginning, I've laid out the code/explanations using a literate programming sort of style. If this isn't in the style of the library, feel free to request a change!

### DIFF
--- a/src/Categories/Category/Instance/Simplex.agda
+++ b/src/Categories/Category/Instance/Simplex.agda
@@ -6,7 +6,7 @@ module Categories.Category.Instance.Simplex where
 
 open import Level using (0ℓ)
 open import Data.Product
-open import Data.Fin.Base using (Fin; zero; suc; _≤_; _<_; inject₁)
+open import Data.Fin.Base using (Fin; zero; suc; _≤_; _<_; inject₁; punchIn)
 open import Data.Nat.Base using (ℕ; zero; suc; z≤n; s≤s)
 open import Function renaming (id to idF; _∘_ to _∙_)
 
@@ -42,21 +42,20 @@ data _Δ⇒_ : ℕ → ℕ → Set where
   _⊚_ : ∀ {l m n} → m Δ⇒ n → l Δ⇒ m → l Δ⇒ n
 
 -- However, this raises some tricky questions about equality of morphisms. It is tempting
--- to use the Simplical Identities[2]:
+-- to use the Simplical Identities[2] directly as our notion of equality:
 --   * δᵢ ∘ δⱼ = δⱼ₊₁ ∘ δᵢ if i ≤ j
 --   * σⱼ ∘ σᵢ = σᵢ ∘ σⱼ₊₁ if i ≤ j
 --   * σⱼ ∘ δᵢ = δᵢ ∘ σⱼ₋₁ if i < j
 --   * σⱼ ∘ δᵢ = id        if i = j or i = j + 1
 --   * σⱼ ∘ δᵢ = δᵢ₋₁ ∘ σⱼ if j + 1 < i
 --
--- However, we can do better than this. Instead, we appeal to the denotational
--- semantics of our formal chains of morphisms. In this case we interpret
--- our morphisms as maps between finite ordinals:
+-- However, working with these isn't exactly pleasant and will prove these
+-- as theorems later. Instead of quotienting by these identities,
+-- we appeal to the denotational semantics of our formal chains of morphisms.
+-- In this case we interpret our morphisms as maps between finite ordinals:
 
 face : ∀ {n} → Fin (ℕ.suc n) → Fin n → Fin (ℕ.suc n)
-face Fin.zero    k           = Fin.suc k
-face (Fin.suc i) Fin.zero    = Fin.zero
-face (Fin.suc i) (Fin.suc k) = Fin.suc (face i k)
+face = punchIn
 
 degen : ∀ {n} → Fin n → Fin (ℕ.suc n) → Fin n
 degen Fin.zero    Fin.zero    = Fin.zero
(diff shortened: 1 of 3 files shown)


src/Categories/Category/Instance/Simplex.agda
```diff
@@ -1,67 +1,152 @@
...
+infixr 9 _⊚_
+
+data _Δ⇒_ : ℕ → ℕ → Set where
+  ε  : ∀ {n} → n Δ⇒ n
+  δ  : ∀ {n} → (i : Fin (suc n)) → n Δ⇒ (suc n)
+  σ  : ∀ {n} → (j : Fin n) → (suc n) Δ⇒ n
+  _⊚_ : ∀ {l m n} → m Δ⇒ n → l Δ⇒ m → l Δ⇒ n
+
+-- However, this raises some tricky questions about equality of morphisms. It is tempting
+-- to use the Simplical Identities[2]:
```
JacquesCarette, line 45: Add a comment here that you will prove these identities `face` and `degen` below?

src/Categories/Category/Instance/Simplex.agda
```diff
@@ -1,67 +1,152 @@
...
-  ; equiv     = record
-    { refl  = λ _ → refl
-    ; sym   = λ eq x → sym (eq x)
-    ; trans = λ eq₁ eq₂ x → trans (eq₁ x) (eq₂ x)
+  { Obj = ℕ
+  ; _⇒_ = _Δ⇒_
+  ; _≈_ = _≗_
+  ; id = ε
+  ; _∘_ = _⊚_
+  ; assoc = Δ-eq refl
```
JacquesCarette, line 95: That these are then 'just true' is very nice.

src/Categories/Category/Instance/Simplex.agda
```diff
@@ -1,67 +1,152 @@
...
-degeneracy i = degeneracy-map i , degeneracy-mono i
+-- δᵢ ∘ δⱼ = δⱼ₊₁ ∘ δᵢ if i ≤ j
+face-comm : ∀ {n} {i j : Fin (suc n)}  → i ≤ j → δ (inject₁ i) ∘ δ j ≈ δ (suc j) ∘ δ i
+face-comm {_} {zero}  {j}     z≤n      = Δ-eq refl
+face-comm {_} {suc i} {suc j} (s≤s le) = Δ-eq (λ { {zero} → refl ; {suc x} → cong suc (Δ-pointwise (face-comm le)) })
+
+-- σⱼ ∘ σᵢ = σᵢ ∘ σⱼ₊₁ if i ≤ j
+degen-comm : ∀ {n} {i j : Fin n} → i ≤ j → σ j ∘ σ (inject₁ i) ≈ σ i ∘ σ (suc j)
+degen-comm {_} {zero}  {zero}  z≤n      = Δ-eq λ { {zero} → refl ; {suc x} → refl }
+degen-comm {_} {zero}  {suc j} z≤n      = Δ-eq λ { {zero} → refl ; {suc x} → refl }
```
JacquesCarette, line 121: I think you should be able to split on `i ≤ j` directly, without the implicits and still get this?

src/Categories/Category/Instance/Simplex.agda
```diff
@@ -1,67 +1,152 @@
...
-degeneracy-mono Fin.zero    {Fin.zero}  {_}         _        = z≤n
-degeneracy-mono Fin.zero    {Fin.suc _} {Fin.suc _} (s≤s le) = le
-degeneracy-mono (Fin.suc i) {Fin.zero}  {_}         _        = z≤n
-degeneracy-mono (Fin.suc i) {Fin.suc _} {Fin.suc _} (s≤s le) = s≤s (degeneracy-mono i le)
-
-degeneracy : ∀ {n} → Fin n → ℕ.suc n ⇒ n
-degeneracy i = degeneracy-map i , degeneracy-mono i
+-- δᵢ ∘ δⱼ = δⱼ₊₁ ∘ δᵢ if i ≤ j
+face-comm : ∀ {n} {i j : Fin (suc n)}  → i ≤ j → δ (inject₁ i) ∘ δ j ≈ δ (suc j) ∘ δ i
+face-comm {_} {zero}  {j}     z≤n      = Δ-eq refl
```
JacquesCarette, line 115: I think you should be able to split on `i ≤ j` directly, without the implicits and still get this?

src/Categories/Category/Instance/Simplex.agda
```diff
@@ -1,67 +1,152 @@
...
+degen-face-comm {_} {zero}  {j}     (s≤s le) = Δ-eq refl
+degen-face-comm {_} {suc i} {suc j} (s≤s le) = Δ-eq (λ { {zero} → refl ; {suc x} → cong suc (Δ-pointwise (degen-face-comm le)) })
+
+-- σⱼ ∘ δᵢ = id        if i = j
+degen-face-id : ∀ {n} {i j : Fin n} → i ≡ j → σ j ∘ δ (inject₁ i) ≈ id
+degen-face-id {_} {zero}  {zero}  refl = Δ-eq refl
+degen-face-id {_} {suc i} {suc i} refl = Δ-eq (λ { {zero} → refl ; {suc x} → cong suc (Δ-pointwise (degen-face-id {i = i} refl)) })
+
+-- σⱼ ∘ δᵢ = id        if i = j + 1
+degen-face-suc-id : ∀ {n} {i : Fin (suc n)} {j : Fin n} → i ≡ suc j → σ j ∘ δ i ≈ id
```
JacquesCarette, line 135: maybe make `j` explicit in the statement here? I think it is more natural to split on explicit arguments.

JacquesCarette (changes requested):
Generally, this is fantastic.  Just a few minor things that would further improve the code.

Also: are all those `Lift`s really needed/wanted? I know that sometimes, they really are - is this one of those cases?

src/Categories/Category/Instance/Simplex.agda
```diff
@@ -1,67 +1,152 @@
...
+--   * σⱼ ∘ δᵢ = δᵢ₋₁ ∘ σⱼ if j + 1 < i
+--
+-- However, we can do better than this. Instead, we appeal to the denotational
+-- semantics of our formal chains of morphisms. In this case we interpret
+-- our morphisms as maps between finite ordinals:
+
+face : ∀ {n} → Fin (ℕ.suc n) → Fin n → Fin (ℕ.suc n)
+face Fin.zero    k           = Fin.suc k
+face (Fin.suc i) Fin.zero    = Fin.zero
+face (Fin.suc i) (Fin.suc k) = Fin.suc (face i k)
```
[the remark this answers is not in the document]
TOTBWF: Good to know! Perhaps it makes sense to add `degen` to the stdlib as well under a less domain-specific name.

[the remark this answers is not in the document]
JacquesCarette: I agree - if you can figure out a good name. I couldn't...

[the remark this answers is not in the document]
TOTBWF: https://github.com/agda/agda-stdlib/pull/1492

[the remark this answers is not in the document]
JacquesCarette: I understand that it will make sense to still have the names `face` and `degen` in this file, right? So there is no need to wait for the standard library to merge these in. The definition of `face` can be shortened now, and that of `degen` later.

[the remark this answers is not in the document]
TOTBWF: Yeah, I was still going to give them new names here for clarity, so we aren't blocked on that PR.

src/Categories/Category/Instance/Simplex.agda
TOTBWF: Agda will complain about some unification problem getting stuck if you try to do this sadly.

src/Categories/Category/Instance/Simplex.agda
TOTBWF: Same deal as above, some gnarly unification problem error.

src/Categories/Category/Instance/Simplex.agda
TOTBWF: Should we do this with `i` as well, and for every one of these proofs?

src/Categories/Category/Instance/Simplex.agda
JacquesCarette: Right, it probably needs to know more for `σ` to reduce. Then maybe `i` and `j` should be explicit too, here and below.

src/Categories/Category/Instance/Simplex.agda
JacquesCarette: Any variable that we split on, because we need to know more about it for evaluation to progress, should probably be an explicit argument to these proofs. With a comment that explains why, since it might look like the equational argument might be enough to resolve the implicits.

TOTBWF:
The `Lift`s are  a preemptive measure, otherwise these would live at level `0`, which would make them a bit frustrating to use.

HuStmpHrrr:
this looks very nice to me

TOTBWF:
Is this good to go? I have some other stuff that I've been toying with that is related, but don't want to tack onto this PR.

JacquesCarette (approved):

PR #278, opened 2021-04-28
Functor composition and identity preserve braided/symmetric monoidality

sstucki:


JacquesCarette:
Is this independent of #276 and can be merged independently?  It looks like it, but I figured I'd ask to be safe.

sstucki:
Yes it's independent from #276.

JacquesCarette (approved):

PR #279, opened 2021-05-04
[WIP] Monoidal Category Tactic

TOTBWF:
## Patch Description

This PR implements the start of a monoidal category solver, based roughly off of the ideas found in [1].

## Notes
I'm almost over the finish line on this one, just have to show that normalization preserves equality for associators, and prove a couple of rather stubborn lemmas. Of note is `strict-⊗`, which is _not_ being well behaved. The proof ought to follow by some simple rewrites like `strict-∘`, but Agda does _not_ seem happy about that.

## References
1. http://www.cse.chalmers.se/~peterd/papers/Coherence_Monoidal.pdf

### DIFF
--- a/src/Categories/Tactic/Monoidal.agda
+++ b/src/Categories/Tactic/Monoidal.agda
@@ -1,0 +1,260 @@
+{-# OPTIONS --without-K --safe #-}
+
+-- A Solver for monoidal categories.
+-- Roughly based off of "Extracting a proof of coherence for monoidal categories from a formal proof of normalization for monoids",
+-- by Ilya Beylin and Peter Dybjer.
+module Categories.Tactic.Monoidal where
+
+open import Level
+open import Data.Product using (_,_)
+
+open import Relation.Binary.PropositionalEquality
+
+open import Categories.Category.Core using (Category)
+open import Categories.Category.Monoidal.Core using (Monoidal)
+
+import Categories.Morphism.Reasoning as MR
+
+--------------------------------------------------------------------------------
+-- Introduction:
+-- The basic idea of this solver is similar to a coherence theorem for
+-- monoidal categories. We are going to show that every single
+-- chain of coherence morphisms have some canonical form.
+-- Once we have done that, it is simply a matter of mapping two
+-- chains of coherence morphisms to their normal forms, and checking
+-- if the two are equal.
+
+module _ {o ℓ e} {𝒞 : Category o ℓ e} (𝒱 : Monoidal 𝒞) where
+
+  infixr 9 _∘′_
+
+  infixr 10 _⊗₀′_ _⊗₁′_
+
+  open Category 𝒞
+  open Monoidal 𝒱
+
+  open HomReasoning
+  open MR 𝒞
+
+  --------------------------------------------------------------------------------
+  -- A 'Word' reifies all the parenthesis/tensors/units of some object
+  -- in a monoidal category into a data structure
+  --------------------------------------------------------------------------------
+  data Word : Set o where
+    _⊗₀′_ : Word → Word → Word
+    unit′ : Word
+    _′    : Obj → Word
+
+  reify : Word → Obj
+  reify (w₁ ⊗₀′ w₂) = reify w₁ ⊗₀ reify w₂
+  reify unit′ = unit
+  reify (x ′) = x
+
+  private
+    variable
+      X Y Z   : Obj
+      A B C D : Word
+
+  -------------------------------------------------------------------------------- 
+  -- An 'Expr' reifies all unitors, associators and their compositions
+  -- into a data structure.
+  -------------------------------------------------------------------------------- 
+  data Expr : Word → Word → Set (o ⊔ ℓ) where
+    id′  : Expr A A
+    _∘′_ : Expr B C → Expr A B → Expr A C
+    _⊗₁′_ : Expr A C → Expr B D → Expr (A ⊗₀′ B) (C ⊗₀′ D)
+    α′   : Expr ((A ⊗₀′ B) ⊗₀′ C) (A ⊗₀′ (B ⊗₀′ C))
+    α⁻¹′ : Expr (A ⊗₀′ (B ⊗₀′ C)) ((A ⊗₀′ B) ⊗₀′ C) 
+    ƛ′   : Expr (unit′ ⊗₀′ A) A
+    ƛ⁻¹′ : Expr A (unit′ ⊗₀′ A)
+    ρ′   : Expr (A ⊗₀′ unit′) A
+    ρ⁻¹′ : Expr A (A ⊗₀′ unit′)
+    
+  -- Embed a morphism in 'Expr' back into '𝒞' without normalizing.
+  [_↓] : Expr A B → (reify A) ⇒ (reify B)
+  [ id′ ↓]    = id
+  [ f ∘′ g ↓] = [ f ↓] ∘ [ g ↓]
+  [ f ⊗₁′ g ↓] = [ f ↓] ⊗₁ [ g ↓]
+  [ α′ ↓]     = associator.from
+  [ α⁻¹′ ↓]   = associator.to
+  [ ƛ′ ↓]     = unitorˡ.from
+  [ ƛ⁻¹′ ↓]   = unitorˡ.to
+  [ ρ′ ↓]     = unitorʳ.from
+  [ ρ⁻¹′ ↓]   = unitorʳ.to
+
+  -- Invert a composition of coherence morphisms
+  invert : Expr A B → Expr B A
+  invert id′ = id′
+  invert (f ∘′ g) = invert g ∘′ invert f
+  invert (f ⊗₁′ g) = invert f ⊗₁′ invert g
+  invert α′ = α⁻¹′
+  invert α⁻¹′ = α′
+  invert ƛ′ = ƛ⁻¹′
+  invert ƛ⁻¹′ = ƛ′
+  invert ρ′ = ρ⁻¹′
+  invert ρ⁻¹′ = ρ′
+
+  -- Reassociate all the tensors to the right.
+  -- 
+  -- Note [reassoc + lists]:
+  -- We could use a list here, but this version is somewhat nicer,
+  -- as we can get things like right-identity for free.
+  reassoc : Word → (Word → Word)
+  reassoc (w₁ ⊗₀′ w₂) rest = reassoc w₁ (reassoc w₂ rest)
+  reassoc unit′ rest = rest
+  reassoc (x ′) rest = (x ′) ⊗₀′ rest
+
+  -- This is the key proof of the entire tactic.
+  -- 'coherence e' proves that all of our coherence morphisms
+  -- in 'e' are not required after reassociation, as they are on-the-nose equal.
+  coherence : Expr A B → (X : Word) → reassoc A X ≡ reassoc B X
+  coherence id′                         X = refl
+  coherence (f ∘′ g)                    X = trans (coherence g X) (coherence f X)
+  coherence (_⊗₁′_ {A} {B} {C} {D} f g) X = trans (cong (reassoc A) (coherence g X)) (coherence f (reassoc D X))
+  coherence α′                          X = refl
+  coherence α⁻¹′                        X = refl
+  coherence ƛ′                          X = refl
+  coherence ƛ⁻¹′                        X = refl
+  coherence ρ′                          X = refl
+  coherence ρ⁻¹′                        X = refl
+
+  -- Place every word into a normal form
+  -- > nf ((W ′ ⊗₀′ X ′) ⊗₀′ (Y ′ ⊗₀′ Z ′))
+  --   W ′ ⊗₀ X ′ ⊗₀ Y ′ ⊗₀ Z ′ ⊗₀ unit′
+  nf : Word → Word
+  nf w = reassoc w unit′
+
+  -- Given some coherence morphism, build a morphisms between
+  -- the normal forms of it's domain and codomain.
+  -- This will be equal to the identity morphism.
+  strict : Expr A B → Expr (nf A) (nf B)
+  strict {A = A} {B = B} e = subst (λ X → Expr (reassoc A unit′) X) (coherence e unit′) id′
+
+  -- If we reassociate and tensor after that, we can find some coherence
+  -- morphism that removes the pointless unit.
+  slurp : ∀ (A B : Word) → Expr (reassoc A unit′ ⊗₀′ B) (reassoc A B)
+  slurp (A ⊗₀′ B) C = slurp A (reassoc B C) ∘′ (id′ ⊗₁′ slurp B C) ∘′ α′ ∘′ (invert (slurp A (reassoc B unit′) ⊗₁′ id′))
+  slurp unit′     B = ƛ′
+  slurp (x ′)     B = ρ′ ⊗₁′ id′
+
+  -- Coherence morphism witnessing the concatentation of normal forms.
+  nf-homo : ∀ (A B : Word) → Expr (nf A ⊗₀′ nf B) (nf (A ⊗₀′ B))
+  nf-homo A B = slurp A (reassoc B unit′)
+
+  -- Build a coherence morphism out of some word into it's normal form.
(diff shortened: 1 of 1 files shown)


sstucki:
This is very cool!

I'd love to help out but I'm super busy this and next week. What I can offer in the meantime is some observations:

 1. I'm not surprised that you have unresolved metas in exactly the places you do because you have not yet used any of the coherence laws (triangle, pentagon) that make it possible to prove coherence for monoidal categories. I suspect these will be crucial in the missing steps/proofs though I'm not 100% certain since the subset of the free monoidal category you're using is somewhat degenerate (you don't have any "non-trivial" morphisms).
 2. Why use `slurp`? It's nice that you can proof the extra `unit` isn't necessary, but who cares? A normal form with the extra `unit` is as good as one without, no? I suspect the `slurp`s could get in the way in some of the proofs.
 3. Is the use of `rewrite` necessary? I'm surprised it worked, I thought `rewrite` was not `--safe`.
 4. I was surprised that you added syntax for objects (`Word`) and for morphisms (`Expr`) but not for equations on morphisms. If you had those, your syntax would be the free monoidal category over a discrete category (with the objects of 𝒞). I was expecting you to prove `preserves-≈` in that category (i.e. w.r.t. to the "synthetic" equality) rather than in the embedding of that category into 𝒞. Maybe that would simplify some proofs by removing the extra `[_↓]` in the types. Maybe not. But worth considering if you remain stuck.
 5. I would actually have gone a step further and added yet another syntax for normal forms — the free _strict_ monoidal category over the objects of 𝒞 — but you don't seem to need it. I suspect if one added support for non-trivial morphisms (i.e. anything other than combinations of associators and unitors) this might become necessary because the normal forms are then more complicated. But I'm just speculating.
 6. Nitpick: I find the name "coherence morphism" a bit misleading. I think it should be "coherent morphism" because these morphisms don't ensure coherence, they are coherent (it's the coherence laws that ensure their coherence). Anyway, that's a detail.

TOTBWF:
Yeah, I roughly see where I need to use the associators and unitors. The reason I am using this degenerate version is that during reflection, I plan on splitting the morphisms into the ones that just do coherence things, and non trivial ones. If we included the non trivial ones in our subcategory, then the key theorem `coherence : Expr A B → (X : Word) → reassoc A X ≡ reassoc B X` woudn't be true!

As for `rewrite`, it's a special version of `with` abstraction that does some extra stuff for us (See https://agda.readthedocs.io/en/v2.6.1.3/language/with-abstraction.html#rewrite). There is another _very unsafe_ feature with the same name, but that uses pragmas.

I originally tried working with synthetic equality actually, but it didn't seem to be buying me a whole lot. Perhaps it could help with versions of the problematic `strict-⊗` lemma.

Adding a syntax for normal forms is a great idea, I hadn't considered it! The whole proof is sort of a bizzare version of a coherence theorem anyways, so it makes sense that an explicit version of a stricit moniodal category could be useful. Will give this a go :)

WRT to naming, good point. I suppose just writing unitors/associators is clearer.

sstucki:
I managed to make some progress by using an explicit representation of normal forms. It allowed me to solve the case for the tensor product `_⊗₁′_` in `preserves-≈` and re-prove the other cases you had. I don't know how best to share this (is there a way to collaborate on PRs?) so I pushed it into a branch in my fork of the library:
https://github.com/sstucki/agda-categories/blob/27dc6adf0332a243747f83014b98fc3e7ee7786d/src/Categories/Tactic/Monoidal.agda

Hope this is helpful.

JacquesCarette:
I think it is possible to add commits to this PR, at least for people who have push access to the repo (which I believe you do @sstucki ).  Though it might make sense to push to you another branch that is started from this one, to let @TOTBWF do the merge, if he so chooses.

TOTBWF:
@sstucki This is great! I think we might have ourselves a solver pretty soon :). WRT commiting to the branch, this is the most annoying thing about git IMO. I'll grab those commits off your branch and integrate them into mine.

TOTBWF:
Looks like using explicit equalities might be the right move after all. This lemma would be waaaaaay easier if we could pattern match on equality proofs.
```
invert-resp-≈ : ∀ (f g : Expr A B) → f ≈↓ g → invert f ≈↓ invert g
```

sstucki:
Can't you use the fact that `f` and `inverse f` form an iso? Lift `f ≈↓ g` to an equality on isos, the project out the component on the inverses.

sstucki:
> Moving forward, you should probably have your branch track mine so that the histories are the same (makes merges easier)

Good idea, I'll do that!

TOTBWF:
Ok, more progress. `⌊⌋-α` is done, and all that remains of the associator case is a gnarly reassociation. What are everyones thoughts on using the category tactic to implement this tactic? They both will be importing reflection code at the end of the day, so the dependency tree doesn't get any deeper.

JacquesCarette:
Because of the stated reason, using the category tactic here seems reasonable.

sstucki:
> What are everyones thoughts on using the category tactic to implement this tactic? They both will be importing reflection code at the end of the day, so the dependency tree doesn't get any deeper.

Actually, a lot of the code in this module is independently useful. It's a weak version of the coherence theorem for (discrete) monoidal categories. It just so happens that this proof is constructive so we can use it as an algorithm to "solve" equations in such categories. I think there's a point exposing that part of the module and making it independent of tactics code.

If the issue is just the "gnarly reassociation" for the associator case, then we can either prove it by hand (shouldn't be all that bad) or use the category tactic to generate a proof and substitute it in. Apparently the command for this is `C-c C-m` but I have never tried using it.

TOTBWF:
The proofs are finally done! Now all that remains is writing the reflection code, and figuring out what needs to get factored out.

@sstucki After checking through the relevant sections of CWM, I think this is actually the proof that Maclane has. Specifically, that coherence theorem states that "every diagram built out of unitors, associators, and multiplication commutes"[1], which is exactly what this proves.

## References
[1] Categories for the Working Mathematician VII.2

src/Categories/Tactic/Monoidal.agda
```diff
@@ -0,0 +1,444 @@
...
+  invert-iso f = record
+    { isoˡ = invert-isoˡ f
+    ; isoʳ = invert-isoʳ f
+    }
+
+  NfWord : Set o
+  NfWord = List Obj
+
+  data NfExpr : NfWord → NfWord → Set o where
+    idⁿ : ∀ {N} → NfExpr N N
```
JacquesCarette, line 157: I would be tempted, to avoid subst-juggling, to try
```agda
  idⁿ : ∀ {N M} → N == M → NfExpr N M
```
There's also some alternate way that @sstucki wrote to do this, but I can't recall where it is...

sstucki: @JacquesCarette I think you're thinking of `hid` which itself uses subst-juggling, so not sure it's helpful here.

But I was thinking: the use of subst in this module could maybe be avoided if the normal forms of the unitors and associators were defined by recursion on the domains following the same structure as the unit/associativity laws for lists themselves. That might also simplify the proofs of the corresponding "monoidality" laws which are proven by recursion and need helper lemmas to deal with the subst stuff. Anyway, just an idea, I don't know whether it would actually work out.

JacquesCarette: Right, `hid`. If all `subst`-juggling can be isolated to one place, that counts as a win.

TOTBWF:
Ok, I've done some more thinking on this, and I have a idea for how we can normalize non-discrete things. If we update our `Expr` to:
```adga
  data Expr : Word → Word → Set (o ⊔ ℓ) where
    id′  : Expr A A
    _∘′_ : Expr B C → Expr A B → Expr A C
    _⊗₁′_ : Expr A C → Expr B D → Expr (A ⊗₀′ B) (C ⊗₀′ D)
    α′   : Expr ((A ⊗₀′ B) ⊗₀′ C) (A ⊗₀′ (B ⊗₀′ C))
    α⁻¹′ : Expr (A ⊗₀′ (B ⊗₀′ C)) ((A ⊗₀′ B) ⊗₀′ C)
    ƛ′   : Expr (unit′ ⊗₀′ A) A
    ƛ⁻¹′ : Expr A (unit′ ⊗₀′ A)
    ρ′   : Expr (A ⊗₀′ unit′) A
    ρ⁻¹′ : Expr A (A ⊗₀′ unit′)
    [_↑] : X ⇒ Y → Expr (X ′) (Y ′)
```
and our normal forms to something like:
```agda
  data NfExpr : NfWord → NfWord → Set (o ⊔ ℓ) where
    emptyⁿ : NfExpr [] []
    consⁿ  : ∀ {X Y} {N M} → X ⇒ Y → NfExpr N M → NfExpr (X ∷ N) (Y ∷ M)
```

Then things _seem_ to work out. Most of the proofs seem to fall out pretty naturally by induction, though I haven't reworked the monster that is `preserves-≈`.

One annoying thing is that we lose out on `invert`, but we can flip most of the structural morphisms we need by hand. This seems like a totally acceptable price to pay for a better tactic though!

JacquesCarette:
Everything else in `Expr` represents an iso, so why not for
```agda
   [_↑] : X ⇒ Y → Expr (X ′) (Y ′)
```
as well? You should make the lift be on an iso too. You really do want that for `NfExpr` too, no? Then you'll get `invert` back too.

sstucki:
This is definitely a big step forward, but it doesn't go all the way. The issue is that the domain and codomain of `[_↑]` should really be words because you'd want to be able to represent things like `id ⊗ f ∘ g ⊗ id` where `f : B ⊗ C ⇒ X` and `g : Y ⇒ A ⊗ B`, otti.e. where `f`and `g` are "shifted" w.r.t. each other. I suspect this would complicate normal forms a lot though. But I might be wrong. You've already come much further than I thought would be possible!

In any case, the tactic would already be enormously helpful even in the form you propose, because one could solve string diagrams "layer by layer" with a few manual proofs to merge and split these layers, when necessary.

sstucki:
> You should make the lift be on an iso too.

Actual string diagrams that contain arrows other than the monoidal maps (unit, assoc, etc.) need not be invertible. So there's no reason to think you'd only be embedding isos, is there?

TOTBWF:
We can still represent those morphisms with this encoding actually!
```
  example : ∀ {V W X Y Z} (f : W ⊗₀ X ⇒ Y) → (g : Z ⇒ V ⊗₀ W) → Expr ((Z ′) ⊗₀′ ((W ⊗₀ X) ′)) (((V ⊗₀ W) ′) ⊗₀′ (Y ′))
  example f g = id′ ⊗₁′ [ f ↑] ∘′ [ g ↑]  ⊗₁′ id′
```

The trick here is that we are "quoting" the objects `W ⊗₀ X` and `(W ⊗₀ X)`

JacquesCarette:
The `Expr` without `[_↑]` could only represent isos. That's why `invert` could exist.

sstucki:
> The trick here is that we are "quoting" the objects `W ⊗₀ X` and `V ⊗₀ W`

I don't understand how the composition can type check when different parts of the domain and codomain are quoted.

TOTBWF:
Yeah, that example worked out fine by coincidence. You run into trouble when you want to start composing onto associators/unitors. You were definitely right when you stated that the domain/codomain of the quoted morphisms ought to be reified words.

sstucki:
Yea, what I had in mind was
```
   Y     C
   |     |
[  g  ]  |
 |   |   |
 |  [  f  ]
 |     |
 A     X
```

TOTBWF:
Ok, I have a possible solution:

During reflection, we insert quoting/unquoting morphisms when needed:
```
    quot   : Expr A ((reify A) ′)
    unquot : Expr ((reify A) ′) A
```

These can read-back into `id` morphisms

sstucki:
I think the fundamental problem is certain diagrams just cannot be 'linearized' into a single list of morphisms. They need some form of tiling with multiple levels. Here's another version of the example I gave above with an explicit associator:
```
   Y     C
   |     |
[  g  ]  |  
 |   |   |
[  assoc  ]
 |   |   |
 |  [  f  ]
 |     |
 A     X
```
I cannot see how you'd turn that into a single layer. But maybe there's a way to turn it into a multi-layer normal form somehow.

JacquesCarette:
Would there be something to learn from @gelisam 's [premonoidal](https://github.com/gelisam/premonoidal) solver?

gelisam:
The monoidal version of the problem is harder than the premonoidal version, so it is I who has the most to learn from this conversation :)

In particular, while I do have a definition for a free premonoidal category in that repo, I have not yet made a solver out of it.

What distinguishes a premonoidal category from a monoidal category is that the law `(f *** id) >>> (id *** g) = f *** g = (id *** g) >>> (f *** id)`

      |     |         |     |         |     |
    [ f ]   |         |     |         |   [ g ]
      |     |    =  [ f ] [ g ]  =    |     |
      |   [ g ]       |     |       [ f ]   |
      |     |         |     |         |     |

is required to hold in a monoidal category but not in a premonoidal category (in fact there is no `(***)` in a premonoidal category, only the special cases `first f = f *** id` and `second f = id *** f`).

When looking for a definition for a free X, e.g. the free monoidal category, I start from a representation for the AST and I apply all the laws to reduce all the degrees of freedom. Above, we can see that the `f` and the `g` are free to slide past each other; by applying the corresponding law, I eliminate that freedom and force all the free-sliding blocks all the way to one side, say, the bottom. The end result is that the canonical representation looks like a tetris pile: blocks stacked on top of each other, with some overhangs.

      |        |            |        |
      |   [    f    ]       |        |
      |     |     |         |        |
    [ g ]   |     |         |        |
      |     |   [ h ]       |        |
      |     |     |    =    |        |
      |   [    i    ]       |        |
      |     |     |         |        |
    [    j    ]   |         |   [    f    ]
         |        |         |     |   [ h ]
         |      [ k ]     [ g ] [    i    ]
         |        |       [    j   ]  [ k ]

The canonical representation for the premonoidal case is much easier: the representation on the left is already canonical. It's just a list of atoms, shifted so that it only affects a subset of the strings at that height of the string diagram.

gelisam:
Taking a closer look at this branch, it looks to me like you don't support [atoms](https://github.com/agda/agda-categories/blob/b4da48f13545646422a0e0475c33b80e1dec1ff2/src/Categories/Tactic/Category.agda#L45) at all, only the structural bits such as the associators, unit-introduction, and unit elimination? That explains why your only normal form is [the identity](https://github.com/agda/agda-categories/pull/279/files#diff-97f610962ee655ed2182a8265630d06912514dd1ad97d3c6969fa1f5403e4da5R157)! Well, that certainly makes the problem a lot simpler.

And yet, from what I can see in the rest of this thread, it has already been quite a challenge! I guess I was wrong to think that the main challenge when writing this kind of solver was to figure out what the canonical representations look like, and that the rest (converting to and from the canonical representation, proving that an expression is equal to its canonical representation, writing a macro reflecting the type into an expression) would be routine.

gelisam:
> Maybe it's worth defining Free Monoidal Categories in a separate PR.

Where is your definition of a Free Monoidal Category? I'd be curious to see how yours compares to my tetris pile approach. I don't see it in this nor any recent PR.

JacquesCarette:
So is this completely stalled then?

TOTBWF:
I've been pretty busy as of late (started a new job and moved countries) so I haven't had the time to think about this too hard 🙁. Hopefully I'll have some time to devote to this soon!

JacquesCarette:
Those are rather good reasons for not having time! Hope it all went smoothly.

JacquesCarette:
I take it that this is still stalled (and there was no progress on something similar over in 1lab?)

gelisam:
We're currently discussing this in the Monoidal Café discord. We found a flaw in my Tetris pile idea from [earlier in this thread](https://github.com/agda/agda-categories/pull/279#issuecomment-844036944): it doesn't support scalars (morphisms from unit to unit), which commute with each other but not with identity arrows. Word on the street is that @pigworker has found a complicated normal form for monoidal categories he calls "staircases", but it's not ready to be published yet.

JacquesCarette:
I can't find the Monoidal Café discord - could I get an invite?

zanzix:
I was summoned here by @gelisam. Here is an invite to the Monoidal Cafe, feel welcome to join!   
https://discord.gg/kswesAKQ

PR #280, opened 2021-05-26
Product functors are (braided/symmetric) monoidal

sstucki:
... if they go between (braided/symmetric) monoidal categories.

### DIFF
--- a/Everything.agda
+++ b/Everything.agda
@@ -281,6 +281,7 @@ import Categories.Functor.Instance.Twisted
 import Categories.Functor.Limits
 import Categories.Functor.Monoidal
 import Categories.Functor.Monoidal.Braided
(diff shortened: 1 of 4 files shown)


JacquesCarette (approved):
Wow, 495 lines of what is basically completely straightforward code, except for ~5. This is going to become my go-to example for wanting some meta-programming.

It looks ready - is it?

sstucki:
Yep it should be ready. I checked using `make test` locally, so I expect Travis to be happy too.

Yes, it's all boilerplate. It would be nice to avoid doing these things by hand. There seems to be a lot of boilerplate stuff for monoidal categories in general...

JacquesCarette:
My guess is that this is true of many "lifting" operations, where properties in the base case lift across a nice operation. The same probably will happen from some unary functors too.

PR #290, opened 2021-06-30
Add Cartesian multicategories

laMudri:
While general multicategories have proved to be a pain to implement, [Cartesian multicategories](https://ncatlab.org/nlab/show/cartesian+multicategory) are easy. We can carry over techniques from simply typed λ-calculus, in particular simultaneous substitution. Indeed, the contents of this pull request give the machinery to derive the category of contexts and substitutions from some basic properties of the actions of substitutions.

Things left to do (along with general cleaning up and responding to comments):

- [x] Populate `HomReasoning` and `HomˢReasoning` with similar stuff as in `Category`.
- [ ] Move `_∈_` properties over to stdlib.


JacquesCarette:
That's looking really great!  Is this still WIP, or is it ready to go?

laMudri:
I'd like to at least give it a once-over. Do you have any opinions on the things that should be in stdlib rather than here, and getting versions synced up, &c? (Funnily enough, this kind of problem has been a topic in the SPLS pub.)

laMudri:
Also, I think we should eventually axiomatise all of the required properties about `List` and `_∈_`. The only challenge I see there is what to do about `prod`, `⟨_⟩Π`, and `π[_]`, but maybe the literature has an answer (or we just axiomatise everything as-is).

src/Categories/Multi/Category/Cartesian.agda
```diff
@@ -0,0 +1,115 @@
...
+    ∘-resp-≈ : ∀ {Γ Δ A} {f f′ : Δ ⇒ A} {σ σ′ : Γ ⇒ˢ Δ} →
+      f ≈ f′ → σ ≈ˢ σ′ → f ∘ σ ≈ f′ ∘ σ′
+
+    identityˡ : ∀ {Γ Δ A} {i : A ∈ Δ} {σ : Γ ⇒ˢ Δ} → id i ∘ σ ≈ σ i
+    identityʳ : ∀ {Γ A} {f : Γ ⇒ A} → f ∘ idˢ ≈ f
+    identity² : ∀ {Γ A} {i : A ∈ Γ} → id i ∘ idˢ ≈ id i
+    assoc : ∀ {Γ Δ Θ A} {f : Θ ⇒ A} {σ : Δ ⇒ˢ Θ} {τ : Γ ⇒ˢ Δ} →
+      f ∘ σ ∘ τ ≈ f ∘ (σ ∘ˢ τ)
+    sym-assoc : ∀ {Γ Δ Θ A} {f : Θ ⇒ A} {σ : Δ ⇒ˢ Θ} {τ : Γ ⇒ˢ Δ} →
+      f ∘ (σ ∘ˢ τ) ≈ f ∘ σ ∘ τ
```
laMudri, line 69: Does `sym-assoc` achieve anything? It exists in `Category` so as to support `op` being a definitional involution. But there is no `op` for multicategories. There's an `op` for polycategories, but then I think the composition is so different that proofs would never just pass through.

Similar question for `identity²`. It at least feels like it still has a place, intermediate between the other two identity laws. But I can't tell whether any benefits based on that are worth the extra field.

JacquesCarette: I would omit them until there's a compelling reason to add them. I'm sure some kinds of extra laws will arise when some things ought to be definitionally true - I just don't know what they are (yet).

src/Categories/Multi/Category/Cartesian.agda
```diff
@@ -0,0 +1,115 @@
...
+open import Data.List hiding ([_])
+open import Data.List.Membership.Propositional
+open import Data.Wrap
+open import Level
+open import Relation.Binary using (Rel; Setoid; IsEquivalence)
+import Relation.Binary.Reasoning.Setoid as SetoidR
+
+infix 4 [_]_≈ᵉ_
+
+Env : ∀ {o ℓ} {X : Set o} (V : X → Set ℓ) (Δ : List X) → Set (o ⊔ ℓ)
```
JacquesCarette, line 25: I think this file could do with a lot more comments, giving a bit more intuition. While a lot of category theory is well documented in a lot of textbooks, items like cartesian multicategories only really exist in research papers. Plus, you're drawing intuition from both PL and logic, which isn't always going to be clear to all readers.

I would actually be happy to provide such documentation - but I'd need access to your fork to do so.

src/Categories/Multi/Category/Cartesian.agda
```diff
@@ -0,0 +1,115 @@
...
+Env V Δ = ∀ {y} → y ∈ Δ → V y
+
+[_]_≈ᵉ_ : ∀ {o ℓ e} {X : Set o} {V : X → Set ℓ} {Δ} →
+  (∀ {x} → Rel (V x) e) → Rel (Env V Δ) (o ⊔ e)
+[_]_≈ᵉ_ = Wrap λ _≈_ ρ σ → ∀ {y} (i : y ∈ _) → ρ i ≈ σ i
+
+record CartesianMultiCategory (o ℓ e : Level) : Set (suc (o ⊔ ℓ ⊔ e)) where
+  infix 4 _≈_ _≈ˢ_ _⇒_ _⇒ˢ_
```
JacquesCarette, line 33: For example, does the superscript 's' *mean* 'simultaneous'? That would be good to write down.

src/Categories/Multi/Category/Cartesian/Representable.agda
```diff
@@ -0,0 +1,174 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Multi.Category.Cartesian.Representable where
+
+open import Data.List hiding ([_])
```
JacquesCarette, line 5: I prefer to use `using` as much as possible. It actually makes a non-trivial difference in the size of .agdai files, and from there to other things. But if basically everything is used, it's ok to leave it off.

src/Categories/Multi/Category/Cartesian/Representable.agda
```diff
@@ -0,0 +1,174 @@
...
+open import Categories.Category.Cartesian
+import Categories.Category.Cartesian.Properties as ×P
+open import Categories.Multi.Category.Cartesian
+
+private
+  variable
+    o ℓ e : Level
+
+module Extras {a} {A : Set a} where
+  -- TODO: contribute to stdlib
```
JacquesCarette, line 27: I think some of these have landed in stdlib now. But yes, do push these to there, they can be eliminated from here 'later'.

src/Categories/Multi/Category/Cartesian/Representable.agda
```diff
@@ -0,0 +1,174 @@
...
+    ; _⇒_ = λ Γ A → prod Γ ⇒ A
+    ; _≈_ = _≈_
+    ; id = π[_]
+    ; _∘_ = λ f σ → f ∘ ⟨ σ ⟩Π
+    ; equiv = equiv
+    ; ∘-resp-≈ = λ ff σσ → ∘-resp-≈ ff ⟨ σσ ⟩Π-resp-≈
+    ; identityˡ = λ {_} {_} {_} {i} → projectΠ i _
+    ; identityʳ = λ {Γ} → trans (∘-resp-≈ʳ (⟨π⟩Π {Γ})) identityʳ
+    ; identity² = λ {_} {_} {i} → projectΠ i _
+    ; assoc = λ {Γ Δ Θ A f σ τ} → begin
```
JacquesCarette, line 95: So it turns out that inline equational proofs take a lot longer to typecheck (and produce bloated results); best to pull these out, either in a prepended `private` block or using a `where` block below.

src/Categories/Multi/Category/Cartesian/Representable.agda
```diff
@@ -0,0 +1,174 @@
...
+      (f ∘ ⟨ σ ⟩Π) ∘ ⟨ τ ⟩Π  ≈⟨ assoc ⟩
+      f ∘ (⟨ σ ⟩Π ∘ ⟨ τ ⟩Π)
+        ≈⟨ refl⟩∘⟨ (uniqueness* {ys = Θ} λ i → begin
```
JacquesCarette, line 98: doubly-nested proofs end up hard to read too - could you denest?

src/Categories/Multi/Category/Cartesian/Representable.agda
```diff
@@ -0,0 +1,174 @@
...
+-- Cartesian category is the category of contexts and simultaneous
+-- substitutions.
+
+module _ (CM : CartesianMultiCategory o ℓ e) where
+  open CartesianMultiCategory CM
+  open Equiv
+  open Equivˢ
+
+  Free : ∃ \ C → Cartesian C
+  Free = record
```
JacquesCarette, line 130: It would be a lot more readable to define the `Category` instance and then the `Cartesian` instance in either `private` or `where blocks, and then use `_,_` to pair them.  Or maybe use the `CartesianCategory` bundle?

JacquesCarette (changes requested):
Sorry for taking so long to do a full review - the code arrived at a time when I was really busy, and it took me forever to get back to this.

This is great though (up to minor things), so it would be nice to actually get it in!

src/Categories/Multi/Category/Cartesian.agda
laMudri: Yes, though maybe there could be a better notation (`_⇒*_` might work, with the `*` being a mnemonic for lists and multiplicity, though maybe also clashing with reflexive-transitive closure). I'll note it down, in any case.

src/Categories/Multi/Category/Cartesian.agda
laMudri: I'll give you access anyway, but also set about documenting stuff a bit this weekend. If I don't do it this weekend, that'll be a cue that I need some help.

src/Categories/Multi/Category/Cartesian/Representable.agda
laMudri: How about a separate lemma in a `where` block?

src/Categories/Multi/Category/Cartesian/Representable.agda
laMudri: I don't spot any of them.

src/Categories/Multi/Category/Cartesian.agda
JacquesCarette: I agree that `*` is too overloaded and might lead to unnecessary confusion. Perhaps `∀` in the symbol? It's almost too bad you've already used `⇒Π`.

src/Categories/Multi/Category/Cartesian/Representable.agda
JacquesCarette: Ah indeed. I was fooled because they do exist for `Fin`, but not for `List`.

laMudri:
I suppose I'll need to rebase at some point to turn `Structure` into `Bundle`.

JacquesCarette:
Yes please!

JacquesCarette:
@laMudri  I'm looking back at this now: can you tell me what you think the status is, in your opinion?

laMudri:
Thanks for the reminder. agda/agda-stdlib#1756 should now be ready, so after that gets merged I should be able to delete those parts from this PR.

JacquesCarette:
I have a few cycles right now - what can I do to help push this through to completion?

laMudri:
I think moving the properties about list membership and appending over to stdlib (if they're not there already by now) was the main thing that needed doing.

JacquesCarette:
I have access to your repo, so I should be able to do this.

JacquesCarette:
I've pulled it in to a branch so that I can work on it more easily.

PR #292, opened 2021-07-04
FinSetoids is Cartesian

Taneb:
This is a WIP because:
- [ ] I want to implement Cocartesian also
- [ ] I want to look into doing something with ObjectRestriction to make this work a little more general

### DIFF
--- a/src/Categories/Category/Monoidal/Instance/FinSetoids.agda
+++ b/src/Categories/Category/Monoidal/Instance/FinSetoids.agda
@@ -1,0 +1,86 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Category.Monoidal.Instance.FinSetoids where
+
+open import Data.Fin.Base
+open import Data.Fin.Properties hiding (setoid)
+import Data.Nat.Base as ℕ
+open import Data.Product
+open import Data.Product.Relation.Binary.Pointwise.NonDependent
+open import Data.Unit.Polymorphic
+open import Function
+open import Function.Equality using (Π)
+open import Level using (Lift)
+open import Relation.Binary
+open import Relation.Binary.Reasoning.MultiSetoid
+open import Relation.Binary.PropositionalEquality
+
+open import Categories.Category.Cartesian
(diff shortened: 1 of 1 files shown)


JacquesCarette:
I wonder if there might be an opportunity for proving a theorem of the sort "if category X has property P, then sub-category Y of X (as witnesses by relation R) also has property P if P and R are related in way Z".  As sort of "R preserves P".

Is this what you meant by part 2 of the TODO?

JacquesCarette:
I wonder if there might be an opportunity for proving a theorem of the sort "if category X has property P, then sub-category Y of X (as witnesses by relation R) also has property P if P and R are related in way Z".  As sort of "R preserves P".

Is this what you meant by part 2 of the TODO?

Taneb:
> I wonder if there might be an opportunity for proving a theorem of the sort "if category X has property P, then sub-category Y of X (as witnesses by relation R) also has property P if P and R are related in way Z". As sort of "R preserves P".
> 
> Is this what you meant by part 2 of the TODO?

Yeah, that's exactly what I meant by part 2.

Taneb:
The main thing stopping me from working on this PR is that I don't know where to put the tools for constructing terminal objects and binary products and things like that for object restrictions of categories. Any suggestions? I was thinking something like `Categories.Category.Construction.ObjectRestriction.Properties.Terminal` etc but I'm not satisfied by that

JacquesCarette:
Why not put them directly in `Categories.Category.Construction.ObjectRestriction.Properties` ? Only if that grows very large would it need to be split (probably as you propose).

It might well be that some of the actual constructions should go elsewhere, and then just 'assembled' in the above?

JacquesCarette:
@Taneb It would be really nice to get this finished and merged in. What can I do to help?

Taneb:
@JacquesCarette thanks for reminding me that this PR exists! I need to take another look at it to refamiliarise myself, it's been a while

JacquesCarette:
Another ping on this one - it would be nice to finish. I have some cycles to help.

PR #307, opened 2021-08-26
Trivial cleanup of imports that were commented out.

sstucki:
The only reason I'm submitting this as a PR (rather than pushing it to master) is [this change](https://github.com/agda/agda-categories/commit/f3942db23f28deb1afd596e462040502195f2989#diff-dc71001c1756ed5198ddae507344f0d6eafaad4f4c69cf149439768544b7e270L20-L26) in `src/Categories/Category/Cocartesian/Bundle.agda`. It looks like there's some unfinished work there that was left in a comment. Did you leave it there intentionally @JacquesCarette?


JacquesCarette (approved):

JacquesCarette:
Indeed, I didn't finish that, but the extra items were not needed anywhere, so I figured I'd come back to it when there was a real need. It's fine to delete.

PR #316, opened 2021-10-29
add diagrammatic composition

tetrapharmakon:
diagrammatic composition of morphisms, i.e.

`f ; g : A --f--> B --g--> C`

tuns out to be useful in some proofs, and it usually makes very fast the process of dualising a thing into a co-thing (just put a co- in front of names, and replace \circ with `;` -which, again, is U+FF1B *not* U+003B)

### DIFF
--- a/src/Categories/Morphism.agda
+++ b/src/Categories/Morphism.agda
@@ -116,3 +116,7 @@ module ≅ = IsEquivalence ≅-isEquivalence
   ; _≈_           = _≅_
   ; isEquivalence = ≅-isEquivalence
   }
+
+-- diagrammatic composition of morphisms; the symbol is (U+FF1B), "Fullwidth Semicolon", not semicolon (U+003B)
+_；_ : {A B C : Obj} → (f : A ⇒ B) → (g : B ⇒ C) → (A ⇒ C)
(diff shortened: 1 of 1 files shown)


src/Categories/Morphism.agda
```diff
@@ -116,3 +116,7 @@ module ≅ = IsEquivalence ≅-isEquivalence
   ; _≈_           = _≅_
   ; isEquivalence = ≅-isEquivalence
   }
+
+-- diagrammatic composition of morphisms; the symbol is (U+FF1B), "Fullwidth Semicolon", not semicolon (U+003B)
+_；_ : {A B C : Obj} → (f : A ⇒ B) → (g : B ⇒ C) → (A ⇒ C)
```
JacquesCarette, line 121: I am super thorn. In the next release of Agda, there will be "fat semicolon", via `\z;` which is much nicer. So I'm kind of leaning towards waiting, as "full width semicolon" does make the typesetting look kind of ugly. And it's way too easy to mistake for `;` which is a problem for usability.

tetrapharmakon: Given this, it's probably a bad idea.

PR #318, opened 2021-11-03
Stub for representable profunctor

tetrapharmakon:
Any functor `F : C -> D` induces two "representable" profunctors 

`hom(F,1) : (c,d) \mapsto D(Fc,d)`

and 

`hom(1,F) : (d,c) \mapsto D(d, Fc)`

### DIFF
--- a/src/Categories/Functor/Profunctor.agda
+++ b/src/Categories/Functor/Profunctor.agda
@@ -12,40 +12,67 @@ open import Categories.Functor hiding (id)
 open import Categories.Functor.Bifunctor
 open import Categories.Functor.Hom
 
-Profunctor : ∀ {o ℓ e} {o′ ℓ′ e′} → Category o ℓ e → Category o′ ℓ′ e′ → Set _
-Profunctor {ℓ = ℓ} {e} {ℓ′ = ℓ′} {e′} C D = Bifunctor (Category.op D) C (Setoids (ℓ ⊔ ℓ′) (e ⊔ e′))
+module Profunctor {o ℓ e} (C : Category o ℓ e) (D : Category o ℓ e) where
 
-id : ∀ {o ℓ e} → {C : Category o ℓ e} → Profunctor C C
-id {C = C} = Hom[ C ][-,-]
+  open import Categories.Morphism.Reasoning D as MR
 
-private
-  variable
-    o ℓ e o′ ℓ′ e′ : Level
-    C : Category o ℓ e
-    D : Category o′ ℓ′ e′
+  Profunctor : ∀ {o ℓ e} {o′ ℓ′ e′} → Category o ℓ e → Category o′ ℓ′ e′ → Set _
+  Profunctor {ℓ = ℓ} {e} {ℓ′ = ℓ′} {e′} C D = Bifunctor (Category.op D) C (Setoids (ℓ ⊔ ℓ′) (e ⊔ e′))
 
--- stub for representable profunctors
-repˡ : (F : Functor C D) → Profunctor D C
-repˡ F = record
-  { F₀ = {!   !}
-  ; F₁ = {!   !}
-  ; identity = {!   !}
-  ; homomorphism = {!   !}
-  ; F-resp-≈ = {!   !}
-  }
-  where
-  F0 : Σ (Category.Obj C) (λ x → Category.Obj D) → Setoid (ℓ ⊔ ℓ′) (e ⊔ e′)
-  F0 (c , d) = record
-    { Carrier = {!   !}
-    ; _≈_ = {! _≈_   !}
-    ; isEquivalence = record { refl = {!   !} ; sym = {!   !} ; trans = {!   !} }
+  id : ∀ {o ℓ e} → {C : Category o ℓ e} → Profunctor C C
+  id {C = C} = Hom[ C ][-,-]
+
+  -- representable profunctor
+  -- hom(f,1)
+  repˡ : (F : Functor C D) → Profunctor {o} {ℓ} {e} {o} {ℓ} {e} D C
+  repˡ F = record
+    { F₀ = λ {(c , d) → record
+        { Carrier = D [ Functor.F₀ F c , d ]
+        ; _≈_ = D._≈_
+        ; isEquivalence = D.equiv
+        }}
+    ; F₁ = λ { {A = A0 , A1} {B = B0 , B1} (f , g) → record
+      { _⟨$⟩_ = λ x → g D.∘ x D.∘ Functor.F₁ F f
+      ; cong = λ x → D.∘-resp-≈ D.Equiv.refl (D.∘-resp-≈ x D.Equiv.refl)
+      }}
+    ; identity = λ {x} {y} {y'} y≈y' → begin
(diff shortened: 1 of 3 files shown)


src/Categories/Functor/Profunctor.agda
```diff
@@ -4,13 +4,91 @@ module Categories.Functor.Profunctor where
...
+      ; cong = λ x → begin _ ≈⟨ refl⟩∘⟨ x ⟩∘⟨refl ⟩ _ ∎
+      }
+    ; identity = λ {x} {y} {y'} y≈y' → begin
+        F₁ C.id ∘ y ∘ D.id ≈⟨ elimˡ D identity ⟩
+        y ∘ D.id             ≈⟨ D.identityʳ ⟩
+        y                    ≈⟨ y≈y' ⟩
+        y'                   ∎
+    ; homomorphism = λ { {f = f0 , f1} {g = g0 , g1} {x} {y} x≈y → begin
+        F₁ (g1 C.∘ f1) ∘ x ∘ f0 ∘ g0    ≈⟨ refl⟩∘⟨ x≈y ⟩∘⟨refl ⟩
+        F₁ (g1 C.∘ f1) ∘ y ∘ f0 ∘ g0    ≈⟨ Functor.homomorphism F ⟩∘⟨refl ⟩
```
JacquesCarette, line 78: `F.homomorphism`

src/Categories/Functor/Profunctor.agda
```diff
@@ -4,13 +4,91 @@ module Categories.Functor.Profunctor where
...
+        (F₁ g1 ∘ F₁ f1) ∘ y ∘ f0 ∘ g0 ≈⟨ D.assoc ⟩
+        F₁ g1 ∘ F₁ f1 ∘ y ∘ f0 ∘ g0       ≈⟨ refl⟩∘⟨ D.Equiv.sym assoc² ⟩
```
JacquesCarette, line 80: these 2 lines are some kind of pull/push.

JacquesCarette (changes requested):
One big thing, and then a few little fixes.

src/Categories/Functor/Profunctor.agda
```diff
@@ -89,19 +91,46 @@ module Profunctor {o ℓ e} {o′} (C : Category o ℓ e) (D : Category o′ ℓ
...
-    ; identityˡ = {!   !}
-    ; identityʳ = {!   !}
-    ; identity² = {!   !}
-    ; equiv = {!   !}
-    ; ∘-resp-≈ = {!   !}
-    }
\ No newline at end of file
+-- each Prof(C,D) is a category
+homProf : {o o′ ℓ e : Level} → (C : Category o ℓ e) → (D : Category o′ ℓ e) → Category _ _ _
+homProf {o} {o′} {ℓ} {e} C D = record
```
JacquesCarette, line 96: you don't use the levels below, maybe omit binding them needlessly here?

src/Categories/Functor/Profunctor.agda
```diff
@@ -89,19 +91,46 @@ module Profunctor {o ℓ e} {o′} (C : Category o ℓ e) (D : Category o′ ℓ
...
+  { Obj = Profunctor C D
+  ; _⇒_ = λ P Q → NaturalTransformation P Q
```
JacquesCarette, line 98: does eta contracting this work?

src/Categories/Functor/Profunctor.agda
```diff
@@ -89,19 +91,46 @@ module Profunctor {o ℓ e} {o′} (C : Category o ℓ e) (D : Category o′ ℓ
...
+  ; _∘_ = _∘ᵥ_
+  ; assoc     = λ { {f = f} {g} {h} → assoc-lemma {f = f} {g} {h}}
+  ; sym-assoc = λ { {f = f} {g} {h} → sym-assoc-lemma {f = f} {g} {h}}
+  ; identityˡ = λ { {f = f} → id-lemmaˡ {f = f}}
+  ; identityʳ = λ { {f = f} → id-lemmaʳ {f = f} }
+  ; identity² = λ z → z
+  ; equiv = ≃-isEquivalence
+  ; ∘-resp-≈ = λ { {f = f} {h} {g} {i} eq eq' → ∘ᵥ-resp-≃ {f = f} {h} {g} {i} eq eq' }
+  }
+  where
```
JacquesCarette, line 110: Can't these 1-liners not be inlined in the above?  As `D` is visible, you could
```
  where
    module D = Category Y
```
and then everything is quite simple.

src/Categories/Morphism/Reasoning/Core.agda
```diff
@@ -48,8 +48,8 @@ module Utils where
   assoc²' : (i ∘ (h ∘ g)) ∘ f ≈ i ∘ (h ∘ (g ∘ f))
   assoc²' = Equiv.trans assoc (∘-resp-≈ʳ assoc)
 
-  assoc²'' : (i ∘ h) ∘ g ∘ f ≈ i ∘ ((h ∘ g) ∘ f)
-  assoc²'' = Equiv.trans assoc (∘-resp-≈ʳ  (Equiv.sym assoc))
+  assoc²'' : i ∘ ((h ∘ g) ∘ f) ≈ (i ∘ h) ∘ (g ∘ f)
+  assoc²'' = Equiv.trans (∘-resp-≈ʳ assoc) (Equiv.sym assoc)
```
JacquesCarette: May as well use `sym-assoc` instead of `Equiv.sym assoc` ?

JacquesCarette (changes requested):
Nice. Just a few minor tweaks, and this is ready.

src/Categories/Functor/Profunctor.agda
iwilare: We thought about inlining these lemmas, but we can't seem to get it to work without generalizing those vertical composition properties over any two categories. The implicit arguments on the 1-liners should be:

```agda
  ; assoc     = λ { {f = f} {g} {h} → assoc-lemma     {C = Product (Category.op D) C} {D = Setoids ℓ e} {f = f} {g} {h}}
  ; sym-assoc = λ { {f = f} {g} {h} → sym-assoc-lemma {C = Product (Category.op D) C} {D = Setoids ℓ e} {f = f} {g} {h}}
  ; identityˡ = λ { {f = f} → id-lemmaˡ {C = Product (Category.op D) C} {D = Setoids ℓ e} {f = f} }
  ; identityʳ = λ { {f = f} → id-lemmaʳ {C = Product (Category.op D) C} {D = Setoids ℓ e} {f = f} }
```

but unfortunately directly substituting in the lemmas does not work. If you manage to find some other way to remove them let us know!

tetrapharmakon: How did this end?

JacquesCarette (approved):

src/Categories/Functor/Profunctor.agda
JacquesCarette: It is 'ending' in me pulling it in as is (once the build passes). It wasn't clear to me that assoc (etc) were being called 'at' such different categories. So the current setup is just as nice as the one with the implicits spelled out.

PR #333, opened 2022-01-02
Define Lifting Properties

TOTBWF:
## Patch Description

This PR defines [Lifting Properties](http://nlab-pages.s3.us-east-2.amazonaws.com/nlab/show/lift), [Injective and Projective Morphisms](nlab-pages.s3.us-east-2.amazonaws.com/nlab/show/injective+or+projective+morphism), and also proves a pile of lemmas about them.

## Notes
I often find myself looking for the following definition, so I've added it:
```agda
record IsIso (from : A ⇒ B) : Set (ℓ ⊔ e) where
  field
    inv : B ⇒ A
    iso : Iso from inv 

  open Iso iso public
  ```
  
This does _somewhat_ violate the naming scheme we use in the library, but it lines up with the literature so I think it's a reasonable name.

I've also been using `Retract Arrow` for retracts of morphisms, which feels a bit klunky? Perhaps it might make sense to define that notion on it's own.

### DIFF
--- a/src/Categories/Category/Construction/Arrow.agda
+++ b/src/Categories/Category/Construction/Arrow.agda
@@ -22,6 +22,7 @@ private
     A B D E : Obj
 
 record Morphism : Set (o ⊔ ℓ) where
+  constructor mor
   field
     {dom} : Obj
     {cod} : Obj
--- a/src/Categories/Diagram/Pushout.agda
+++ b/src/Categories/Diagram/Pushout.agda
@@ -6,12 +6,17 @@ module Categories.Diagram.Pushout {o ℓ e} (C : Category o ℓ e) where
 
 open Category C
 open HomReasoning
+open Equiv
+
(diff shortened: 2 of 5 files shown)


JacquesCarette (approved):
I'm fine with things as they are. I do prefer to 'line up' my equational proofs differently (stuff on the left, justification on the right, same line), but it's not a deal breaker.

Note that in general it's a good idea to give names to equational proofs (even if it's a local 'where' proof), as that makes the resulting records smaller. 

I'll leave this up for a bit for others to comment on, if they so wish.

JacquesCarette:
Also: I'm fine with `IsIso`. I don't have a better suggestion for `Retract Arrow`.

JacquesCarette:
Sorry to take so long, I've been swamped.

TOTBWF:
No worries, so have I!

PR #336, opened 2022-02-13
Prove that Cats is complete

Taneb:
I'm very close but I can't quite get this proof across the line. Help would be very welcome.

### DIFF
--- a/src/Categories/Category/Instance/Properties/Cats/Complete.agda
+++ b/src/Categories/Category/Instance/Properties/Cats/Complete.agda
@@ -1,0 +1,182 @@
+{-# OPTIONS --safe --without-K #-}
+
+module Categories.Category.Instance.Properties.Cats.Complete where
+
+open import Categories.Category
+open import Categories.Category.Complete
+open import Categories.Category.Instance.Cats
+open import Categories.Diagram.Cone
+open import Categories.Functor
+open import Categories.Morphism
+open import Categories.Morphism.Reasoning
+open import Categories.NaturalTransformation.NaturalIsomorphism
+
+open import Data.Product
+open import Level
+
+private module _ {o ℓ e} {J : Category o ℓ e} (F : Functor J (Cats (o ⊔ ℓ ⊔ e) (o ⊔ ℓ ⊔ e) (o ⊔ e))) where
+  open Functor F
+  module FJ j = Category (F₀ j)
+
+  -- The objects of the limit of F.
+  record NObj : Set (o ⊔ ℓ ⊔ e) where
+    field
+      -- For each object j of J, we have an object of F j.
+      ψ : ∀ j → FJ.Obj j
+      -- For each morphism f : j ⇒ k in J, we have the isomorphism (in F k):
+      --   F f (ψ j) ≅ ψ k.
+      commute : ∀ {j k} (f : J [ j , k ]) → _≅_ (F₀ k) (Functor.F₀ (F₁ f) (ψ j)) (ψ k)
+    module commute {j k} (f : J [ j , k ]) = _≅_ (commute f)
+
+  -- The morphisms of the limit of F.
+  record N⇒ (A B : NObj) : Set (o ⊔ ℓ ⊔ e) where
+    module A = NObj A
+    module B = NObj B
+    field
+      -- For each object j in J, we have a morphism in F j from A.ψ j to B.ψ j.
+      arr : ∀ j → F₀ j [ A.ψ j , B.ψ j ]
+      -- For each morphism f : j ⇒ k in J, we have the equality in F k:
+      --   B.commute f ∘ F f (arr j) ≈ arr k ∘ A.commute f.
+      commute : ∀ {j k} (f : J [ j , k ]) → F₀ k [ F₀ k [ B.commute.from f ∘ Functor.F₁ (F₁ f) (arr j) ] ≈ F₀ k [ arr k ∘ A.commute.from f ] ]
+
+  -- The actual category that is the limit of F.
+  N : Category (o ⊔ ℓ ⊔ e) (o ⊔ ℓ ⊔ e) (o ⊔ e)
+  N = record
+    { Obj = NObj
+    ; _⇒_ = N⇒
+    ; _≈_ = λ f g → ∀ j → F₀ j [ N⇒.arr f j ≈ N⇒.arr g j ]
+    ; id = λ {A} → let module A = NObj A in record
+      { arr = λ j → FJ.id j
+      ; commute = λ {j} {k} g → let open FJ.HomReasoning k in begin
+        F₀ k [ A.commute.from g ∘ Functor.F₁ (F₁ g) (FJ.id j) ] ≈⟨ refl⟩∘⟨ Functor.identity (F₁ g) ⟩
+        F₀ k [ A.commute.from g ∘ FJ.id k ]                     ≈⟨ id-comm (F₀ k) ⟩
+        F₀ k [ FJ.id k ∘ A.commute.from g ]                     ∎
+      }
+    ; _∘_ = λ {A} {B} {C} f g →
+      let
+        module A = NObj A
+        module B = NObj B
+        module C = NObj C
+      in record
+        { arr = λ j → F₀ j [ N⇒.arr f j ∘ N⇒.arr g j ]
+        ; commute = λ {j} {k} h → let open FJ.HomReasoning k in begin
+          F₀ k [ C.commute.from h ∘ Functor.F₁ (F₁ h) (F₀ j [ N⇒.arr f j ∘ N⇒.arr g j ]) ]                     ≈⟨ pushʳ (F₀ k) (Functor.homomorphism (F₁ h)) ⟩
+          F₀ k [ F₀ k [ C.commute.from h ∘ Functor.F₁ (F₁ h) (N⇒.arr f j) ] ∘ Functor.F₁ (F₁ h) (N⇒.arr g j) ] ≈⟨ pushˡ (F₀ k) (N⇒.commute f h) ⟩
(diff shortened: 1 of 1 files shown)


src/Categories/Category/Instance/Properties/Cats/Complete.agda
```diff
@@ -0,0 +1,113 @@
...
+
+-- TODO: figure out what I really want to be doing with levels
+Cats-Complete : (o ℓ e o′ ℓ′ e′ : Level) → Complete o ℓ e (Cats (o ⊔ ℓ) (o ⊔ ℓ) (o ⊔ ℓ))
+Cats-Complete o ℓ e o′ ℓ′ e′ {J} F = record
+  { terminal = record
+    { ⊤ = record
+      { N = record
+        -- this is a very simplified version of "NaturalTransformation (const One) F"
+        { Obj = Σ[ S ∈ (∀ j → Category.Obj (Functor.F₀ F j)) ] (∀ {j k} (f : J [ j , k ]) → _≅_ (Functor.F₀ F k) (Functor.F₀ (Functor.F₁ F f) (S j)) (S k))
+        ; _⇒_ = λ {(S₁ , _) (S₂ , _) → ∀ j → Functor.F₀ F j [ S₁ j , S₂ j ] }
```
ohad, line 27: I'm a bit confused by this definition. Shouldn't we be requiring the compatibility condition for morphisms as well as for objects?

Taneb: That might be what I'm missing

JacquesCarette: I've found that trying to simplify things often goes sideways, and that going with the full condition (in this case `NaturalTransformation (const One) F`) tends to end up being clearest in the end.

JacquesCarette (commented):
Also: for the final version, please use some `where` clauses (or private definitions) instead of putting everything inline, for the nested records. This also ends up improving efficiency.

HuStmpHrrr:
we don't have this already? how surprising!

JacquesCarette:
I have a (surprisingly long) list of simple category theory that's still not in the library. Planning to get back to it this summer.

src/Categories/Category/Instance/Properties/Cats/Complete.agda
Taneb: > I'm a bit confused by this definition. Shouldn't we be requiring the compatibility condition for morphisms as well as for objects?

Looking at this again, I've figured out what the condition on morphisms wants to be. I'm rewriting it as private definitions rather than nested records and products, because it was getting a bit hairy to work on.

Taneb:
I had another go trying to complete this proof, and I got stuck at about the same point. If anyone has any ideas what I'm missing, please do share.

HuStmpHrrr:
I think there is an easier way than directly work all the details out, as sometimes details could be too overwhelming.

notice that completeness is just the same as having all products (not just finite) and equalizers. This is proved here by myself:

https://github.com/agda/agda-categories/blob/master/src/Categories/Category/Complete/Properties/Construction.agda

now, what's left is just to show Cats have all products and equalizers. Between two, showing all products doesn't sound difficult. I am not able to say much about equalizers just by running it in my head, but I suppose it shouldn't be too horrible if  all we need to show is up to isomorphisms.

nmvdw:
> I think there is an easier way than directly work all the details out, as sometimes details could be too overwhelming.
> 
> notice that completeness is just the same as having all products (not just finite) and equalizers. This is proved here by myself:
> 
> https://github.com/agda/agda-categories/blob/master/src/Categories/Category/Complete/Properties/Construction.agda
> 
> now, what's left is just to show Cats have all products and equalizers. Between two, showing all products doesn't sound difficult. I am not able to say much about equalizers just by running it in my head, but I suppose it shouldn't be too horrible if all we need to show is up to isomorphisms.

Since the equalizer diagram is required to commute up to the setoid equality and since the setoid equality in `Cats` is natural isomorphism, I think that we can construct the equalizer of `F, G : C -> D` as follows:
- Objects: pairs `x : C` with an iso `F x -> G x`
- Morphisms from `h_x : F x -> G x` to `h_y : F y -> G y`: `f : x -> y` such that `h_y o F f = G f o h_x`

(if we would look at this from a 2-categorical perspective, this would be the isoinserter (https://ncatlab.org/nlab/show/isoinserter)).

HuStmpHrrr:
> > I think there is an easier way than directly work all the details out, as sometimes details could be too overwhelming.
> > notice that completeness is just the same as having all products (not just finite) and equalizers. This is proved here by myself:
> > https://github.com/agda/agda-categories/blob/master/src/Categories/Category/Complete/Properties/Construction.agda
> > now, what's left is just to show Cats have all products and equalizers. Between two, showing all products doesn't sound difficult. I am not able to say much about equalizers just by running it in my head, but I suppose it shouldn't be too horrible if all we need to show is up to isomorphisms.
> 
> Since the equalizer diagram is required to commute up to the setoid equality and since the setoid equality in `Cats` is natural isomorphism, I think that we can construct the equalizer of `F, G : C -> D` as follows:
> 
>     * Objects: pairs `x : C` with an iso `F x -> G x`
> 
>     * Morphisms from `h_x : F x -> G x` to `h_y : F y -> G y`: `f : x -> y` such that `h_y o F f = G f o h_x`
> 
> 
> (if we would look at this from a 2-categorical perspective, this would be the isoinserter (https://ncatlab.org/nlab/show/isoinserter)).

I think it might be worth trying out the equalizer first, as the morphism part seems a bit brittle.

Taneb:
@nmvdw 

> Since the equalizer diagram is required to commute up to the setoid equality and since the setoid equality in `Cats` is natural isomorphism, I think that we can construct the equalizer of `F, G : C -> D` as follows:
> 
>     * Objects: pairs `x : C` with an iso `F x -> G x`
> 
>     * Morphisms from `h_x : F x -> G x` to `h_y : F y -> G y`: `f : x -> y` such that `h_y o F f = G f o h_x`
> 
> 
> (if we would look at this from a 2-categorical perspective, this would be the isoinserter (https://ncatlab.org/nlab/show/isoinserter)).

I tried to define the equalizer with these definitions, and... I got stuck in the same place, proving that the arrows commute in the natural isomorphism. You can see my attempt [here](https://github.com/Taneb/agda-categories/blob/cat-complete-by-equalizer/src/Categories/Category/Instance/Properties/Cats/Complete.agda)

I'm pretty sure that this definition of equalizer leads to something more or less equivalent with my explicit definition of completeness, so it makes sense that I hit the same problem

HuStmpHrrr:
I haven't run it myself, but I have a feeling that this kind of "set comprehension"-ish operations tend to fail because we are running setoids. it might imply that the comprehension should be a bit more complex to take into account the proof relevant details, or we have missing predicates in our definitions. I hope it is not the latter.

nmvdw:
> @nmvdw
> 
> > Since the equalizer diagram is required to commute up to the setoid equality and since the setoid equality in `Cats` is natural isomorphism, I think that we can construct the equalizer of `F, G : C -> D` as follows:
> > ```
> > * Objects: pairs `x : C` with an iso `F x -> G x`
> > 
> > * Morphisms from `h_x : F x -> G x` to `h_y : F y -> G y`: `f : x -> y` such that `h_y o F f = G f o h_x`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (if we would look at this from a 2-categorical perspective, this would be the isoinserter (https://ncatlab.org/nlab/show/isoinserter)).
> 
> I tried to define the equalizer with these definitions, and... I got stuck in the same place, proving that the arrows commute in the natural isomorphism. You can see my attempt [here](https://github.com/Taneb/agda-categories/blob/cat-complete-by-equalizer/src/Categories/Category/Instance/Properties/Cats/Complete.agda)
> 
> I'm pretty sure that this definition of equalizer leads to something more or less equivalent with my explicit definition of completeness, so it makes sense that I hit the same problem

There are two ways to understand the statement that `Cat` is complete:
1. The bicategory `Cat` is complete. 'Equalizers' (more precisely, iso-inserters) are constructed as follows:
```
Equalizer of F, G : C1 -> C2
- Objects: x : C1 with iso F x -> G x
- Morphisms from f : F x -> G x to g : F y -> G y: pairs h : x -> y such that F x -> F y -> G y and F x -> G x -> G y are equal
```
Let's call this one `IsoInserter F G`.
Note that this construction indeed is similar to how you proved completeness, since every object came with an isomorphism.

2. The category `StrictCat` is complete. Equalizers are constructed as follows:
```
Equalizer of F, G : C1 -> C2
- Objects: x : C1 such that F x = G x
- Morphisms: f : x1 -> x2 such that F f = G f
```
Let's call this one `Equalizer F G`. Note that the definition of `Equalizer F G` speaks about equality of objects.

In mathematical books on category theory, one uses set-theoretic foundations. In that case, the statement `Cat` is a complete category usually refers to the second of these two variants. 

However, the category of categories in Agda-categories is a bit different than the aforementioned `StrictCat`. In `StrictCat`, one considers functors up to equality, while Agda-categories considers functors up to natural isomorphism. For that reason, `Equalizer F G` is too strict to be the equalizer.

In addition, `IsoInserter F G` satisfies a 2-categorical universal property rather than a 1-categorical universal property. For that property, uniqueness follows under stronger assumptions: the additional assumption being precisely the equality that you are missing in the construction.

For that reason, neither construction gives equalizer in `Cats`, and I think the current definition of `Cat` actually might not give rise a complete category. There are two possible fixes:
1. Study at the bicategory of categories instead. In that case, `IsoInserter F G` would work well.
2. Change the setoid relation on morphisms in `Cats` (functors up to equality), and then one could use `Equalizer F G`.

PR #337, opened 2022-03-07
Implementation of Lax Slice bicategory

maxsnew:
The lax slice bicategory is like the slice category, but rather than a commuting triangle you get a 2-cell.

This is my first time using Agda in earnest so I am not very adept at using the module system, so some code review would help here for sure. I'm not really sure if I got the naming conventions right. I based most of the naming on the slice 1-category.

Some caveats
1. I still have two spots of yellow in Emacs when I type check on Line 310. I'm not sure how to debug this.
2. Type checking at least while it was incomplete became *very* slow. When I was working on the unitors I had to comment out the associator proofs to speed things up. Not sure what the best practice is for mitigating this. Should I break it up into multiple files? Maybe move the SliceHom module into a different file.
3. The proofs sometimes had several lines in a row of associativity, is there a simpler way to do this? A reflective free category solver maybe?
4. One proof (inverse of the right unitor) is a copy-paste-rename of the proof for the inverse of the left unitor. This should probably be a lemma instead.

### DIFF
--- a/src/Categories/Bicategory/Construction/LaxSlice.agda
+++ b/src/Categories/Bicategory/Construction/LaxSlice.agda
@@ -53,13 +53,11 @@ module SliceHom (A : Obj) where
       {ϕ} : J.h ⇒₂ K.h
       E   : K.Δ ≈ (Y.arr ▷ ϕ ∘ᵥ J.Δ)
 
-
   _∘ᵥ/_ : ∀ {X Y : SliceObj A}{J K L : Slice⇒₁ X Y} → Slice⇒₂ K L → Slice⇒₂ J K → Slice⇒₂ J L
   _∘ᵥ/_ {X}{Y}{J}{K}{L} (slicearr₂ {ϕ = ϕ} E) (slicearr₂ {ϕ = ψ} F) = slicearr₂ {ϕ = ϕ ∘ᵥ ψ} $ begin
       L.Δ                             ≈⟨ E ⟩
       (Y.arr ▷ ϕ ∘ᵥ K.Δ)              ≈⟨ refl⟩∘⟨ F ⟩
-      Y.arr ▷ ϕ ∘ᵥ (Y.arr ▷ ψ ∘ᵥ J.Δ) ≈˘⟨ assoc ⟩
-      (Y.arr ▷ ϕ ∘ᵥ Y.arr ▷ ψ) ∘ᵥ J.Δ ≈⟨ (∘ᵥ-distr-▷ ⟩∘⟨refl) ⟩
+      Y.arr ▷ ϕ ∘ᵥ (Y.arr ▷ ψ ∘ᵥ J.Δ) ≈⟨ pullˡ ∘ᵥ-distr-▷ ⟩
       Y.arr ▷ (ϕ ∘ᵥ ψ) ∘ᵥ J.Δ         ∎
     where module X = SliceObj X
           module Y = SliceObj Y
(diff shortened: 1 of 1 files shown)


src/Categories/Bicategory/Construction/LaxSlice.agda
```diff
@@ -0,0 +1,337 @@
...
+-- Mentioned in passing https://ncatlab.org/nlab/show/slice+2-category
+
+open import Categories.Bicategory
+
+module Categories.Bicategory.Construction.LaxSlice
+       {o ℓ e t}
+       (𝒞 : Bicategory o ℓ e t)
+       where
+
+open import Categories.Enriched.Category
```
JacquesCarette, line 12: The more modern parts of agda-categories tend to be quite specific on all imports (unless there's tons and tons from one file).

src/Categories/Bicategory/Construction/LaxSlice.agda
```diff
@@ -0,0 +1,337 @@
...
+        )
+    where module X = SliceObj X
+          module Y = SliceObj Y
+          module H = Slice⇒₁ H
+          module K = Slice⇒₁ K
+          module L = Slice⇒₁ L
+          open 1Category (hom X.Y A)
+          open HomReasoning
+          open Equiv
+          open import Categories.Morphism.Reasoning (hom X.Y A)
```
JacquesCarette, line 69: All imports should go at the top of the file.

Often it's best to
```agda
import Categories.Morphism.Reasoning as MR
```
and then do `open MR (hom X.Y A)` at the point of use.

src/Categories/Bicategory/Construction/LaxSlice.agda
```diff
@@ -0,0 +1,337 @@
...
+          open import Relation.Binary.Core using (Rel)
+          open import Function.Base using (_$_)
+          
+  open SliceObj
+  SliceHomCat : SliceObj A → SliceObj A → 1Category (o ⊔ ℓ) (ℓ ⊔ e) e
+  SliceHomCat X Y = record
+    { Obj = Slice⇒₁ X Y
+    ; _⇒_ = Slice⇒₂
+    ; _≈_ = λ where
+      (slicearr₂ {ϕ} _) (slicearr₂ {ψ} _) → ϕ ≈ ψ
```
JacquesCarette, line 79: As you end up using the implicits in slicearr all the time, I think that says that they should actually be explicit.

src/Categories/Bicategory/Construction/LaxSlice.agda
```diff
@@ -0,0 +1,337 @@
...
+    private
+      module Y = SliceObj Y
+      module H = Slice⇒₁ H
+      module K = Slice⇒₁ K
+
+    field
+      {ϕ} : H.h ⇒₂ K.h
+      E   : K.Δ ≈ (Y.arr ▷ ϕ ∘ᵥ H.Δ)
+
+  open hom.Equiv
```
JacquesCarette, line 49: One reason for explicit `using` is things like this: took me forever to figure out (in the github editor, not emacs) where `hom` came from.

src/Categories/Bicategory/Construction/LaxSlice.agda
```diff
@@ -0,0 +1,337 @@
...
+
+LaxSlice : Obj → Bicategory (o ⊔ ℓ) (ℓ ⊔ e) e (o ⊔ t)
+LaxSlice A   = record
+  { enriched = record
+    { Obj = SliceObj A
+    ; hom = SliceHomCat
+    ; id = const (SliceHom.slicearr₁ ρ⇐)
+    ; ⊚ = _⊚'_
+    ; ⊚-assoc = λ {W X Y Z} →
+      let module W = SliceObj W
```
JacquesCarette, line 113: `let` is very badly behaved in agda (it gets inlined), so that tends to make things slow. `where` is preferred. So you should put all of this in a `where` for the record. Or define it as a private function above `LaxSlice`.

JacquesCarette (changes requested):
About assoc: there are a bunch of push/pull/center/etc combinators that help a lot with that. They are used all over the library. They embody common reasoning patterns (that you are indeed using here) with assoc that can be hidden away.

Also, please try to line up your = 's in equational reasoning.

src/Categories/Bicategory/Construction/LaxSlice.agda
maxsnew: Seems fair. I based this on the analogous constructor in the slice 1-category library. It works well when using the constructor because the morphism can be inferred from the equality proof.

JacquesCarette: Note that this isn't a "for sure" thing. You need to differentiate between needing to refer to it explicitly in fundamental definitions versus in applications. If it's seldom needed in applications, then leaving it implicit might be fine.

maxsnew:
Ok I think it's much better now. I did the following:
1. Moved all the imports to the top
2. Used explicit using for the top-level opens
3. Lifted a lot of local functions to definitions and got rid of all the lets instead using wheres
4. Used push/pull/center etc for the associativity arguments, which made the proofs much shorter.
5. Made the naming of arguments consistent.

src/Categories/Bicategory/Construction/LaxSlice.agda
```diff
@@ -0,0 +1,245 @@
...
+    ; _⇒_ = Slice⇒₂
+    ; _≈_ = λ (slicearr₂ {ϕ} _) (slicearr₂ {ψ} _) → ϕ ≈ ψ 
+    ; id = slice-id _
+    ; _∘_ = _∘ᵥ/_
+    ; assoc = hom.assoc
+    ; sym-assoc = hom.sym-assoc
+    ; identityˡ = hom.identityˡ
+    ; identityʳ = hom.identityʳ
+    ; identity² = hom.identity²
+    ; equiv = record
```
JacquesCarette, line 84: When it's that short, I tend to put such `equiv` records all on one line. Your choice.

JacquesCarette (changes requested):
This is looking fantastic! Mostly a tiny bookkeeping thing about the imports that would be good to twiddle.

src/Categories/Bicategory/Construction/LaxSlice.agda
```diff
@@ -0,0 +1,245 @@
...
+
+open import Categories.Category using () renaming (Category to 1Category)
+import Categories.Morphism.Reasoning as MR
+open import Categories.Bicategory.Extras 𝒞
+open Shorthands
+
+open import Categories.Functor.Construction.Constant using (const)
+open import Categories.Functor.Bifunctor using (Bifunctor)
+open import Categories.NaturalTransformation.NaturalIsomorphism using (NaturalIsomorphism; niHelper)
+open import Function using (_$_)
```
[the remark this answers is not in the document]
maxsnew: Done

JacquesCarette (approved):

maxsnew:
Sorry for all the commits, I intended to squash them when the review was done

JacquesCarette:
Actually, I kind of like fine-grained commits.

PR #339, opened 2022-03-11
Add bundled CCCs and basic def. for CC functors.

Trebor-Huang:
This is barely nothing apart from definitions. I'm new to contributing so I'm still looking at proving some basic properties about CCFs.

### DIFF
--- a/src/Categories/Functor/CartesianClosed.agda
+++ b/src/Categories/Functor/CartesianClosed.agda
@@ -1,0 +1,37 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Functor.CartesianClosed where
+
+open import Level
+
+open import Categories.Category.CartesianClosed.Bundle using (CartesianClosedCategory)
+open import Categories.Category.CartesianClosed
+open import Categories.Functor using (Functor; _∘F_)
+open import Categories.Functor.Cartesian
+
+import Categories.Morphism as M
+
+private
+  variable
+    o ℓ e o′ ℓ′ e′ : Level
+
+record IsCartesianClosedF (C : CartesianClosedCategory o ℓ e) (D : CartesianClosedCategory o′ ℓ′ e′)
+  (F : Functor (CartesianClosedCategory.U C) (CartesianClosedCategory.U D)) : Set (o ⊔ ℓ ⊔ e ⊔ o′ ⊔ ℓ′ ⊔ e′) where
+  private
+    module C = CartesianClosedCategory C
+    module D = CartesianClosedCategory D
+    open Functor F
+    module CC = CartesianClosed C.cartesianClosed
+    module DC = CartesianClosed D.cartesianClosed
+  field
+    F-cartesian : IsCartesianF C.cartesianCategory D.cartesianCategory F
+  private
+    open IsCartesianF F-cartesian
+    F-mor : ∀ A B
+      → F₀ (A CC.^ B) D.⇒ F₀ A DC.^ F₀ B
+    F-mor A B = DC.λg (F₁ CC.eval′ D.∘ ×-iso.to (A CC.^ B) B)
+  field
+    F-closed : ∀ {A B}
+      → M.IsIso D.U (F-mor A B)
+
+
--- a/src/Categories/Functor/CartesianClosed.agda
+++ b/src/Categories/Functor/CartesianClosed.agda
@@ -15,7 +15,7 @@ private
   variable
     o ℓ e o′ ℓ′ e′ : Level
 
-record IsCartesianClosedF {C : CartesianClosedCategory o ℓ e} {D : CartesianClosedCategory o′ ℓ′ e′}
+record IsCartesianClosedF (C : CartesianClosedCategory o ℓ e) (D : CartesianClosedCategory o′ ℓ′ e′)
   (F : Functor (CartesianClosedCategory.U C) (CartesianClosedCategory.U D)) : Set (o ⊔ ℓ ⊔ e ⊔ o′ ⊔ ℓ′ ⊔ e′) where
   private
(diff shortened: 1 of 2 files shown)


JacquesCarette:
Small incremental contributions are quite welcome too. Now on to doing a review.

src/Categories/Functor/CartesianClosed.agda
```diff
@@ -0,0 +1,37 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Functor.CartesianClosed where
+
+open import Level
+
+open import Categories.Category.CartesianClosed.Bundle using (CartesianClosedCategory)
+open import Categories.Category.CartesianClosed
```
JacquesCarette, line 8: can you add `using` both here and for `Categories.Functor.Cartesian` below?

src/Categories/Functor/CartesianClosed.agda
```diff
@@ -0,0 +1,37 @@
...
+open import Categories.Functor using (Functor; _∘F_)
+open import Categories.Functor.Cartesian
+
+import Categories.Morphism as M
+
+private
+  variable
+    o ℓ e o′ ℓ′ e′ : Level
+
+record IsCartesianClosedF (C : CartesianClosedCategory o ℓ e) (D : CartesianClosedCategory o′ ℓ′ e′)
```
JacquesCarette, line 18: `C` and `D` can probably be implicits.

src/Categories/Functor/CartesianClosed.agda
```diff
@@ -0,0 +1,37 @@
...
+  (F : Functor (CartesianClosedCategory.U C) (CartesianClosedCategory.U D)) : Set (o ⊔ ℓ ⊔ e ⊔ o′ ⊔ ℓ′ ⊔ e′) where
+  private
+    module C = CartesianClosedCategory C
```
JacquesCarette, line 21: Because of how Agda works, this causes a huge amount of bloat. Best to use `using` here too, so that the resulting module have a small interface.

src/Categories/Functor/CartesianClosed.agda
```diff
@@ -0,0 +1,37 @@
...
+    module D = CartesianClosedCategory D
+    open Functor F
+    module CC = CartesianClosed C.cartesianClosed
+    module DC = CartesianClosed D.cartesianClosed
+  field
+    F-cartesian : IsCartesianF C.cartesianCategory D.cartesianCategory F
+  private
```
JacquesCarette, line 28: Why make `F-mor` private? I wouldn't be surprised if many applications ended up needing it.

src/Categories/Functor/CartesianClosed.agda
```diff
@@ -0,0 +1,37 @@
...
+    open IsCartesianF F-cartesian
+    F-mor : ∀ A B
+      → F₀ (A CC.^ B) D.⇒ F₀ A DC.^ F₀ B
+    F-mor A B = DC.λg (F₁ CC.eval′ D.∘ ×-iso.to (A CC.^ B) B)
+  field
+    F-closed : ∀ {A B}
```
JacquesCarette, line 34: This should be all on one line.

src/Categories/Functor/CartesianClosed.agda
```diff
@@ -0,0 +1,37 @@
...
+    module C = CartesianClosedCategory C
+    module D = CartesianClosedCategory D
+    open Functor F
+    module CC = CartesianClosed C.cartesianClosed
+    module DC = CartesianClosed D.cartesianClosed
+  field
+    F-cartesian : IsCartesianF C.cartesianCategory D.cartesianCategory F
+  private
+    open IsCartesianF F-cartesian
+    F-mor : ∀ A B
```
JacquesCarette, line 30: This signature should be on one line too.

JacquesCarette (changes requested):

src/Categories/Functor/CartesianClosed.agda
Trebor-Huang: It turns out that, since the type of `F` involves `U C` instead of just `C`, Agda can't infer `C` and `D`.

Trebor-Huang:
I'm sorry but I recently need to focus more on intracurriculum stuff, so I probably won't have time to do much. :(

src/Categories/Functor/CartesianClosed.agda
Trebor-Huang: By the way, what should it be named? Some sort of canonical morphism?

src/Categories/Functor/CartesianClosed.agda
JacquesCarette: Ok!

JacquesCarette:
No problem - any and all contributions are appreciated. This is almost ready to be merged!

src/Categories/Functor/CartesianClosed.agda
sstucki: +1 for asking on Zulip, but I don't think these sort of canonical morphisms have standard names. In nlab [the corresponding morphism for closed functors](https://ncatlab.org/nlab/show/closed+functor#definition) is just denoted $\hat{F}$. And it seems similar with the maps in e.g. monoidal functors, which are just given some Greek names (not following any particular convention). In the library these are called `ε` and `⊗-homo`. So by analogy this would be `^-homo`. Maybe we could come up with our own conventions? E.g. if a morphism witnesses the (lax) preservation of X (here X = exponentials) it could be called `preserves-X`? (Not particularly inventive, I know).

Trebor-Huang: Maybe we can open a separate PR to properly name all these.

JacquesCarette: Sure, that can be done as a second step.

JacquesCarette (approved):

PR #343, opened 2022-03-21
bicategory renaming

### LINKED ISSUE #338: Bicategory terminology
Categories.Bicategory.agda defines the following as "horizontal composition"
```agda
  -- horizontal composition
  _∘ₕ_ : {A B C : Obj} → B ⇒₁ C → A ⇒₁ B → A ⇒₁ C
  _∘ₕ_ = _⊚₀_
```
This terminology is a bit strange to me. This is the composition of the 1-cells, whereas I've always heard "horizontal composition" to refer to the "parallel" composition of 2-cells, as [here](https://ncatlab.org/nlab/show/horizontal+composition), which is in the library of course:
```agda
  _⊚₁_ : {A B C : Obj} {f h : B ⇒₁ C} {g i : A ⇒₁ B} → f ⇒₂ h → g ⇒₂ i → f ⊚₀ g ⇒₂ h ⊚₀ i
  α ⊚₁ β = Functor.F₁ ⊚ (α , β)
```

I think the traditional terminology makes sense because vertical/horizontal are both constructions on 2-cells. I suggest we rename `_∘ₕ_` to be an alias for `_⊚₁_`, and maybe we can give `_∘₁_` as the alias for 1-cell composition instead.

maxsnew:
Fixes #338

### DIFF
--- a/src/Categories/Bicategory.agda
+++ b/src/Categories/Bicategory.agda
@@ -23,10 +23,10 @@ record Bicategory o ℓ e t : Set (suc (o ⊔ ℓ ⊔ e ⊔ t)) where
   module ComHom {A B} = Commutation (hom A B)
 
   infix 4 _⇒₁_ _⇒₂_ _≈_
-  infixr 7 _∘ᵥ_ _∘ₕ_
+  infixr 7 _∘ᵥ_ _∘₁_
   infixr 9 _▷_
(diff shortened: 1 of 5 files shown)


JacquesCarette (approved):
Excellent, thank you.

PR #346, opened 2022-04-15
[draft] Add draft definition of a double category

Boarders:
Opening this pull request to get style advice / denunciations for aesthetic crimes.

Still to do:
- Add interchange law(s)
- Defining the underlying vertical and horizontal 2-categories

### DIFF
--- a/src/Categories/Double/Core.agda
+++ b/src/Categories/Double/Core.agda
@@ -1,0 +1,274 @@
+{-# OPTIONS --without-K --safe #-}
+module Categories.Double.Core where
+
+open import Level
+
+open import Relation.Binary using (Rel; IsEquivalence; Setoid; REL)
+open import Relation.Binary.PropositionalEquality as ≡ using (_≡_)
+import Relation.Binary.Reasoning.Setoid as SetoidR
+
+import Categories.Category.Unbundled as Cat
+
+-- Equality of squares in a double category
+module _ {o ℓ e : Level} {Obj : Set o} (Hor Ver : Cat.Category Obj ℓ e) where
+  private
+    module H = Cat.Category Hor
+    module V = Cat.Category Ver
+    _≈ₕ_ : ∀ {A B} → Rel (A H.⇒ B) e
+    _≈ₕ_ = H._≈_
+    _≈ᵥ_ : ∀ {A B} → Rel (A V.⇒ B) e
+    _≈ᵥ_ = V._≈_
+  record SqEquality
+    {T₁ T₂ B₁ B₂ : Obj}
+    (hT₁ hT₂ : T₁ H.⇒ T₂)
+    (hB₁ hB₂ : B₁ H.⇒ B₂)
+    (vL₁ vL₂ : T₁ V.⇒ B₁)
+    (vR₁ vR₂ : T₂ V.⇒ B₂) : Set e where
+      field
+        horT≈ : hT₁ ≈ₕ hT₂
+        horB≈ : hB₁ ≈ₕ hB₂
+        verL≈ : vL₁ ≈ᵥ vL₂
+        verR≈ : vR₁ ≈ᵥ vR₂
+
+module _ {o ℓ e : Level} {Obj : Set o} (Hor Ver : Cat.Category Obj ℓ e) where
+  private
+    module H = Cat.Category Hor
+    module V = Cat.Category Ver
+  dual≈ :
+    ∀ {T₁ T₂ B₁ B₂ : Obj} →
+    {hT₁ hT₂ : T₁ H.⇒ T₂} →
+    {hB₁ hB₂ : B₁ H.⇒ B₂} →
+    {vL₁ vL₂ : T₁ V.⇒ B₁} →
+    {vR₁ vR₂ : T₂ V.⇒ B₂} →
+    SqEquality Hor Ver hT₁ hT₂ hB₁ hB₂ vL₁ vL₂ vR₁ vR₂ →
+    SqEquality Ver Hor vL₁ vL₂ vR₁ vR₂ hT₁ hT₂ hB₁ hB₂
+  dual≈ S = record
+      { horT≈ = S.verL≈
+      ; horB≈ = S.verR≈
+      ; verL≈ = S.horT≈
+      ; verR≈ = S.horB≈
+      }
+    where module S = SqEquality S
+
+-- Basic definition of a |Double Category|
+record Category (o ℓ e : Level) : Set (suc (o ⊔ ℓ ⊔ e)) where
+  field
+    Obj : Set o
+    Hor : Cat.Category Obj ℓ e
+    Ver : Cat.Category Obj ℓ e
+  private
+    module H = Cat.Category Hor
+    module V = Cat.Category Ver
+  _⇒ₕ_ : Rel Obj ℓ
+  _⇒ₕ_ = H._⇒_
+  _≈ₕ_ : ∀ {A B} → Rel (A ⇒ₕ B) e
+  _≈ₕ_ = H._≈_
+  _⇒ᵥ_ : Rel Obj ℓ
+  _⇒ᵥ_ = V._⇒_
+  _≈ᵥ_ : ∀ {A B} → Rel (A ⇒ᵥ B) e
+  _≈ᵥ_ = V._≈_
+  _∘₁ₕ_ : ∀ {A B C} → (B ⇒ₕ C) → (A ⇒ₕ B) → (A ⇒ₕ C)
+  _∘₁ₕ_ = H._∘_
+  _∘₁ᵥ_ : ∀ {A B C} → (B ⇒ᵥ C) → (A ⇒ᵥ B) → (A ⇒ᵥ C)
+  _∘₁ᵥ_ = V._∘_
+
+  field
+    Sq₂ : {T₁ T₂ B₁ B₂ : Obj} → T₁ ⇒ₕ T₂ → B₁ ⇒ₕ B₂ → T₁ ⇒ᵥ B₁ → T₂ ⇒ᵥ B₂ → Set ℓ
+
+  Sq≈ : ∀ {T₁ T₂ B₁ B₂ : Obj}
+    {hT₁ hT₂ : T₁ ⇒ₕ T₂} →
+    {hB₁ hB₂ : B₁ ⇒ₕ B₂} →
+    {vL₁ vL₂ : T₁ ⇒ᵥ B₁} →
+    {vR₁ vR₂ : T₂ ⇒ᵥ B₂} →
+    REL (Sq₂ hT₁ hB₁ vL₁ vR₁) (Sq₂ hT₂ hB₂ vL₂ vR₂) e
+  Sq≈ {_} {_} {_} {_} {hT₁} {hT₂} {hB₁} {hB₂} {vL₁} {vL₂} {vR₁} {vR₂} _ _ =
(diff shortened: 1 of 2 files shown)


Boarders:
It looks like these unicode characters are variable width in different fonts so I might have to re-think the diagrams (which I do think are valuable once you have higher dimensions happening).

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,297 @@
...
+    SqEquality Ver Hor vL₁ vL₂ vR₁ vR₂ hT₁ hT₂ hB₁ hB₂
+  dual≈ S = record
+      { horT≈ = S.verL≈
+      ; horB≈ = S.verR≈
+      ; verL≈ = S.horT≈
+      ; verR≈ = S.horB≈
+      }
+    where module S = SqEquality S
+
+-- Basic definition of a |Double Category|
```
maxsnew, line 53: It would be good to document somewhere (probably the top) what level of strictness you are using. It looks like these are "maximally" strict, i.e., equivalent to strict categories internal to cat. But often people say double category to mean vertically strict and horizontally weakly associative/unital, so best to clarify that up front.

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,297 @@
...
+open import Level
+
+open import Relation.Binary using (Rel; IsEquivalence; Setoid; REL)
+open import Relation.Binary.PropositionalEquality as ≡ using (_≡_)
+import Relation.Binary.Reasoning.Setoid as SetoidR
+
+import Categories.Category.Unbundled as Cat
+
+-- Equality of squares in a double category
+module _ {o ℓ e : Level} {Obj : Set o} (Hor Ver : Cat.Category Obj ℓ e) where
```
JacquesCarette, line 13: I'm pretty sure that we can usefully have both categories live at different levels, so that this would involve 5 levels instead of 3.

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,297 @@
...
+  private
+    module H = Cat.Category Hor
+    module V = Cat.Category Ver
+    _≈ₕ_ : ∀ {A B} → Rel (A H.⇒ B) e
+    _≈ₕ_ = H._≈_
+    _≈ᵥ_ : ∀ {A B} → Rel (A V.⇒ B) e
+    _≈ᵥ_ = V._≈_
+  record SqEquality
```
JacquesCarette, line 21: Some comments here about the purpose and design of this `SqEquality` would be good to add.

src/Categories/Double/Core.agda
JacquesCarette: I think this might just be because it's WIP. Already the definition is weaker than usual because of the concept of equality on squares being used is much less strict.

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,297 @@
...
+      ; verL≈ = S.horT≈
+      ; verR≈ = S.horB≈
+      }
+    where module S = SqEquality S
+
+-- Basic definition of a |Double Category|
+record Category (o ℓ e : Level) : Set (suc (o ⊔ ℓ ⊔ e)) where
+  field
+    Obj : Set o
+    Hor : Cat.Category Obj ℓ e
```
JacquesCarette, line 57: Definitely should have 5 levels.

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,297 @@
...
+      {hB₁ : B₁ ⇒ₕ B₂} {hB₂ : B₂ ⇒ₕ B₃} →
+      {vL₁  : T₁ ⇒ᵥ M₁} {vL₂  : M₁ ⇒ᵥ B₁} →
+      {vM₁  : T₂ ⇒ᵥ M₂} {vM₂  : M₂ ⇒ᵥ B₂} →
+      {vR₁  : T₃ ⇒ᵥ M₃} {vR₂  : M₃ ⇒ᵥ B₃} →
+      (s₄ : Sq₂ hM₂ hB₂ vM₂ vR₂) →
+      (s₃ : Sq₂ hM₁ hB₁ vL₂ vM₂) →
+      (s₂ : Sq₂ hT₂ hM₂ vM₁ vR₁)
+      (s₁ : Sq₂ hT₁ hM₁ vL₁ vM₁) →
+      Sq≈ ((s₄ ∘₂ₕ s₃) ∘₂ᵥ (s₂ ∘₂ₕ s₁)) ((s₄ ∘₂ᵥ s₂) ∘₂ₕ (s₃ ∘₂ᵥ s₁))
+  -- When a category is quantified, it is convenient to refer to the levels from a module,
```
JacquesCarette, line 264: Turns out these things (that I know you took from Category's own definition) are barely used - best to leave them out.

src/Categories/Double/Core.agda
Boarders: Not completely sure what would be best to put here. This feels like a correct definition for a setoid enriched category in the sense that the source and target for 2-cells must be equipped with maps between the associated hom-setoids and so equality of 2-cells must include the data of equality between each of the 1-cells. What do you both think would be best?

src/Categories/Double/Core.agda
```diff
@@ -261,19 +290,8 @@ interchange law:
...
-  o-level = o
-
-  ℓ-level : Level
-  ℓ-level = ℓ
-
-  e-level : Level
-  e-level = e
 
-  dual : Category o ℓ e
+  dual : Category o ℓ ℓ' e e'
```
JacquesCarette, line 294: I'm surprised that this `dual` didn't require the levels be permuted?

src/Categories/Double/Core.agda
JacquesCarette: I agree that this is the most 'natural' definition of Double Category (and agrees with the one on nLab) for E-Categories. There are indeed different definitions for Pseudo Double and Double Bicategory.

JacquesCarette:
@maxsnew can you give this another look?

maxsnew:
I'm on vacation so I won't be able to look for about a week and a half

src/Categories/Double/Core.agda
Boarders: Quite right and good spot, the levels were all messed up in the definition.

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,315 @@
...
+         (Hor : Cat.Category Obj ℓ e) (Ver : Cat.Category Obj ℓ' e') where
+  private
+    module H = Cat.Category Hor
+    module V = Cat.Category Ver
+    _≈ₕ_ : ∀ {A B} → Rel (A H.⇒ B) e
+    _≈ₕ_ = H._≈_
+    _≈ᵥ_ : ∀ {A B} → Rel (A V.⇒ B) e'
+    _≈ᵥ_ = V._≈_
+  record SqEquality
+    {T₁ T₂ B₁ B₂ : Obj}
```
JacquesCarette, line 50: So now that the big things are over, we get to the most controversial part: naming!

I suggest that:
- you name the objects A B C D
- you name the arrows T, B, L and R (with subscript 1 and 2), i.e. drop the h and v prefix
- name the fields T≈, .., R≈.

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,315 @@
...
+        verL≈ : vL₁ ≈ᵥ vL₂
+        verR≈ : vR₁ ≈ᵥ vR₂
+
+module _ {o ℓ ℓ' e e' : Level} {Obj : Set o}
+         (Hor : Cat.Category Obj ℓ e) (Ver : Cat.Category Obj ℓ' e') where
+  private
+    module H = Cat.Category Hor
+    module V = Cat.Category Ver
+  dual≈ :
+    ∀ {T₁ T₂ B₁ B₂ : Obj} →
```
JacquesCarette, line 67: I would then use the same naming convention here.

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,315 @@
...
+  _⇒ᵥ_ = V._⇒_
+  _≈ᵥ_ : ∀ {A B} → Rel (A ⇒ᵥ B) e
+  _≈ᵥ_ = V._≈_
+  _∘₁ₕ_ : ∀ {A B C} → (B ⇒ₕ C) → (A ⇒ₕ B) → (A ⇒ₕ C)
+  _∘₁ₕ_ = H._∘_
+  _∘₁ᵥ_ : ∀ {A B C} → (B ⇒ᵥ C) → (A ⇒ᵥ B) → (A ⇒ᵥ C)
+  _∘₁ᵥ_ = V._∘_
+
+  field
+    Sq₂ : {T₁ T₂ B₁ B₂ : Obj} → T₁ ⇒ₕ T₂ → B₁ ⇒ₕ B₂ → T₁ ⇒ᵥ B₁ → T₂ ⇒ᵥ B₂ → Set ℓ
```
JacquesCarette: And I would continue here too, i.e. name the objects A, B, C and D.

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,315 @@
...
+
+  Sq≈ : ∀ {T₁ T₂ B₁ B₂ : Obj}
```
JacquesCarette, line 107: So this definition is odd: it reads like it's a relation between two squares, but it's not: it's a relation between two **perimeters**. The insides of the square are ignored!

This is a useful notion, but it should have a different name.

JacquesCarette: And then actual 'square equality' should be defined to say that given two squares whose perimeters are equal, we get a type of witnesses that the squares themselves are equal.

src/Categories/Double/Core.agda
```diff
@@ -0,0 +1,315 @@
...
+      ; verR≈ = S.horB≈
+      }
+    where module S = SqEquality S
+
+-- Basic definition of a strict setoid-enriched |Double Category|.
+record Category (o ℓ ℓ' e e' : Level) : Set (suc (o ⊔ ℓ ⊔ e ⊔ ℓ' ⊔ e')) where
+  field
+    Obj : Set o
+    Hor : Cat.Category Obj ℓ e
+    Ver : Cat.Category Obj ℓ e
```
JacquesCarette: Should the `Vec` category have primed levels?

JacquesCarette (changes requested):
Looks like there's still some stuff that needs looked at.

src/Categories/Double/Core.agda
Boarders: I think so, I'll change it so there can be 5 levels as you mentioned.

Boarders:
@JacquesCarette Thanks for the review, I'll make these changes shortly.

src/Categories/Double/Core.agda
maxsnew: We could call them "frames" rather than perimeters to follow Mike Shulman's "framed Bicategory" terminology.

JacquesCarette: If that's a correct use of the term (I don't know - reference?), then I definitely like it!

Boarders:
@JacquesCarette Apologies for the huge delay, life got in the way. I have made the changes you suggested and updated the name to `FrameEquality`. Let me know if there are more style changes you would like and I'll be happy to get this over the line.

JacquesCarette:
No worries. My life is quite eventful in all of July, this may end up waiting until early August. Though I will try next week, we'll see.

JacquesCarette:
Oh my, this completely slipped through. Reviewing now.

JacquesCarette (approved):

PR #348, opened 2022-04-22
[draft] Add definition of an adjunction in Bicategory

Boarders:
Hoping to get some feedback on design decisions. I am still planning to prove that this gives an equivalence in terms of hom-categories. If there is anything else that should be included feel free to let me know.


src/Categories/Bicategory/Adjunction.agda
```diff
@@ -0,0 +1,56 @@
...
+private
+  variable
+    o ℓ e t : Level
+
+module _ (𝒞 : Bicategory o ℓ e t) where
+  open Bicategory 𝒞
+
+  record Adjunction (A B : Obj) : Set (o ⊔ ℓ ⊔ e ⊔ t) where
+    private
+      module C = Extras 𝒞
```
TOTBWF, line 19: We should probably pull this module out of the record, and into the enclosing module.

src/Categories/Bicategory/Adjunction.agda
```diff
@@ -0,0 +1,56 @@
...
+                          ↓            L
+                          R
+-}
+      r-triangle-l : id₁ ⊚₀ R ⇒₂ R
+      r-triangle-l = C.unitorʳ.from ∘ᵥ (id₂ ⊚₁ counit) ∘ᵥ C.associator.from ∘ᵥ (unit ⊚₁ id₂)
+
+      r-triangle-r : id₁ ⊚₀ R ⇒₂ R
+      r-triangle-r = C.unitorˡ.from
+    field
+      l-triangle : l-triangle-l C.≈ l-triangle-r
```
TOTBWF, line 55: Perhaps we should name these `zig` and `zag`  for alignment with `Categories.Adjoint`.

TOTBWF (changes requested):
Looks good! A few minor nits, but I like where this is going 🙂.

If you are up to the task, it would be cool to prove that this is equivalent to the elementary definition in `Cat` like we do in `Bicategory.Monad.Properties`. If not, that's totally fine!

src/Categories/Bicategory/Adjunction.agda
JacquesCarette: Yes. And may as well open it extracting the unitors and ≈.

src/Categories/Bicategory/Adjunction.agda
JacquesCarette: Agreed!

JacquesCarette (changes requested):
Agree with @TOTBWF : looks good, but a couple of tweaks would make it even better, and ready for inclusion.

JacquesCarette:
@Boarders this was really good - a couple of small changes, and it's ready to go in. Will you be able to do there? Or would like us to?

Boarders:
@JacquesCarette Sorry for the delay, I'll be happy to get this finished soon (give me a week or so as I have a current knee injury which makes too much time at the computer somewhat tricky)

JacquesCarette:
No worries! I'm patient and not in a hurry. Don't strain your knee for this.

JacquesCarette:
I'd like to get this done - if you give me permission to push to your fork, I can do those small changes needed and then merge it in?

PR #350, opened 2022-05-16
Removing a redundant field in CartesianClosed/Canonical

MiddleAdjunction:
In the canonical definition of CCCs, the field `curry-resp-≈` is redundant and can be defined as a function in the record. I personally cannot see a reason justifying the redundancy but it might be there to improve definitional equality in some way (similar to introducing the redundant `sym-assoc` as opposed to a function using `assoc`), in which case there possibly needs to be a comment nearby explaining such reason.


sstucki:
LGTM! When I wrote this module I didn't realize this assumption was redundant. I also don't see any good reason to keep it.

JacquesCarette (approved):
Sorry to be so slow in reviewing. Thanks a lot for your PR.

PR #352, opened 2022-06-30
make it work with the agda stdlib v2.0 (c12568c64)

cmcmA20:
^subj

### DIFF
--- a/agda-categories.agda-lib
+++ b/agda-categories.agda-lib
@@ -1,3 +1,3 @@
 name: agda-categories
-depend: standard-library-1.7.2
+depend: standard-library-2.0
 include: src/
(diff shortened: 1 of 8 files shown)


MatthewDaggitt:
A heads up that `v2.0` is not released so I wouldn't recommend merging this in, unless you want your users to have to upgrade to the unstable library version.

cmcmA20:
Let it chill until then, I will update it after release.

cmcmA20:
Tested with agda master `269204b682`
Works on agda-stdlib master `7c5f3ff90`

JacquesCarette:
Thanks.  [Note: there seems to be two outstanding conflicts?]

JacquesCarette:
So now that `stdlib` 2.0 is 'essentially' live, it would be nice to revisit this. Looks like things have broken since this was last looked at.

JacquesCarette:
If you give me permission to push to your branch, I can help.

JacquesCarette:
In the end, 64 different places needed to be fixed to push things to 2.0, and this was done in #406 which has been merged. So this one is no longer needed - but keep PRs coming!

PR #355, opened 2022-10-06
Simplex: degen is pinch from agda-stdlib

phadej:


phadej:
ping

JacquesCarette (approved):
Perfect, thanks. Now if only stdlib had the equivalent to `degen-comm` too!

PR #356, opened 2022-10-09
Rels is also a symmetric monoidal category with Set's cartesian product

Taneb:


JacquesCarette (approved):
This is coherent with what's already in the library, so I'll merge it in as is.

However, this all looks very strict. I'm thinking maybe this should be called StrictRels. But that can be for the future.

PR #359, opened 2022-11-01
Add `IsNI` to be consistent with `IsIso` in for morphisms

91khr:
Sometimes we may say that a natural transformation is a natural isomorphism, so it may be better to add a helper to express this.

This is my first time trying to contribute to an Agda project, please excuse if there's anything missing.

### DIFF
--- a/src/Categories/NaturalTransformation/NaturalIsomorphism.agda
+++ b/src/Categories/NaturalTransformation/NaturalIsomorphism.agda
@@ -71,6 +71,31 @@ record NaturalIsomorphism {C : Category o ℓ e}
       }
     }
 
+-- Like IsIso for morphisms, a natural transformation may be an isomorphism
+-- if there exists some inverse to it.
+-- To be consistent, there's also an IsNI for natural isomorphisms
+record IsNI {C : Category o ℓ e}
+            {D : Category o′ ℓ′ e′}
+            {F G : Functor C D}
+            (F⇒G : NaturalTransformation F G) : Set (o ⊔ ℓ ⊔ ℓ′ ⊔ e′) where
+  field
+    inv : NaturalTransformation G F
+
+  module F⇒G = NaturalTransformation F⇒G
+  module inv = NaturalTransformation inv
+
+  field
+    F⇒G-iso : ∀ X → Morphism.Iso D (F⇒G.η X) (inv.η X)
+
+  natiso : NaturalIsomorphism F G
+  natiso = record
+    { F⇒G = F⇒G
+    ; F⇐G = inv
+    ; iso = F⇒G-iso
+    }
+
+  open NaturalIsomorphism natiso public
+
 -- This helper definition lets us specify only one of the commuting
 -- squares and have the other one derived.
 
--- a/src/Categories/NaturalTransformation/NaturalIsomorphism.agda
+++ b/src/Categories/NaturalTransformation/NaturalIsomorphism.agda
@@ -79,22 +79,15 @@ record IsNI {C : Category o ℓ e}
             {F G : Functor C D}
             (F⇒G : NaturalTransformation F G) : Set (o ⊔ ℓ ⊔ ℓ′ ⊔ e′) where
   field
-    inv : NaturalTransformation G F
-
-  module F⇒G = NaturalTransformation F⇒G
-  module inv = NaturalTransformation inv
-
-  field
-    F⇒G-iso : ∀ X → Morphism.Iso D (F⇒G.η X) (inv.η X)
+    F⇐G : NaturalTransformation G F
+    iso : ∀ X → Morphism.Iso D (NaturalTransformation.η F⇒G X) (NaturalTransformation.η F⇐G X)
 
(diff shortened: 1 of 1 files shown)


src/Categories/NaturalTransformation/NaturalIsomorphism.agda
```diff
@@ -71,6 +71,31 @@ record NaturalIsomorphism {C : Category o ℓ e}
...
+-- To be consistent, there's also an IsNI for natural isomorphisms
+record IsNI {C : Category o ℓ e}
+            {D : Category o′ ℓ′ e′}
+            {F G : Functor C D}
+            (F⇒G : NaturalTransformation F G) : Set (o ⊔ ℓ ⊔ ℓ′ ⊔ e′) where
+  field
+    inv : NaturalTransformation G F
+
+  module F⇒G = NaturalTransformation F⇒G
+  module inv = NaturalTransformation inv
```
JacquesCarette, line 85: For consistency with NaturalIsomorphism`, I think `inv` should be called `F⇐G` (everywhere in `IsNI`).

src/Categories/NaturalTransformation/NaturalIsomorphism.agda
```diff
@@ -71,6 +71,31 @@ record NaturalIsomorphism {C : Category o ℓ e}
...
+-- if there exists some inverse to it.
+-- To be consistent, there's also an IsNI for natural isomorphisms
+record IsNI {C : Category o ℓ e}
+            {D : Category o′ ℓ′ e′}
+            {F G : Functor C D}
+            (F⇒G : NaturalTransformation F G) : Set (o ⊔ ℓ ⊔ ℓ′ ⊔ e′) where
+  field
+    inv : NaturalTransformation G F
+
+  module F⇒G = NaturalTransformation F⇒G
```
JacquesCarette, line 84: As these 2 modules are only used once each, and only for 1 field, I don't think they are worth defining. [I know this is the pattern in other parts of the library, but I'm moving away from it as it causes a fair amount of bloat in the actual interfaces.]

You could keep them if you 1) make them private, and 2) add `using (η)` to them. Then they are quite thin, and that's fine.

src/Categories/NaturalTransformation/NaturalIsomorphism.agda
```diff
@@ -71,6 +71,31 @@ record NaturalIsomorphism {C : Category o ℓ e}
...
+    F⇒G-iso : ∀ X → Morphism.Iso D (F⇒G.η X) (inv.η X)
+
+  natiso : NaturalIsomorphism F G
+  natiso = record
+    { F⇒G = F⇒G
+    ; F⇐G = inv
+    ; iso = F⇒G-iso
+    }
+
+  open NaturalIsomorphism natiso public
```
JacquesCarette, line 97: Let's not have this at all, at least right now, that definitely adds too much noise to the interface. If in usage patterns, we find that using `IsNI` is "long winded" because of extra qualification, we can revisit this.

JacquesCarette (changes requested):
Thanks. Great first contribution. Just some tiny style things to tweak and it can be merged in.

src/Categories/NaturalTransformation/NaturalIsomorphism.agda
91khr: Thank you for your review!

It's true that `F⇐G` is more consistent with NaturalIsomorphism; a concern here is that the name `inv` is taken from `IsIso`, and should `IsNI` try to be consistent with `IsIso`... ?

src/Categories/NaturalTransformation/NaturalIsomorphism.agda
```diff
@@ -71,6 +71,26 @@ record NaturalIsomorphism {C : Category o ℓ e}
...
+-- Like IsIso for morphisms, a natural transformation may be an isomorphism
+-- if there exists some inverse to it.
+-- To be consistent, there's also an IsNI for natural isomorphisms
+record IsNI {C : Category o ℓ e}
+            {D : Category o′ ℓ′ e′}
+            {F G : Functor C D}
+            (F⇒G : NaturalTransformation F G) : Set (o ⊔ ℓ ⊔ ℓ′ ⊔ e′) where
+  field
+    F⇐G : NaturalTransformation G F
+
```
JacquesCarette, line 83: You can now delete this blank line and extra `field`

91khr: Sorry for the misstyling ><

JacquesCarette: No problem! The time to review is small compared to the effort of doing PRs.

JacquesCarette (approved):

src/Categories/NaturalTransformation/NaturalIsomorphism.agda
JacquesCarette: It's a tough call. I think the instance of `NaturalIsomorphism` below 'wins' as far as consistency is concerned.

PR #360, opened 2022-11-21
Parametric adjunctions form cowedges

iwilare:
We define the notion of parametric adjunction Lc ⊣ Rc between functors  L : C x D → E and R : Cᵒ x E → D and show that, under suitable assumptions, the counit of the adjunction ε[c,x] : L(c,R(c,x)) → x is a cowedge in the parameter c.

For more information consult [this question on StackExchange](https://math.stackexchange.com/questions/4580923/parametric-adjunctions-form-cowedges).

j.w.w. @tetrapharmakon

### DIFF
--- a/src/Categories/Adjoint/Parametric.agda
+++ b/src/Categories/Adjoint/Parametric.agda
@@ -2,30 +2,22 @@ module Categories.Adjoint.Parametric where
 
 open import Level
 
-open import Data.Product
+open import Data.Product using (_×_; _,_)
 open import Categories.Category.Core using (Category)
 open import Categories.Functor using (Functor; _∘F_) renaming (id to idF)
 open Categories.Functor.Functor using (F₀; F₁; homomorphism; identity; F-resp-≈)
-open import Categories.Adjoint
-open import Categories.Category.Construction.Functors
-open import Categories.Diagram.Cowedge
-open import Categories.NaturalTransformation.Dinatural
-open import Categories.Functor.Bifunctor
-open import Categories.Morphism.Properties
-open import Categories.Functor.Bifunctor.Properties
-open import Categories.NaturalTransformation
-open import Categories.NaturalTransformation.NaturalIsomorphism
-open import Categories.Morphism.Reasoning as MorphismReasoning
+open import Categories.Adjoint using (Adjoint)
+open import Categories.Category.Construction.Functors using (Functors)
+open import Categories.Diagram.Cowedge using (Cowedge)
+open import Categories.NaturalTransformation.Dinatural using (dtHelper)
+open import Categories.Functor.Bifunctor using (Bifunctor)
+import Categories.NaturalTransformation
+open Categories.NaturalTransformation.NaturalTransformation using (η; commute)
 import Categories.Morphism.Reasoning as MR
-open Categories.NaturalTransformation.NaturalTransformation using (η)
-open import Function.Equality using (_⟨$⟩_; _⟶_; cong; _∘_)
-
-open import Categories.Adjoint.Equivalents
 
 private
   variable
     o ℓ e : Level
-    C D E : Category o ℓ e
 
 -- A parametric adjunction is a family of functors Lc ⊣ Rc indexed by an object c of
 -- a category C; equivalently, a pair of functors L : C x D → E and R : Cᵒ x E → D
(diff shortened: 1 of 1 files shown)


src/Categories/Adjoint/Parametric.agda
```diff
@@ -0,0 +1,150 @@
...
+    is commutative; unfolding the definitions, it seems that proving this exactly amounts to show
+    that the *unit* η c of the adjunction is a *wedge*. Viceversa, one also needs this naturality
+    when proving that the cowedge condition holds, so they seem to be logically equivalent.
+    Mac Lane (IV.7.Thm 3) seems to claim, however, that this condition needs not to be assumed
+    and can be derived from a better definition of parametric adjunction that we are not
+    currently able to devise.
+
+    More info at: https://math.stackexchange.com/questions/4581092/parametric-adjunctions-induce-cowedges-a-real-proof
+  -}
+
```
tetrapharmakon, line 68: @JacquesCarette this is the major open point of the PR, otherwise, we believe it is ready for review: do you have any suggestion on how to address this?

src/Categories/Adjoint/Parametric.agda
```diff
@@ -0,0 +1,150 @@
...
+import Categories.Morphism.Reasoning as MR
+open Categories.NaturalTransformation.NaturalTransformation using (η)
+open import Function.Equality using (_⟨$⟩_; _⟶_; cong; _∘_)
+
+open import Categories.Adjoint.Equivalents
+
+private
+  variable
+    o ℓ e : Level
+    C D E : Category o ℓ e
```
JacquesCarette, line 28: No point creating generalized variables here, only to immediately shadow them in the next definition...

src/Categories/Adjoint/Parametric.agda
```diff
@@ -0,0 +1,150 @@
...
+-- such that for every c in C Lc = L(c,-) ⊣ R(c,-) = Rc.
+record ParametricAdjoint {C D E : Category o ℓ e} (L : Functor C (Functors D E)) (R : Functor (Category.op C) (Functors E D)) : Set (o ⊔ ℓ ⊔ e) where
+  private
+    module C = Category C
+    module D = Category D
+    module E = Category E
+    open MR E
+    module L = Functor L
+    module R = Functor R
+    module R₀ A = Functor (R.F₀ A)
```
JacquesCarette, line 41: As you end up using components of `R₀` and `L₀`, the results end up being not-very-readable.

One suggestion: name the arguments to `ParametricAdjoint` LL and RR, that would free up L and R to be used here. [Feel free to do something different to improve readability, which is my real concern.]

src/Categories/Adjoint/Parametric.agda
JacquesCarette: I was going to say that it's usually better in `agda-categories` to work in terms of unit/counit, but I see that `param-nat` is already phrased that way.

Perhaps there's a lead in https://golem.ph.utexas.edu/category/2012/11/parametrized_mates_and_multiva.html ?

src/Categories/Adjoint/Parametric.agda
```diff
@@ -0,0 +1,150 @@
...
+    ; dinatural = dtHelper record
+      { α = λ c → A.counit.η c _
+      ; commute = λ {X} {Y} f →
+        let
+          open Adjoint
+          adjunction-isoˡ : A.Ladjunct X (A.counit.η X A E.∘ L₀.₁ X (η (R.₁ f) A)) D.≈
+                            η (R.₁ f) A
+          adjunction-isoˡ = let open D.HomReasoning in
+            begin _ ≈⟨ homomorphism (R.F₀ X) ⟩∘⟨refl ⟩
+                  _ ≈⟨ D.assoc ⟩
```
JacquesCarette, line 112: The various `pull` and `push` combinators can help shorten these proofs (here by 2 lines).

src/Categories/Adjoint/Parametric.agda
```diff
@@ -0,0 +1,150 @@
+module Categories.Adjoint.Parametric where
+
+open import Level
+
+open import Data.Product
```
JacquesCarette, line 5: Adding some extra `using` below would be appreciated.

JacquesCarette (changes requested):

src/Categories/Adjoint/Parametric.agda
iwilare: I agree, `L.₀.₁` is not really readable. I'd opt to use double (2) letters whenever we refer to a functor in two (2) parameters; we still need to use `L` and `R` in the usual way more often than the two-parameters case, in any case.

src/Categories/Adjoint/Parametric.agda
iwilare: Done, also cleaned up some unused imports.

JacquesCarette (approved):

PR #362, opened 2022-11-22
μ-Bialgebras

cxandru:
I've defined bialgebras for functors. I needed to factor out some definitions in `Algebra` and `Coalgebra`, but the changes don't break any existing code. For more details see the commit message of f8a60272a3aff9284cac5a481dab91da932727ea

### DIFF
--- a/src/Categories/Category/Construction/F-Coalgebras.agda
+++ b/src/Categories/Category/Construction/F-Coalgebras.agda
@@ -108,23 +108,29 @@ termToInit {C = C} {F = F} {T = T} isTT = record
   { ! =
       F-Coalgebra-Morphism⇒coF-Algebra-Morphism ¡
   ; !-unique =
-      λ γ →
+      λ {J} γ  →
         let
-          known : F-Coalgebras F [ ¡ ≈ coF-Algebra-Morphism⇒F-Coalgebra-Morphism γ ]
-          known = ¡-unique (coF-Algebra-Morphism⇒F-Coalgebra-Morphism γ)
+          J' : F-Algebra (Functor.op F)
+          J' = J
+          γ' : F-Algebra-Morphism (F-Coalgebra⇒coF-Algebra T) J
+          γ' = γ
+          known : (F-Coalgebras F) [ ¡ ≈ coF-Algebra-Morphism⇒F-Coalgebra-Morphism γ ]
+          known = ¡-unique {coF-Algebra⇒F-Coalgebra J} (coF-Algebra-Morphism⇒F-Coalgebra-Morphism γ)
           bar :
             F-Algebras (Functor.op F)
               [ F-Coalgebra-Morphism⇒coF-Algebra-Morphism ¡
-                ≈ F-Coalgebra-Morphism⇒coF-Algebra-Morphism (coF-Algebra-Morphism⇒F-Coalgebra-Morphism γ) ]
-          bar = Functor.F-resp-≈ (F-Co-M⇒coF-M F) known
+                ≈ γ ]
+          bar = Functor.F-resp-≈ (F-Co-M⇒coF-M F) {coF-Algebra⇒F-Coalgebra J} {T} {¡} {coF-Algebra-Morphism⇒F-Coalgebra-Morphism γ} known
+          baz : (Category.op C) [ (F-Coalgebra.A T) , F-Algebra.A J ]
+          baz = F-Algebra-Morphism.f γ
+          baz2 : C [ F-Algebra.A J , (F-Coalgebra.A T) ]
+          baz2 = baz
+          -- notably, this is the flipped diagram in C that it would be in C^op.
+          γcommutes : C [ C [ (F-Coalgebra.α T) ∘ baz ]  ≈  C [ Functor.F₁ F baz ∘ F-Algebra.α J ] ]
+          γcommutes = F-Algebra-Morphism.commutes γ
           result : F-Algebras (Functor.op F)
             [ F-Coalgebra-Morphism⇒coF-Algebra-Morphism ¡ ≈ γ ]
-          result = begin
-            F-Coalgebra-Morphism⇒coF-Algebra-Morphism ¡ ≈⟨ bar ⟩
-            F-Coalgebra-Morphism⇒coF-Algebra-Morphism (coF-Algebra-Morphism⇒F-Coalgebra-Morphism γ)
-            ≈⟨ refl ⟩
-            γ ∎
-
+          result = bar
         in result
   }
   where
--- a/src/Categories/Category/Construction/F-Coalgebras.agda
+++ b/src/Categories/Category/Construction/F-Coalgebras.agda
@@ -30,7 +30,7 @@ F-Coalgebras {C = C} F = record
        coF-Algebras.∘
        F-Coalgebra-Morphism⇒coF-Algebra-Morphism α₂)
     }
-  ; id        = coF-Algebra-Morphism⇒F-Coalgebra-Morphism coF-Algebras.id
+  ; id        = record { f = id; commutes = F-Algebra-Morphism.commutes coF-Algebras.id }
   ; assoc     = assoc
   ; sym-assoc = sym-assoc
   ; identityˡ = identityˡ
@@ -57,12 +57,7 @@ F-Coalgebras⇒coF-Algebras {C = C} F = record
   ; F₁           = F-Coalgebra-Morphism⇒coF-Algebra-Morphism
   ; identity     = refl
   ; homomorphism = refl
-  ; F-resp-≈     = λ {_ _ α₁ α₂} α₁≈α₂ →
-     begin
-      F-Algebra-Morphism.f (F-Coalgebra-Morphism⇒coF-Algebra-Morphism α₁) ≈⟨ refl ⟩
-      f α₁ ≈⟨ α₁≈α₂ ⟩
-      f α₂ ≈⟨ refl ⟩
-      F-Algebra-Morphism.f (F-Coalgebra-Morphism⇒coF-Algebra-Morphism α₂) ∎
+  ; F-resp-≈     = λ x → x
   }
   where
     open Category C
@@ -92,7 +87,7 @@ private
       open HomReasoning
       open Equiv
 
-private module CoLambek {C : Category o ℓ e} {F : Endofunctor C} (T : Terminal (F-Coalgebras F)) where
+module CoLambek {C : Category o ℓ e} {F : Endofunctor C} (T : Terminal (F-Coalgebras F)) where
   open Category C
   open Functor F using (F₀)
   open F-Coalgebra using (α)
--- a/src/Categories/Category/Construction/F-Coalgebras.agda
+++ b/src/Categories/Category/Construction/F-Coalgebras.agda
@@ -24,10 +24,12 @@ F-Coalgebras {C = C} F = record
   { Obj       = F-Coalgebra F
   ; _⇒_       = F-Coalgebra-Morphism
   ; _≈_       = λ α₁ α₂ → f α₁ ≈ f α₂
-  ; _∘_       = λ α₁ α₂ → coF-Algebra-Morphism⇒F-Coalgebra-Morphism
-    (F-Coalgebra-Morphism⇒coF-Algebra-Morphism α₁
-    coF-Algebras.∘
-    F-Coalgebra-Morphism⇒coF-Algebra-Morphism α₂)
+  ; _∘_       = λ α₁ α₂ → record
+    { f = f α₁ ∘ f α₂
+    ; commutes = F-Algebra-Morphism.commutes (F-Coalgebra-Morphism⇒coF-Algebra-Morphism α₁
+       coF-Algebras.∘
+       F-Coalgebra-Morphism⇒coF-Algebra-Morphism α₂)
+    }
   ; id        = coF-Algebra-Morphism⇒F-Coalgebra-Morphism coF-Algebras.id
   ; assoc     = assoc
   ; sym-assoc = sym-assoc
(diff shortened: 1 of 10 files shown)


src/Categories/Category/Construction/μ-Bialgebras.agda
```diff
@@ -0,0 +1,68 @@
...
+    open Equiv
+    open μ-Bialgebra-Morphism
+    open μ-Bialgebra
+    module T-Algebras = Category (F-Algebras T)
+    open Functor F
+    open F-Coalgebra-Morphism
+    open F-Coalgebra
+    commut : {A B C : F-Coalgebra F} (α₁ : F-Coalgebra-Morphism B C) (α₂ : F-Coalgebra-Morphism A B) →
+      α C ∘ (f α₁ ∘ f α₂) ≈  F₁ (f α₁ ∘ f α₂) ∘ α A
+    commut {A} {B} {C} α₁ α₂ = begin
```
JacquesCarette, line 62: Please line up the equational proof.

JacquesCarette (changes requested):

Taneb:
We don't currently have any modules with a non-ascii character in the name. Is it OK to add one in this case?

JacquesCarette:
Good point. I think that's asking for trouble, we shouldn't have file names with unicode in them. Internal modules is fine, but not "file-modules".

src/Categories/Functor/Bialgebra.agda
```diff
@@ -0,0 +1,66 @@
...
+[category theory - How are F-bialgebras defined? - Mathematics Stack Exchange](https://math.stackexchange.com/questions/3057781/how-are-f-bialgebras-defined#answer-3057859)
+-}
+
+private
+  variable
+    o ℓ e : Level
+
+module _ {C : Category o ℓ e} where
+
+  module _ (T : Endofunctor C) (F : Endofunctor C) where
```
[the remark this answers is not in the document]
cxandru: `Categories.NaturalTransformation` imports `Categories.Functor.Properties`. Perhaps it's better off in `Categories.NaturalTransformation.Properties`? The reason why I didn't want to place it somewhere very globally is because Distributive Laws are generally defined for monads and since I'm only using them for functors here I don't require the usual extra properties: https://ncatlab.org/nlab/show/distributive+law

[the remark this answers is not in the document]
cxandru: Ah I forgot to write it out: The problem is that putting it in `Categories.Functor.Properties` creates an import cycle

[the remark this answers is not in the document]
JacquesCarette: Ah, now *that* is a real problem! `Categories.Functor.DistributiveLaw` it is then?

src/Categories/Functor/Bialgebra.agda
```diff
@@ -0,0 +1,66 @@
...
+    field
+      A : Obj
+      a₁ : F-Algebra-on T A
+      c₁ : F-Coalgebra-on F A
+    module F = Functor F
+    module T = Functor T
+    field
+      respects-μ : c₁ ∘ a₁ ≈ ((F.₁) a₁) ∘ (μ .η A) ∘ ((T.₁) c₁)
+    --to access the (co)algebras induced by the morphisms:
+    a : F-Algebra T
```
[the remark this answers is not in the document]
cxandru: The problem is: a bialgebra morphism is a C-morphism which _is_ a T-Algebra morphism _and_ an F-Coalg morphism. I use these accessors at line 59 &60 to define the commutation properties for bialgebra morphisms. The reason I needed to factor `is-F-Algebra-Morphism` out of `F-Algebra-Morphism` and `F-Algebra-on` out of `F-Algebra` respectively is because the latter are essentially the former with the object(s) hidden under an existential.
What I _could_ do is define helpers that "package" e.g. `f0: F-Algebra-on A0` into its existential `record{f=f0; A = A0} : F-Algebra`, which I could define in `F-Algebra`. I'm not sure the conversion can be elided entirely though.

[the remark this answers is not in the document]
cxandru: Aren't `a` and `c` in the record module `μ-Bialgebra` though? so they shouldn't escape into the global namespace. I added them because for the definition we need to pass `a₁` and `c₁`, but for usability you will usually want to access the (co)algebras induced by them. They are meant to be used in a way akin to the `₀ = F₀;  ₁ = F₁` shorthands in the `Functor` record module.

src/Categories/Functor/Bialgebra.agda
```diff
@@ -0,0 +1,66 @@
...
+module Categories.Functor.Bialgebra where
+
+open import Level
+open import Categories.Category using (Category)
+open import Categories.Functor
+open import Categories.Functor.Algebra
+open import Categories.Functor.Coalgebra
+open import Categories.NaturalTransformation
+
+{-
```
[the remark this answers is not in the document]
cxandru: Fixed

src/Categories/Functor/Bialgebra.agda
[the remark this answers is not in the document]
JacquesCarette: `a` and `c` will escape if your open this record.  And those are short, not-really-meaningful names. Could they be defined locally at the use site instead?

Functors are pervasive, so using a few shorthands (that are not likely to already be used in user code) makes sense. And it took a lot of discussion to settle on these.

src/Categories/Category/Construction/F-Coalgebras.agda
```diff
@@ -0,0 +1,129 @@
...
+  { Obj       = F-Coalgebra F
+  ; _⇒_       = F-Coalgebra-Morphism
+  ; _≈_       = λ α₁ α₂ → f α₁ ≈ f α₂
+  ; _∘_       = λ α₁ α₂ → record
+    { f = f α₁ ∘ f α₂
+    ; commutes = F-Algebra-Morphism.commutes (F-Coalgebra-Morphism⇒coF-Algebra-Morphism α₁
+       coF-Algebras.∘
+       F-Coalgebra-Morphism⇒coF-Algebra-Morphism α₂)
+    }
+  ; id        = coF-Algebra-Morphism⇒F-Coalgebra-Morphism coF-Algebras.id
```
JacquesCarette, line 33: I fear that this remains a bad idea, i.e. you are better off defining `id` directly. `commutes` above is ok.

src/Categories/Category/Construction/F-Coalgebras.agda
```diff
@@ -0,0 +1,129 @@
...
+    open F-Coalgebra-Morphism
+    module coF-Algebras = Category (Category.op (F-Algebras (Functor.op F)))
+
+F-Coalgebras⇒coF-Algebras : {C : Category o ℓ e} → (F : Endofunctor C) → Functor (F-Coalgebras F) (Category.op (F-Algebras (Functor.op F)))
+F-Coalgebras⇒coF-Algebras {C = C} F = record
+  { F₀           = F-Coalgebra⇒coF-Algebra
+  ; F₁           = F-Coalgebra-Morphism⇒coF-Algebra-Morphism
+  ; identity     = refl
+  ; homomorphism = refl
+  ; F-resp-≈     = λ {_ _ α₁ α₂} α₁≈α₂ →
```
JacquesCarette, line 60: This proof looks complicated, but since it's just `α₁≈α₂` (given enough evaluation), maybe make it just that?

src/Categories/Category/Construction/F-Coalgebras.agda
```diff
@@ -0,0 +1,129 @@
...
+        (¡-unique (coF-Algebra-Morphism⇒F-Coalgebra-Morphism γ))
+    }
+    where
+      open Category (F-Algebras (Functor.op F))
+      open MR (F-Algebras (Functor.op F))
+      open IsTerminal isTT renaming (! to ¡; !-unique to ¡-unique)
+      open HomReasoning
+      open Equiv
+
+private module CoLambek {C : Category o ℓ e} {F : Endofunctor C} (T : Terminal (F-Coalgebras F)) where
```
JacquesCarette, line 95: If this `module` is private in this file, why have it here at all?

src/Categories/Category/Construction/mu-Bialgebras.agda
```diff
@@ -0,0 +1,63 @@
+{-# OPTIONS --without-K --safe #-}
+module Categories.Category.Construction.mu-Bialgebras where
+
+open import Level
+open import Function using (_$_)
+
+open import Categories.Category
```
JacquesCarette, line 7: Could you please put more `using` clauses on your `import`s?

src/Categories/Category/Construction/mu-Bialgebras.agda
```diff
@@ -0,0 +1,63 @@
...
+private
+  variable
+    o ℓ e : Level
+
+μ-Bialgebras : {C : Category o ℓ e} → (T F : Endofunctor C) → (DistributiveLaw T F) → Category (o ⊔ ℓ ⊔ e) (o ⊔ ℓ ⊔ e) e
+μ-Bialgebras {C = C} T F μ = record
+ { Obj       = μ-Bialgebra T F μ
+ ; _⇒_       = μ-Bialgebra-Morphism
+ ; _≈_       = λ β₁ β₂ → f β₁ ≈ f β₂
+ ; _∘_       = λ β₁ β₂ → record
```
JacquesCarette, line 26: It's probably a good idea to define `μ-Bialgebra-Morphism` composition where your define `μ-Bialgebra-Morphism` rather than inline here. Same with `id`. [Many other parts of the library do that already.] The rest can stay here.

JacquesCarette (changes requested):
Sorry to be so picky (and slow). This is a very nice piece of work! Might as well really polish it up.

src/Categories/Category/Construction/F-Coalgebras.agda
cxandru: I guess at this point I wasn't really familiar enough with Agda yet. λ x → x does the trick

src/Categories/Category/Construction/F-Coalgebras.agda
cxandru: I guess that was just a mistake. Whoops

src/Categories/Category/Construction/mu-Bialgebras.agda
cxandru: Ah, I was mostly inspired by `Algebra` in relation to `F-Algebras` (where composition is defined in the Category construction `F-Algebras`)

cxandru: Could you give a concrete example that I could use as reference?

src/Categories/Category/Construction/F-Coalgebras.agda
cxandru: Done c789ca7d

src/Categories/Category/Construction/F-Coalgebras.agda
cxandru: Done c789ca7d

src/Categories/Category/Construction/F-Coalgebras.agda
cxandru: Done c789ca7d

JacquesCarette (commented):
This is looking really good. Is it "ready"?

src/Categories/Functor/Bialgebra.agda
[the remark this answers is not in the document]
cxandru: I renamed them to `alg` & `coalg` respectively in b116824298963c953a67f3145d8dd5ad1f48a83f,  is that fine?

cxandru:
> This is looking really good. Is it "ready"?

Almost! I'm currently implementing a final interesting lemma, that initial objects in T-Algs are induce initial objs in μ-F-T-Bialgebras (dually, final F-Coalgs are also final in μ-F-T-Bialgebras). I also replied to some of your review comments with questions

JacquesCarette:
[mu-Bialgebras](https://trello.com/c/PeIDABf1/288-mu-bialgebras)

src/Categories/Functor/Bialgebra.agda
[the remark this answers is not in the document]
JacquesCarette: excellent.

src/Categories/Category/Construction/mu-Bialgebras.agda
JacquesCarette: [Categories.Category.Construction.Comma](https://agda.github.io/agda-categories/Categories.Category.Construction.Comma.html) does it in a `where` block.  [Functors](https://agda.github.io/agda-categories/Categories.Category.Construction.Functors.html#1304) defines it elsewhere (but it's cheating a little bit).

I think [LawvereTheories](https://agda.github.io/agda-categories/Categories.Category.Instance.LawvereTheories.html) is a good example of the 'separate' style.

src/Categories/Category/Construction/mu-Bialgebras.agda
cxandru: If I use more than 3 things from a module it becomes a kind of egg-hunt to post hoc add them all as qualified imports (add members, check errors, repeat), and even more annoying to do it as you go (if you want to use a new member, you always need to go back and add it to the import list, also, you need to remove members you no longer use).
I think it would really help this project (and agda development in general) if there was language support for import management as there is in e.g. Haskell, with `haskell-language-server`. I opened an issue on `agda-unused` to ask why module applications are such a stumbling block, but haven't heard back yet: https://github.com/msuperdock/agda-unused/issues/1 .

cxandru:
I added some last nice things to round off this MR (see commit descriptions). After review everything should be ready to merge!

src/Categories/Category/Construction/mu-Bialgebras.agda
JacquesCarette: I agree, it's a big pain. It should be a feature of Agda.

JacquesCarette:
Thanks - at POPL right now, so might not get to do the review until I'm back home. We'll see.

cxandru:
> Thanks - at POPL right now, so might not get to do the review until I'm back home. We'll see.

yes, I saw on Mastodon

JacquesCarette (approved):
Excellent!

PR #364, opened 2022-12-09
Add uncurry and composition with curry

iwilare:
Define uncurry and provide the isomorphism that shows Cat is cartesian closed.


JacquesCarette (approved):
This is fine, althoughfor future PRs I'm not a big fan of using all `_` in the lhs of equational reasoning. I can see eliding some of them, but perhaps not all?

PR #365, opened 2022-12-20
Make Rels less strict, add a StrictRels for the previous version

Taneb:


Taneb:
I'd like comments now definitely, please! I still need to copy the top-level comment from (what is now) `Categories.Category.Monoidal.Instance.StrictRels` to `Categories.Category.Monoidal.Rels`.

src/Categories/Category/Dagger/Instance/Rels.agda
```diff
@@ -1,24 +1,37 @@
...
-open import Data.Product
-open import Function
-open import Relation.Binary.PropositionalEquality
-open import Level
+module Categories.Category.Dagger.Instance.Rels where
 
 open import Categories.Category.Dagger
 open import Categories.Category.Instance.Rels
 
+open import Data.Product
```
JacquesCarette, line 8: Please put explicit `using` in as many places as possible, strongly so on stdlib imports.

src/Categories/Category/Instance/StrictRels.agda
```diff
@@ -0,0 +1,36 @@
...
+              , (λ { (b , fxb , c , gbc , hcy) → c , ((b , (fxb , gbc)) , hcy)})
+  ; identityˡ = (λ { (b , fxb , lift refl) → fxb}) , λ {_} {y} fxy → y , fxy , lift refl
+  ; identityʳ = (λ { (a , lift refl , fxy) → fxy}) , λ {x} {_} fxy → x , lift refl , fxy
+  ; identity² = (λ { (_ , lift p , lift q) → lift (trans p q)}) , λ { (lift refl) → _ , lift refl , lift refl }
+  ; equiv = record
+    { refl = id , id
+    ; sym = swap
+    ; trans = λ { (p₁ , p₂) (q₁ , q₂) → (q₁ ∘′ p₁) , p₂ ∘′ q₂}
+    }
+  ; ∘-resp-≈ = λ f⇔h g⇔i →
```
JacquesCarette, line 33: as you're already using pattern-matching lambdas, might as well continue to do so instead of using proj  here?

src/Categories/Category/Monoidal/Instance/Rels.agda
```diff
@@ -1,204 +1,284 @@
...
+            }
+          }
+        ; π₂ = record
+          { fst = [ (λ _ _ → ⊥) , Setoid._≈_ B ]′
+          ; snd = record
+            { fst = λ { (inj₂ p) q → Setoid.trans B (Setoid.sym B p) q }
+            ; snd = λ { {inj₂ _} p q → Setoid.trans B q p }
+            }
+          }
+        ; ⟨_,_⟩ = λ L R → record
```
JacquesCarette, line 61: probably should use pattern-matching lambda here too

src/Categories/Category/Monoidal/Instance/Rels.agda
```diff
@@ -1,204 +1,284 @@
...
+open import Data.Unit.Polymorphic using (⊤)
+open import Function.Base using (case_of_)
+open import Level using (_⊔_; Lift; lift)
+open import Relation.Binary.Bundles using (Setoid)
+
 module _ {o ℓ} where
 
-  Rels-Cartesian : Cartesian (Rels o ℓ)
+  Rels-Cartesian : Cartesian (Rels o (o ⊔ ℓ))
   Rels-Cartesian = record
```
JacquesCarette, line 24: So if anyone is going to use this record somewhere else, it is going to be a problem of sorts. Agda doesn't like huge explicit records. It much prefers records full of things that have local names, i.e. given using a `where` clause. That makes the unifier much happier.

src/Categories/Category/Monoidal/Instance/StrictRels.agda
```diff
@@ -0,0 +1,204 @@
...
+        ; unique = λ { (p , q) (p′ , q′) → (λ { {inj₁ a} r → case (q {a} r) of λ { (inj₁ .a , lift refl , s) → s}
+                                              ; {inj₂ b} r → case (q′ {b} r) of λ { (inj₂ .b , lift refl , s) → s} })
+                                          , λ { {inj₁ a} hxa → p (inj₁ a , lift refl , hxa)
+                                              ; {inj₂ b} hxb → p′ (inj₂ b , lift refl , hxb) } }
+        }
+      }
+    }
+
+  StrictRels-Monoidal : Monoidal (StrictRels o ℓ)
+  StrictRels-Monoidal = monoidalHelper _ record
```
JacquesCarette, line 88: these monster records are probably best broken up in pieces too.

If you don't like `where`, I've also done `private` blocks above the definition where I put everything.

src/Categories/Category/Monoidal/Instance/StrictRels.agda
```diff
@@ -0,0 +1,204 @@
...
+    { ⊗ = record
+      { F₀ = λ { (A , B) → A ×.× B }
```
JacquesCarette, line 90: `uncurry ×.×` ? The reason to see it that way is that there might be room to re-use properties of `×.×` directly, instead of expanding out everything. Yes, all the proofs are trivial, but they are also not 'conceptual'. [This one's more of a suggestion than a must fix]

JacquesCarette (changes requested):

src/Categories/Category/Monoidal/Instance/Rels.agda
Taneb: How huge is huge here? How much should I split the records up?

JacquesCarette: It's not the size of the record, it's the size of the contents. 1-liners are fine, but embedded proofs with > 3 steps should definitely be named instead of inlined.

JacquesCarette:
If you give me permission to push to your fork, I can help push this to completion.

PR #367, opened 2023-03-17
Add right adjoint to Functor.Slice.Forgetful

Taneb:
Based on proposition 1.33 in Freyd's Aspects of Topoi


Taneb:
I feel like both `Forgetful` and `Free` should have better names (especially as `Free` sometimes has a right adjoint of its own), but I don't know what they should be.

JacquesCarette:
Nice. I have no suggestions as to a better name either. But maybe you should put the PR comment as a comment in the file too?

JacquesCarette (approved):

JacquesCarette (approved):
We've waiting long enough, no one has said anything about the name. So let's merge, it can be changed later.

PR #372, opened 2023-04-30
Bump Agda to 2.6.3 and stdlib to 1.7.2

jpoiret:
Hi, everything already type-checks without changes on 2.6.3/1.7.2 locally (although a CI run will increase confidence).
This would also be a good spot to release a new version that is compatible with the above new versions.

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -45,8 +45,8 @@ on:
 ########################################################################
 
 env:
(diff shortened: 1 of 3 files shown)


HuStmpHrrr:
could you maybe change the README as follows:

> The current library works with Agda-2.6.3 and stdlib-1.7.2. 

then we can just do a release.

HuStmpHrrr:
lgtm

HuStmpHrrr (approved):

PR #373, opened 2023-06-29
the naturality conditions for ann and distrib were missing.

JacquesCarette:
So this adds them. And a partial refactoring to use variables.

### DIFF
--- a/src/Categories/Category/RigCategory.agda
+++ b/src/Categories/Category/RigCategory.agda
@@ -17,15 +17,22 @@ open import Categories.NaturalTransformation.NaturalIsomorphism
 open import Categories.NaturalTransformation using (_∘ᵥ_; _∘ₕ_; _∘ˡ_; _∘ʳ_; NaturalTransformation)
 open import Categories.Morphism C
 
+private
+  module C = Category C
+  
+open C hiding (_≈_)
+
+private
+  variable
+    X Y Z W A B : Obj
+    f g h i a b : X ⇒ Y
+
 -- Should probably split out Distributive Category out and make this be 'over' that.
 record RigCategory {M⊎ M× : Monoidal C} (S⊎ : Symmetric M⊎)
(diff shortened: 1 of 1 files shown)


JacquesCarette:
(Asking multiple people for review because this change is blocking me on other work that I'm doing, so I'd like to have this merged sooner rather than later)

src/Categories/Category/RigCategory.agda
```diff
@@ -17,15 +17,22 @@ open import Categories.NaturalTransformation.NaturalIsomorphism
...
+
+private
+  variable
+    X Y Z W A B : Obj
+    f g h i a b : X ⇒ Y
+
 -- Should probably split out Distributive Category out and make this be 'over' that.
 record RigCategory {M⊎ M× : Monoidal C} (S⊎ : Symmetric M⊎)
-   (S× : Symmetric M×) : Set (o ⊔ ℓ ⊔ e) where
+   (S× : Symmetric M×) : Set (o ⊔ ℓ ⊔ e) where
```
Taneb, line 32: Whitespace at end of line?

src/Categories/Category/RigCategory.agda
```diff
@@ -72,6 +79,13 @@ record RigCategory {M⊎ M× : Monoidal C} (S⊎ : Symmetric M⊎)
     ⊕α⇐ = M⊎.associator.to
 
 
+  -- that ann and distrib are natural was missing
```
Taneb, line 82: I don't think this comment should be in the code. Maybe change to something like  `naturality conditions for ann and distrib`

Taneb (changes requested):

src/Categories/Category/RigCategory.agda
JacquesCarette: Fixed

src/Categories/Category/RigCategory.agda
JacquesCarette: Good point - fixed.

Taneb (approved):

PR #377, opened 2023-07-14
Remove broken, unused, AllProducts and IndexedProduct

### LINKED ISSUE #258: Categories.Object.Product.Indexed.AllProducts seems wrong
It's defined as:

```agda
record IndexedProduct {i} (I : Set i) : Set (i ⊔ o ⊔ e ⊔ ℓ) where
  field
    P         : I → Obj
    productOf : IndexedProductOf P

  open IndexedProductOf productOf public

AllProducts : ∀ i → Set (o ⊔ ℓ ⊔ e ⊔ suc i)
AllProducts i = (I : Set i) → IndexedProduct I
```

This is saying that for any index set, there is a product for that indexed set. It says nothing about what objects it could contain. There's nothing stopping it from just having the product over many copies of the same object. In fact, I think that's the only thing that it _can_ have.

Taneb:
This construction was significantly weaker than its name suggested. See https://github.com/agda/agda-categories/issues/258#issuecomment-889119426 for proof that `AllProducts` was implied by `Terminal`. It is also unused.

Closes #258


JacquesCarette (approved):
Given that this is dead code, indeed, best remove it. Thanks.

PR #378, opened 2023-07-15
Cleanup proofs in Kleisli category construction

Taneb:
I'm still not quite happy with the proof of `assoc′`. Open to suggestions there


JacquesCarette (approved):
We're going to have to find better names for those glue combinators... but that can be done later.

PR #379, opened 2023-07-17
small improvements to Morphism.Reasoning

JacquesCarette:
Add more documentation to Categories.Morphism.Reasoning.Core.  Make some module names more reasonable. Simplify some proofs. Fix up downstream code that broke.

### DIFF
--- a/src/Categories/Morphism/Reasoning/Core.agda
+++ b/src/Categories/Morphism/Reasoning/Core.agda
@@ -5,11 +5,16 @@ open import Categories.Category
   Helper routines most often used in reasoning with commutative squares,
   at the level of arrows in categories.
 
-  Basic  : reasoning about identity
+  Identity : reasoning about identity
+  Assoc4   : associativity short-cuts for 4-term
   Pulls  : use a ∘ b ≈ c as left-to-right rewrite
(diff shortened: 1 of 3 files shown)


TOTBWF (approved):
Adding explicit parens to the type signatures is a good move!

src/Categories/Morphism/Reasoning/Core.agda
```diff
@@ -5,11 +5,16 @@ open import Categories.Category
   Helper routines most often used in reasoning with commutative squares,
   at the level of arrows in categories.
 
-  Basic  : reasoning about identity
+  Identity : reasoning about identity
+  Assoc4   : associativity short-cuts for 4-term
```
Taneb, line 9: I'm not sure this line makes grammatical sense

Taneb (approved):
This is definitely a great improvement

src/Categories/Morphism/Reasoning/Core.agda
JacquesCarette: You are right -- will fix.

JacquesCarette:
@TOTBWF I have no idea what mechanism produces these "dismissed stale review", I find them rather rude, and it wasn't my explicit doing!

PR #380, opened 2023-07-25
Improve level polymorphism in one place.

JacquesCarette:
@xplat noticed that the product construction could have more general Levels. Also use variables a bit more while at it.


Taneb (approved):
Looks good!

PR #381, opened 2023-07-27
add kleisli triple

Reijix:
Small proof that kleisli triples are equivalent to monads. I'm looking for some feedback and then I'd do the same for comonads.

I've defined the kleisli-triple as a typedef for `RMonad idF` to avoid redundancy, but this might make things a little unclear.

### DIFF
--- a/src/Categories/Monad/Construction/Kleisli.agda
+++ b/src/Categories/Monad/Construction/Kleisli.agda
@@ -1,3 +1,5 @@
+{-# OPTIONS --without-K --safe #-}
+
 module Categories.Monad.Construction.Kleisli where
 
 -- Definition of kleisli triple as a relative monad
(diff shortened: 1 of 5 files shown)


src/Categories/Monad/Construction/Kleisli.agda
```diff
@@ -0,0 +1,116 @@
+module Categories.Monad.Construction.Kleisli where
+
+-- Definition of kleisli triple as a relative monad
+-- and the equivalence of kleisli triple and monad
+-- see https://ncatlab.org/nlab/show/extension+system (the naming differs)
+
+open import Level
+open import Categories.Category
```
JacquesCarette, line 8: Can you please use `Categories.Category.Core` (unless you really need more) and also `using` clauses and and for all `open import` below.

src/Categories/Monad/Construction/Kleisli.agda
```diff
@@ -0,0 +1,116 @@
...
+open import Categories.Monad.Relative renaming (Monad to RMonad)
+open import Categories.Monad
+open import Categories.Functor renaming (id to idF)
+open import Categories.NaturalTransformation renaming (id to idN)
+import Categories.Morphism.Reasoning as MR
+
+private
```
JacquesCarette, line 15: Here and below: just 2 space indent please. Otherwise things get too wide too fast.

src/Categories/Monad/Construction/Kleisli.agda
```diff
@@ -0,0 +1,116 @@
...
+        o ℓ e : Level
+
+module _ (C : Category o ℓ e) where
+    open Category C
+    open HomReasoning
+    open Equiv
+    open MR C
+
+    -- a kleisli triple is a relative monad with J = idF
+    Kleisli : Set (o ⊔ ℓ ⊔ e)
```
JacquesCarette, line 26: I'm wondering if the type should be called `KleisliTriple` as there are too many `Kleisli` things already? (The implication below can keep its name)

src/Categories/Monad/Construction/Kleisli.agda
```diff
@@ -0,0 +1,116 @@
...
+                    extend (id ∘ extend (unit ∘ f))                 ≈⟨ K.extend-≈ id-comm-sym ⟩
+                    extend (extend (unit ∘ f) ∘ id)                 ≈⟨ K.assoc ⟩
+                    extend (unit ∘ f) ∘ (extend id)                 ∎
+                }
+
+            module μ = NaturalTransformation μ
+            module η = NaturalTransformation η
+
+            assoc' : ∀ {X : Obj} → μ.η X ∘ F₁ (μ.η X) ≈ μ.η X ∘ μ.η (F₀ X)
+            assoc' = begin
```
JacquesCarette, line 70: This is the exact same proof as `commute` above -- factor out please.

src/Categories/Monad/Construction/Kleisli.agda
```diff
@@ -0,0 +1,116 @@
...
+            η = ntHelper {F = idF} {G = F} record
+                { η = λ X → unit
+                ; commute = λ f → sym K.identityʳ
+                }
+
+            -- constructing μ from extend
+            μ = ntHelper {F = F ∘F F} {G = F} record
+                { η = λ X → extend id
+                ; commute = λ f → begin 
+                    extend id ∘ extend (unit ∘ extend (unit ∘ f))   ≈⟨ sym K.assoc ⟩
```
JacquesCarette, line 59: This repeated use of `sym K.assoc` makes me think that `RelativeMonad` should be extended to have `sym-assoc` in its API (this is optional, but would be nice).

src/Categories/Monad/Construction/Kleisli.agda
```diff
@@ -0,0 +1,116 @@
...
+        }
+        where
+            module M = Monad M
+            open Monad M using (F; η; μ)
+            open Functor F
+            
+            identityʳ' : ∀ {X Y} {k : X ⇒ F₀ Y} → (μ.η Y ∘ F₁ k) ∘ η.η X ≈ k
+            identityʳ' {X} {Y} {k}  = begin 
+                ((μ.η Y ∘ F₁ k) ∘ η.η X)   ≈⟨ pullʳ (sym (η.commute _)) ⟩
+                (μ.η Y ∘ η.η (F₀ Y) ∘ k)   ≈⟨ pullˡ M.identityʳ ⟩
```
JacquesCarette, line 103: `cancelˡ M.identityʳ` should cut out 1 line.

JacquesCarette (changes requested):
Very nice - just a few minor things to tweak.

src/Categories/Monad/Construction/Kleisli.agda
sergey-goncharov: As a slightly polemic point, nLab seems to promote the name "extension system" https://ncatlab.org/nlab/show/extension+system. I am more used to the name "Kleisli triple", but nLab may well change the trend. I do not really have any opinion on this.

JacquesCarette: I would be fine with either `KleisliTriple` or `ExtensionSystem`.

src/Categories/Monad/Construction/Kleisli.agda
Reijix: Sure, fixed it!

src/Categories/Monad/Construction/Kleisli.agda
Reijix: Done

src/Categories/Monad/Construction/Kleisli.agda
Reijix: I chose `KleisliTriple` and `KleisliCotriple` for now.

src/Categories/Monad/Construction/Kleisli.agda
Reijix: Changed it for `RelativeMonad` and `RelativeComonad` and also fixed places where its used.

Reijix:
Thanks for the feedback! I've also added `Categories.Comonad.Construction.CoKleisli` now.

JacquesCarette (approved):

PR #382, opened 2023-08-02
Displayed categories

Taneb:
- [x] Displayed categories
- [x] Total categories
- [ ] Displayed setoid reasoning and displayed morphism reasoning
- [ ] Displayed functors (defined locally but would be improved by displayed morphism reasoning)
- [ ] Displayed natural transformations
- [ ] Investigate displayed category of displayed categories#
- [ ] Vertical displayed functors
- [ ] Vertical displayed natural transformations?
- [ ] Category of displayed categories over a fixed base

### DIFF
--- a/src/Categories/Category/Construction/Total.agda
+++ b/src/Categories/Category/Construction/Total.agda
@@ -1,0 +1,55 @@
+{-# OPTIONS --safe --without-K #-}
+
+open import Categories.Category.Core using (Category)
+open import Categories.Category.Displayed using (Displayed)
+
+module Categories.Category.Construction.Total {o ℓ e o′ ℓ′ e′} {B : Category o ℓ e} (C : Displayed B o′ ℓ′ e′) where
+
+open import Data.Product using (proj₁; proj₂; Σ-syntax; ∃-syntax; -,_)
+open import Level using (_⊔_)
+
+open import Categories.Functor.Core using (Functor)
+
+open Category B
+open Displayed C
+open Equiv′
+
+Total : Set (o ⊔ o′)
+Total = Σ[ Carrier ∈ Obj ] Obj[ Carrier ]
+
+record Total⇒ (X Y : Total) : Set (ℓ ⊔ ℓ′) where
+  constructor total⇒
+  field
+    {arr} : proj₁ X ⇒ proj₁ Y
+    preserves : proj₂ X ⇒[ arr ] proj₂ Y
+
+open Total⇒ public using (preserves)
+
+∫ : Category (o ⊔ o′) (ℓ ⊔ ℓ′) (e ⊔ e′)
+∫ = record
+  { Obj = Total
+  ; _⇒_ = Total⇒
+  ; _≈_ = λ f g → ∃[ p ] preserves f ≈[ p ] preserves g
+  ; id = total⇒ id′
+  ; _∘_ = λ f g → total⇒ (preserves f ∘′ preserves g)
+  ; assoc = -, assoc′
+  ; sym-assoc = -, sym-assoc′
+  ; identityˡ = -, identityˡ′
+  ; identityʳ = -, identityʳ′
+  ; identity² = -, identity²′
+  ; equiv = record
+    { refl = -, refl′
+    ; sym = λ p → -, sym′ (proj₂ p)
+    ; trans = λ p q → -, trans′ (proj₂ p) (proj₂ q)
+    }
+  ; ∘-resp-≈ = λ p q → -, ∘′-resp-[]≈ (proj₂ p) (proj₂ q)
+  }
+
+display : Functor ∫ B
+display = record
+  { F₀ = proj₁
+  ; F₁ = Total⇒.arr
+  ; identity = Equiv.refl
+  ; homomorphism = Equiv.refl
+  ; F-resp-≈ = proj₁
+  }
--- a/src/Categories/Category/Displayed.agda
+++ b/src/Categories/Category/Displayed.agda
@@ -1,0 +1,67 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Categories.Category
+
+module Categories.Category.Displayed {o ℓ e} (B : Category o ℓ e) where
+
+open import Data.Product
+open import Level
+
+open import Categories.Functor.Core
+open import Relation.Binary.Displayed
+
+open Category B
+open Equiv
+
+-- A displayed category captures the idea of placing extra structure
+-- over a base category. For example, the category of monoids can be
+-- considered as the category of setoids with extra structure on the
+-- objects and extra conditions on the morphisms.
+record Displayed o′ ℓ′ e′ : Set (o ⊔ ℓ ⊔ e ⊔ suc (o′ ⊔ ℓ′ ⊔ e′)) where
+  infix 4 _⇒[_]_ _≈[_]_
+  infixr 9 _∘′_
+  field
+    Obj[_] : Obj → Set o′
+    _⇒[_]_ : ∀ {X Y} → Obj[ X ] → X ⇒ Y → Obj[ Y ] → Set ℓ′
+    _≈[_]_ : ∀ {A B X Y} {f g : A ⇒ B} → X ⇒[ f ] Y → f ≈ g → X ⇒[ g ] Y → Set e′
+
+    id′ : ∀ {A} {X : Obj[ A ]} → X ⇒[ id ] X
+    _∘′_ : ∀ {A B C X Y Z} {f : B ⇒ C} {g : A ⇒ B}
+         → Y ⇒[ f ] Z → X ⇒[ g ] Y → X ⇒[ f ∘ g ] Z
+
+    identityʳ′ : ∀ {A B X Y} {f : A ⇒ B} → {f′ : X ⇒[ f ] Y} → f′ ∘′ id′ ≈[ identityʳ ] f′
+    identityˡ′ : ∀ {A B X Y} {f : A ⇒ B} → {f′ : X ⇒[ f ] Y} → id′ ∘′ f′ ≈[ identityˡ ] f′
+    identity²′ : ∀ {A} {X : Obj[ A ]} → id′ {X = X} ∘′ id′ ≈[ identity² ] id′
+    assoc′ : ∀ {A B C D W X Y Z} {f : C ⇒ D} {g : B ⇒ C} {h : A ⇒ B}
+           → {f′ : Y ⇒[ f ] Z} → {g′ : X ⇒[ g ] Y} → {h′ : W ⇒[ h ] X}
+           → (f′ ∘′ g′) ∘′ h′ ≈[ assoc ] f′ ∘′ (g′ ∘′ h′)
+    sym-assoc′ : ∀ {A B C D W X Y Z} {f : C ⇒ D} {g : B ⇒ C} {h : A ⇒ B}
+           → {f′ : Y ⇒[ f ] Z} → {g′ : X ⇒[ g ] Y} → {h′ : W ⇒[ h ] X}
+           → f′ ∘′ (g′ ∘′ h′) ≈[ sym-assoc ] (f′ ∘′ g′) ∘′ h′
+
+
+    equiv′ : ∀ {A B X Y} → IsDisplayedEquivalence equiv (_≈[_]_ {A} {B} {X} {Y})
+
+    ∘′-resp-[]≈ : ∀ {A B C X Y Z} {f h : B ⇒ C} {g i : A ⇒ B}
+                    {f′ : Y ⇒[ f ] Z} {h′ : Y ⇒[ h ] Z} {g′ : X ⇒[ g ] Y} {i′ : X ⇒[ i ] Y}
+                    {p : f ≈ h} {q : g ≈ i}
+                → f′ ≈[ p ] h′ → g′ ≈[ q ] i′ → f′ ∘′ g′ ≈[ ∘-resp-≈ p q ] h′ ∘′ i′
+
+  module Equiv′ {A B X Y} = IsDisplayedEquivalence (equiv′ {A} {B} {X} {Y})
+
+  open Equiv′
+
+  ∘′-resp-[]≈ˡ : ∀ {A B C X Y Z} {f h : B ⇒ C} {g : A ⇒ B} {f′ : Y ⇒[ f ] Z} {h′ : Y ⇒[ h ] Z} {g′ : X ⇒[ g ] Y}
+                 → {p : f ≈ h} → f′ ≈[ p ] h′ → f′ ∘′ g′ ≈[ ∘-resp-≈ˡ p ] h′ ∘′ g′
+  ∘′-resp-[]≈ˡ pf = ∘′-resp-[]≈ pf refl′
+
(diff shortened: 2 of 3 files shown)


JacquesCarette:
I've done some work already based on your gist... see my [Categorical Playground repo](https://github.com/JacquesCarette/Categorical-Playground/) where your code is in Categorical/Displayed and mine in 2Level/Category.

Note that I'm not suggesting that my approach in 2Level/Category is 'right', it is very much an experiment that I'm in the middle of. But there might be things in there that could trigger an idea for you.

src/Categories/Category/Displayed.agda
```diff
@@ -0,0 +1,67 @@
...
+
+  ∘′-resp-[]≈ˡ : ∀ {A B C X Y Z} {f h : B ⇒ C} {g : A ⇒ B} {f′ : Y ⇒[ f ] Z} {h′ : Y ⇒[ h ] Z} {g′ : X ⇒[ g ] Y}
+                 → {p : f ≈ h} → f′ ≈[ p ] h′ → f′ ∘′ g′ ≈[ ∘-resp-≈ˡ p ] h′ ∘′ g′
+  ∘′-resp-[]≈ˡ pf = ∘′-resp-[]≈ pf refl′
+
+  ∘′-resp-[]≈ʳ : ∀ {A B C X Y Z} {f : B ⇒ C} {g i : A ⇒ B} {f′ : Y ⇒[ f ] Z} {g′ : X ⇒[ g ] Y} {i′ : X ⇒[ i ] Y}
+                 → {p : g ≈ i} → g′ ≈[ p ] i′ → f′ ∘′ g′ ≈[ ∘-resp-≈ʳ p ] f′ ∘′ i′
+  ∘′-resp-[]≈ʳ pf = ∘′-resp-[]≈ refl′ pf
+
+  hom-setoid′ : ∀ {A B} {X : Obj[ A ]} {Y : Obj[ B ]} → DisplayedSetoid hom-setoid _ _
```
JacquesCarette, line 62: I wouldn't define `hom-setoid'` inside `Displayed`. Turns out it is rarely used.

Taneb: I was planning to use it to pass as an argument to a module for morphism reasoning syntax (similar to how `hom-reasoning` is used in `Category)

JacquesCarette: Ah well, in that case, it should stay!

src/Categories/Category/Construction/Total.agda
```diff
@@ -0,0 +1,55 @@
...
+  ; identity² = -, identity²′
+  ; equiv = record
+    { refl = -, refl′
+    ; sym = λ p → -, sym′ (proj₂ p)
+    ; trans = λ p q → -, trans′ (proj₂ p) (proj₂ q)
+    }
+  ; ∘-resp-≈ = λ p q → -, ∘′-resp-[]≈ (proj₂ p) (proj₂ q)
+  }
+
+display : Functor ∫ B
```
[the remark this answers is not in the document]
JacquesCarette: Part of my doodling is exploring how much `Displayed` is really the categorical equivalent of a `Sigma` type. So it's much more than an implementation detail, there is a very strong analogy at play.

I'm sure there are other, better names (certainly than `proj1`). I'm not set against `display`, BTW, but I do think it's worth thinking about.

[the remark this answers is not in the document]
TOTBWF: This functor is normally called `π` in the literature, so `proj1` is a fitting name.

src/Categories/Functor/Displayed.agda
```diff
@@ -0,0 +1,68 @@
...
+  { F₀′ = id→
+  ; F₁′ = id→
+  ; identity′ = Displayed.Equiv′.refl′ C
+  ; homomorphism′ = Displayed.Equiv′.refl′ C
+  ; F′-resp-≈[] = id→
+  }
+
+_∘F′_ : ∀ {C : Displayed B₁ o ℓ e} {D : Displayed B₂ o′ ℓ′ e′} {E : Displayed B₃ o″ ℓ″ e″}
+          {F : Functor B₂ B₃} {G : Functor B₁ B₂}
+        → DisplayedFunctor D E F → DisplayedFunctor C D G → DisplayedFunctor C E (F ∘F G)
```
Taneb, line 48: If you include the ones hidden by variables, this has 18 `Level` parameters! Is the most in the library (so far)?

src/Categories/Functor/Displayed.agda
```diff
@@ -0,0 +1,68 @@
...
+_∘F′_  {C = C} {D = D} {E = E} F′ G′ = record
+   { F₀′ = F′.₀′ ∙ G′.₀′
+   ; F₁′ = F′.₁′ ∙ G′.₁′
+   ; identity′ = begin′
+     F′.₁′ (G′.₁′ C.id′) ≈[]⟨ F′.F′-resp-≈[] G′.identity′ ⟩
+     F′.₁′ D.id′         ≈[]⟨ F′.identity′ ⟩
+     E.id′               ∎′
+   ; homomorphism′ = λ {_} {_} {_} {_} {_} {_} {_} {_} {f′} {g′} → begin′
```
Taneb, line 56: Is there a good way to avoid having to write quite this many `{_}`s to get two of the hidden arguments in scope?

src/Categories/Functor/Displayed.agda
JacquesCarette: That is definitely, by a long shot, more than I've seen in any library.

src/Categories/Morphism/Reasoning/Core.agda
```diff
@@ -171,11 +171,11 @@ glue {c′ = c′} {a′ = a′} {a = a} {c″ = c″} {c = c} {b′ = b′} {b
   a ∘ (c′ ∘ b′)  ≈⟨ extendʳ sq-a ⟩
   c″ ∘ (a′ ∘ b′) ∎
 
--- A "rotated" version of glue′. Equivalent to 'Equiv.sym (glue (Equiv.sym sq-a) (Equiv.sym sq-b))'
-glue′ : CommutativeSquare a′ c′ c″ a →
-        CommutativeSquare b′ c c′ b →
-        CommutativeSquare (a′ ∘ b′) c c″ (a ∘ b)
-glue′ {a′ = a′} {c′ = c′} {c″ = c″} {a = a} {b′ = b′} {c = c} {b = b} sq-a sq-b = begin
+-- A "rotated" version of glue. Equivalent to 'Equiv.sym (glue (Equiv.sym sq-a) (Equiv.sym sq-b))'
+sym-glue : CommutativeSquare a′ c′ c″ a →
```
Taneb, line 175: This name is not final! I just wanted to use the name `glue′` for the displayed version of `glue` so had to name this something else

Taneb:
One thing I've noticed doing this, I care a lot more about the _exact_ proof that two morphisms are equal than I normally would. This meant it matters, and is a little unfortunate, that equational reasoning syntax has an extraneous `trans refl` at the end. I don't know if this is fixable but it's definitely possible to work around it.

I've also been trying to make sure that little things that I'm building on top of have the "right" proofs ("right" is very subjective here). For example, defining some things in `Categories.NaturalTransformation` so that they don't use `sym` as much when they could use `sym-assoc` instead.

src/Categories/Functor/Displayed.agda
JacquesCarette: That is different, it turns out. Sometimes this version isn't always accepted by Agda, I've never quite figured out why / when.

TOTBWF:
Yeah, I was somewhat worried about this. In the 1Lab we define (pre)categories to have an hSet of morphisms, which means that we can ignore any sort of coherence issues involved with equalities of equalities. In general, `agda-categories` is really working with the theory of wild (2,1)-categories; normally we just perform a sneaky version of truncation by dropping equality proofs, but this is going to cause some pretty serious problems as this work continues.

As a concrete example, [cartesian morphisms](https://1lab.dev/Cat.Displayed.Cartesian.html) are going to be really problematic. The universal property requires that a morphism be displayed over a composite, which often involves some amount of transport. This comes up when defining the factorization system on a fibration; you end up using the fact that every morphism `f' : Hom[ f ] x y` is also displayed over `f . id`.

src/Categories/NaturalTransformation/NaturalIsomorphism/Displayed.agda
```diff
@@ -0,0 +1,159 @@
...
+        → G′ ≃[ G≃H ] H′ → F′ ≃[ F≃G ] G′ → F′ ≃[ G≃H ⓘᵥ F≃G ] H′
+_ⓘᵥ′_ {D′ = D′} G′≃H′ F′≃G′ = record
+  { F′⇒G′ = F′⇒G′ G′≃H′ ∘′ᵥ F′⇒G′ F′≃G′
+  ; F′⇐G′ = F′⇐G′ F′≃G′ ∘′ᵥ F′⇐G′ G′≃H′
+  ; iso′ = λ X → DisplayedIso-∘′ (iso′ F′≃G′ X) (iso′ G′≃H′ X)
+  }
+  where
+    open DMorphismₚ D′
+
+_ⓘₕ′_ : ∀ {C : Category bo bℓ be} {D : Category bo′ bℓ′ be′} {E : Category bo″ bℓ″ be″}
```
Taneb, line 64: Agda struggled to use the variables declared in the `variable` block here, causing me to add the category, functor, and displayed category arguments explicitly. Not sure why.

src/Categories/NaturalTransformation/NaturalIsomorphism/Displayed.agda
```diff
@@ -0,0 +1,159 @@
...
+  unitorˡ′ = record { F′⇒G′ = id′∘F′⇒F′ ; F′⇐G′ = F′⇒id′∘F′ ; iso′ = iso-id′-id′ }
+
+  unitorʳ′ : F′ ∘F′ idF′ ≃[ unitorʳ ] F′
+  unitorʳ′ = record { F′⇒G′ = F′∘id′⇒F′ ; F′⇐G′ = F′⇒F′∘id′ ; iso′ = iso-id′-id′ }
+
+unitor²′ : ∀ {C′ : Displayed C o ℓ e} → idF′ ∘F′ idF′ ≃[ unitor² ] idF′ {C = C′}
+unitor²′ = record { F′⇒G′ = id′∘id′⇒id′ ; F′⇐G′ = id′⇒id′∘id′ ; iso′ = LeftRightId′.iso-id′-id′ idF′ }
+
+-- associator
+module _
```
Taneb, line 131: I've beaten my own record here, this module has 24 different `Level`s as arguments.

src/Categories/Category/Displayed/Instance/DisplayedCats.agda
```diff
@@ -0,0 +1,27 @@
...
+DisplayedCats o ℓ e o′ ℓ′ e′ = record
+  { Obj[_] = λ B → Displayed B o′ ℓ′ e′
+  ; _⇒[_]_ = λ C F D → DisplayedFunctor C D F
+  ; _≈[_]_ = λ F′ F≃G G′ → DisplayedNaturalIsomorphism F′ G′ F≃G
+  ; id′ = id′
+  ; _∘′_ = _∘F′_
+  ; identityˡ′ = unitorˡ′
+  ; identityʳ′ = unitorʳ′
+  ; identity²′ = unitor²′
+  ; assoc′ = λ {_ _ _ _ _ _ _ _ _ _ _ F′ G′ H′} → associator′ H′ G′ F′
```
Taneb, line 23: To revisit later: why do these arguments appear in the wrong order?

src/Categories/NaturalTransformation/NaturalIsomorphism/Displayed.agda
JacquesCarette: Impressive indeed. And a little crazy.

Taneb: I suspect in most use cases, until we come up with a notion of "category where the objects can vary in level", this will collapse down to six (three for the base categories and three more for the displayed categories)

Taneb:
Composition of vertical functors is the first place I've hit where I've not been able to see a way around that we may have lots of different equalities between two functions. In particular, `identity′` wants a `F′.₁′ (G′.₁′ C.id′) E.≈[ refl ] E.id′` but I can only see how to get a ``F′.₁′ (G′.₁′ C.id′) E.≈[ trans refl refl ] E.id′``.

JacquesCarette:
Two things come to mind when I see this problem over things "over" equalities:
1. you're not being weak enough somewhere to that you are demanding a very particular proof instead of any proof 
2. things are not strict enough, i.e. `trans` doesn't compute on `refl`.

Obviously these two things pull in polar opposite directions! I'd like to see more details as I'm stumbling in the dark a bit here. (And I'm super interested, as I think the same problem may well surface for both Multicategory and Double category.)

Influenced by the above, I would speculate that there are two needs for these proofs: we need them to be there so that things 'line up' properly, and we need them to 'properly exist' for well-definedness.

TOTBWF:
Another perspective on this is that E-categories are sort of like half of a bicategory: we've got the 2-dimensional data of a bicategory, yet none of the coherences.

Also, I'm pretty convinced that we need more than a type-family of morphisms: we need a *setoid*-family of morphisms!

JacquesCarette:
I agree that E-categories are very much like bicategories with some of the coherences missing. What's so remarkable is how pretty much all of 1-category theory goes through!

Yes, some amount of transport is *one* way out of this jam. I'm quite curious to see if there are other ways around the problem. [There might not be - but I also thought that the current approach to agda-categories would fail miserably in the first 2 weeks of us porting the old code, and yet here we are...]

TOTBWF:
I'm very interested in this as well; even in settings where you do have transport, it tends to be pretty miserable in practice.
(This is especially true in cubical, where transport does _not_ compute on refl)

I see two ways forward on this: the first involves some co-yoneda trickery; instead of having displayed morphisms living _directly_ over some `f`, you define a type like

```agda
record Over (f : Hom x y) (x' : Ob[ x ]) (y : Ob[ y ]) : Type where
  field
    base : Hom x y
    over : f ≈ base
    hom : Hom[ base ] x' y'
```

and then only work in terms of `Over`. It's pretty heavy-weight, but should work.

The other approach is to treat things like having fibre categories as honest-to-god structure. This is pretty interesting in it's own right, though I'll admit I haven't thought too deeply on it.

TOTBWF:
As another addendum: there is a second perspective on E-categories, which is that they are [wild categories](https://ncatlab.org/nlab/show/wild%20category). FWIW, this makes `agda-categories` even more miraculous; it's really surprising that you can keep truncating and things keep working out!

JacquesCarette:
I prefer the "maximally unsaturated" point of view of Peter LeFanu Lumsdaine and Erik Palmgren (i.e. I think the concept is Palmgren's but I learned it from Peter).  Unfortunately the slides seem to have disappeared from the Internet per se, but still available in various archives / caches. But I think the video  https://youtu.be/glC8tC7xdBY is still available.

But on `Over`: I agree that if `f ≈ g` then Hom[ f ] x' y' and Hom[ g ] x' y' should be equivalent (as types). And this dependency should not itself depend on the choice of proof of `f ≈ g` .

PR #383, opened 2023-08-04
Distributive Category

Reijix:
Distributive categories as described [here](https://ncatlab.org/nlab/show/distributive+category) and their relationship to extensive categories (an extensive and cartesian category is automatically distributive).

I needed some properties of coproducts (namely `up-to-iso` and `Mobile`) so I added them, but I was wondering if they have been left out on purpose and I should've used Duality?

Lastly I've also added bundled versions `ExtensiveCategory` and `ExtensiveDistributiveCategory`.

### DIFF
--- a/src/Categories/Category/Distributive.agda
+++ b/src/Categories/Category/Distributive.agda
@@ -1,10 +1,10 @@
 {-# OPTIONS --without-K --safe #-}
 
 open import Level
-open import Categories.Category
-open import Categories.Category.Cartesian
-open import Categories.Category.BinaryProducts
-open import Categories.Category.Cocartesian
+open import Categories.Category.Core
+open import Categories.Category.Cartesian using (Cartesian)
+open import Categories.Category.BinaryProducts using (BinaryProducts)
+open import Categories.Category.Cocartesian using (Cocartesian)
 import Categories.Morphism as M
 import Categories.Morphism.Reasoning as MR
 
--- a/src/Categories/Category/Distributive.agda
+++ b/src/Categories/Category/Distributive.agda
@@ -1,0 +1,65 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Level
+open import Categories.Category
+open import Categories.Category.Cartesian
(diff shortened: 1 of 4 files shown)


sergey-goncharov:
I suggest refactoring this, by using `IsCoproduct` than working with two coproducts and involving explicit up-to-isomorphism reasoning. To facilitate this, I've added a lemma `pullback-of-cp-is-cp'` to `Categories.Category.Extensive` -- the relevant PR https://github.com/agda/agda-categories/pull/384 is pending.

Reijix:
> I suggest refactoring this, by using IsCoproduct than working with two coproducts and involving explicit up-to-isomorphism reasoning. To facilitate this, I've added a lemma pullback-of-cp-is-cp' to Categories.Category.Extensive -- the relevant PR https://github.com/agda/agda-categories/pull/384 is pending.

Great, I'll adjust the proof once it is finished. This will shorten `distrib` and make `distrib-canon` obsolete.
Having both definitions is valuable I think, since `pullback-of-cp-is-cp` seems to be easier to proof, while `pullback-of-cp-is-cp'` is easier to work with.

JacquesCarette:
I've asked for some changes to #384 , but the interface will stay the same. So you could `postulate` it while waiting.

sergey-goncharov:
> Having both definitions is valuable I think, since `pullback-of-cp-is-cp` seems to be easier to proof, while `pullback-of-cp-is-cp'` is easier to work with.

A small didactic point: only `pullback-of-cp-is-cp` is (part of the) definition, `pullback-of-cp-is-cp'` is a (theorem or) lemma, so a derivable statement. **Definitions** must possibly be designed so as to be as restrictive as possible (because we will have to check them for every instance), while derived claims must be as rich as possible (because, they are the generic facts that we obtain from the definitions "for free").

JacquesCarette:
#384 has now been merged.

Reijix:
Updated it now.

JacquesCarette (approved):

PR #384, opened 2023-08-07
Update Extensive.agda

sergey-goncharov:
added pullback-of-cp-is-cp' to Categories.Category.Extensive; cleaned up the file a bit

### DIFF
--- a/src/Categories/Category/Extensive.agda
+++ b/src/Categories/Category/Extensive.agda
@@ -37,14 +37,15 @@ record Extensive {o ℓ e : Level} (𝒞 : Category o ℓ e) : Set (suc (o ⊔ �
 
   -- a version with non-canonical pullbacks
   module _ {A B C : Obj} {f : A ⇒ B + C} (pb₁ : Pullback 𝒞 f i₁) (pb₂ : Pullback 𝒞 f i₂) where
-      open IsCoproduct (pullback-of-cp-is-cp f)
-      open HomReasoning; open MR 𝒞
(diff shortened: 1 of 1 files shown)


src/Categories/Category/Extensive.agda
```diff
@@ -32,13 +35,56 @@ record Extensive {o ℓ e : Level} (𝒞 : Category o ℓ e) : Set (suc (o ⊔ 
...
-
-
-
-
-
-
-
-
-
+  -- a version with non-canonical pullbacks
```
JacquesCarette, line 38: I think you can make the following shorter by instead using a local module, i.e.

```agda
module _ {A B C : Obj} {f : A ⇒ B + C} (pb₁ : Pullback 𝒞 f i₁) (pb₂ : Pullback 𝒞 f i₂) where
```
and then do a few common `open` at the top.  You could also share some definitions in a `private` block that way.

JacquesCarette (changes requested):

src/Categories/Category/Extensive.agda
sergey-goncharov: Indeed, thanks! Done the `module _` thing. Not sure about the `private` part.

sergey-goncharov: Did you mean to make the declarations before `pullback-of-cp-is-cp'` private to the module?

JacquesCarette (approved):

PR #389, opened 2023-08-21
Category of endofunctors is monoidal

iwilare:
We couldn't find an instance of Monoidal for the category of endofunctors [C,C] for a category C, so we wrote one!

### DIFF
--- a/src/Categories/Category/Monoidal/Instance/Endofunctors.agda
+++ b/src/Categories/Category/Monoidal/Instance/Endofunctors.agda
@@ -1,0 +1,131 @@
+{-# OPTIONS --without-K --safe #-}
+module Categories.Category.Monoidal.Instance.Endofunctors {o ℓ e} where
+
+-- Given a category C, the category of endofunctors on C is a monoidal category.
+
+open import Data.Empty.Polymorphic using (⊥; ⊥-elim)
+open import Data.Product using (_,_; uncurry′)
+open import Data.Sum using (_⊎_; inj₁; inj₂; [_,_]′)
+open import Data.Unit.Polymorphic using (⊤)
+open import Function using (case_of_; flip)
+open import Level using (Lift; lift; _⊔_)
+open import Relation.Binary.PropositionalEquality using (_≡_; refl)
+
+open import Categories.Morphism using (_≅_; Iso)
+open import Categories.Category.Cartesian using (Cartesian)
+open import Categories.Category.Cartesian.Monoidal using (module CartesianMonoidal)
+open import Categories.Category.Core using (Category)
+open import Categories.Category.Cocartesian using (Cocartesian)
+open import Categories.Category.Construction.Functors using (Functors)
+open import Categories.Functor using (Functor; _∘F_) renaming (id to idF)
+open import Categories.NaturalTransformation using (NaturalTransformation; _∘ₕ_)
+open import Categories.Category.Monoidal using (Monoidal; monoidalHelper)
+open import Categories.Category.Monoidal.Symmetric using (Symmetric; symmetricHelper)
+open import Categories.NaturalTransformation.NaturalIsomorphism using (niHelper)
+
+open import Categories.NaturalTransformation.NaturalIsomorphism
+  using (NaturalIsomorphism; unitorˡ; unitorʳ; associator)
+
+import Categories.Morphism.Reasoning as MR
+
+Endofunctors : Category o ℓ e → Category (o ⊔ ℓ ⊔ e) (o ⊔ ℓ ⊔ e) (o ⊔ e)
+Endofunctors C = Functors C C
+
+module _ where
+  -- Conversion function from natural isomorphisms to isomorphisms in the functor category.
+  ≃→≅ : ∀ {C D : Category o ℓ e}
+      → {F G : Functor C D}
+      → NaturalIsomorphism F G
+      → Categories.Morphism._≅_ (Functors C D) F G
+  ≃→≅ i = record
+    { from = i.F⇒G
+    ; to = i.F⇐G
+    ; iso = record { isoˡ = λ {x} → Iso.isoˡ (i.iso x) ; isoʳ = λ {x} → Iso.isoʳ (i.iso x) }
+    } where module i = NaturalIsomorphism i
+
+module _ where
+  Endofunctors-Monoidal : ∀ {C : Category o ℓ e} → Monoidal (Endofunctors C)
+  Endofunctors-Monoidal {C} = record
+    { ⊗ = record
+      { F₀ = uncurry′ _∘F_
+      ; F₁ = uncurry′ _∘ₕ_
+      ; identity = λ { {F , G} →
+        let module F = Functor F
+            module G = Functor G in
+          begin F.F₁ id ∘ id ≈⟨ F.identity ⟩∘⟨refl ⟩
+                id ∘ id ≈⟨ identity² ⟩
+                id ∎ }
+      ; homomorphism = λ { {F , G} {F₂ , G₂} {F₃ , G₃} {f₁ , f₂} {g₁ , g₂} →
+        let module F₃ = Functor F₃
+            module F₂ = Functor F₂
+            module G = Functor G
+            module G₂ = Functor G₂
+            module f₁ = NaturalTransformation f₁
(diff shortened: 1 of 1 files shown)


src/Categories/Category/Monoidal/Instance/Endofunctors.agda
```diff
@@ -0,0 +1,131 @@
...
+  -- Conversion function from natural isomorphisms to isomorphisms in the functor category.
+  ≃→≅ : ∀ {C D : Category o ℓ e}
+      → {F G : Functor C D}
+      → NaturalIsomorphism F G
+      → Categories.Morphism._≅_ (Functors C D) F G
+  ≃→≅ i = record
+    { from = i.F⇒G
+    ; to = i.F⇐G
+    ; iso = record { isoˡ = λ {x} → Iso.isoˡ (i.iso x) ; isoʳ = λ {x} → Iso.isoʳ (i.iso x) }
+    } where module i = NaturalIsomorphism i
```
sstucki, line 44: I think this is the same as the conversions in [Categories.NaturalTransformation.NaturalIsomorphism.Functors](https://github.com/agda/agda-categories/blob/master/src/Categories/NaturalTransformation/NaturalIsomorphism/Functors.agda).

src/Categories/Category/Monoidal/Instance/Endofunctors.agda
```diff
@@ -0,0 +1,131 @@
...
+            module G₂ = Functor G₂
+            module f₁ = NaturalTransformation f₁
+            module f₂ = NaturalTransformation f₂
+            module g₁ = NaturalTransformation g₁
+            module g₂ = NaturalTransformation g₂ in
+          begin
+            F₃.₁ (g₂.η _ ∘ f₂.η _) ∘ g₁.η (G.₀ _) ∘ f₁.η (G.₀ _) ≈⟨ pushˡ F₃.homomorphism ⟩
+            F₃.₁ (g₂.η _) ∘ F₃.₁ (f₂.η _) ∘ g₁.η (G.₀ _) ∘ f₁.η (G.₀ _) ≈⟨ refl⟩∘⟨ extendʳ (g₁.sym-commute _) ⟩
+            F₃.₁ (g₂.η _) ∘ g₁.η (G₂.F₀ _) ∘ F₂.F₁ (f₂.η _) ∘ f₁.η (G.₀ _) ≈⟨ sym-assoc ⟩
+            (F₃.₁ (g₂.η _) ∘ g₁.η (G₂.₀ _)) ∘ F₂.₁ (f₂.η _) ∘ f₁.η (G.₀ _) ∎ }
```
sstucki, line 71: I think the general rule is to factor out large inline-definitions like these into explicit, stand-alone definitions, either in a `where` block after the record definition, or in a `private` block before `Endofunctors-Monoidal`. If the definition is useful in itself, it doesn't even need to be private.

The same goes for the definitions below.

src/Categories/Category/Monoidal/Instance/Endofunctors.agda
```diff
@@ -0,0 +1,131 @@
...
+  ≃→≅ i = record
+    { from = i.F⇒G
+    ; to = i.F⇐G
+    ; iso = record { isoˡ = λ {x} → Iso.isoˡ (i.iso x) ; isoʳ = λ {x} → Iso.isoʳ (i.iso x) }
+    } where module i = NaturalIsomorphism i
+
+module _ where
+  Endofunctors-Monoidal : ∀ {C : Category o ℓ e} → Monoidal (Endofunctors C)
+  Endofunctors-Monoidal {C} = record
+    { ⊗ = record
```
sstucki, line 49: This is already defined [here](https://github.com/agda/agda-categories/blob/master/src/Categories/Category/Construction/Functors.agda#L266) in `Category.Construction.Functors`, might be good to re-use it.

src/Categories/Category/Monoidal/Instance/Endofunctors.agda
```diff
@@ -0,0 +1,131 @@
+{-# OPTIONS --without-K --safe #-}
+module Categories.Category.Monoidal.Instance.Endofunctors {o ℓ e} where
```
sstucki, line 2: Personally, I'd have put this in `Categories.Category.Monoidal.Construction` rather than `Instance` since it's parametrized by an ambient category `C`.

sstucki (commented):
Nice!

I thought I had seen this before, then I realized I had written [something similar a while ago](https://github.com/agda/agda-categories/commit/6e5e05e2934da0bff465906dadc4a6d88dc1a5dc) 6e5e05e2934da0bff465906dadc4a6d88dc1a5dc but never submitted a PR. 😅 Looking at my old code, it seems that there are some existing definitions in the library that can be re-used.

Otherwise, this LGTM.

src/Categories/Category/Monoidal/Instance/Endofunctors.agda
```diff
@@ -0,0 +1,131 @@
...
+    ; iso = record { isoˡ = λ {x} → Iso.isoˡ (i.iso x) ; isoʳ = λ {x} → Iso.isoʳ (i.iso x) }
+    } where module i = NaturalIsomorphism i
+
+module _ where
+  Endofunctors-Monoidal : ∀ {C : Category o ℓ e} → Monoidal (Endofunctors C)
+  Endofunctors-Monoidal {C} = record
+    { ⊗ = record
+      { F₀ = uncurry′ _∘F_
+      ; F₁ = uncurry′ _∘ₕ_
+      ; identity = λ { {F , G} →
```
JacquesCarette, line 52: Here you don't use G at all, so why bother with the `module`?
Also, this is a 1-liner via `cancelˡ F.identity`.

src/Categories/Category/Monoidal/Instance/Endofunctors.agda
```diff
@@ -0,0 +1,131 @@
...
+      ; F-resp-≈ = λ { {_} {E , D} (f≈₁g , f≈₂g) → Functor.F-resp-≈ E f≈₂g ⟩∘⟨ f≈₁g }
+      }
+    ; unit = idF
+    ; unitorˡ = ≃→≅ unitorˡ
+    ; unitorʳ = ≃→≅ unitorʳ
+    ; associator = λ {X} {Y} {Z} → ≃→≅ (associator Z Y X)
+    ; unitorˡ-commute-from = identityˡ
+    ; unitorˡ-commute-to = λ { {f = f} →
+        let module f = NaturalTransformation f in
+        begin (id ∘ f.η _)        ≈⟨ identityˡ ⟩
```
JacquesCarette, line 81: `id-comm-sym` should let you cut this to 2 lines.

src/Categories/Category/Monoidal/Instance/Endofunctors.agda
JacquesCarette: (or maybe it's `elimˡ` )

src/Categories/Category/Monoidal/Instance/Endofunctors.agda
```diff
@@ -0,0 +1,131 @@
...
+              f.η _               ≈⟨ Equiv.sym identityʳ ⟩
+              (f.η _ ∘ id)        ≈⟨ Equiv.sym identityʳ ⟩
+              ((f.η _ ∘ id) ∘ id) ∎ }
+    ; unitorʳ-commute-from = λ { {F} {G} {f = f} →
+        let module G = Functor G in
+        let module f = NaturalTransformation f in
```
JacquesCarette, line 87: `f` is an unusual name for a NaturalTransformation.  Also, with `id-comm-sym` you can shorten the proof here as well.

JacquesCarette (changes requested):
There are some nice idioms in @sstucki 's version in #390 that should be copied to this version as well as the various comments on here. Definitely a nice contribution, thanks.

JacquesCarette:
Note that there will be a harder merge needed because of #390.

iwilare:
Sorry for the late reply, thank you for the useful suggestion! We never stop learning all the various idiosyncracies that make a good Agda PR. I think #390 makes our contribution obsolete, right? The version suggested in #390 looks way better than ours, I'm okay with just keeping that and closing this one!

PR #399, opened 2023-10-11
CI: Updated cached paths for `cabal` and update Haskell config action dep

### LINKED ISSUE #375: Caching on CI seems to be broken
Last few times, the cache was 45 bytes...  Something's definitely quite broken. I'd appreciate some help, as my knowledge in this area is thin (and that's a generous description.)

balacij:
Contributes to #375 

* This _should_ fix the "lack of caching issue" (#375) for the `cabal` cache. It looks like @JasonGross [had the right idea.](https://github.com/agda/agda-categories/issues/375#issuecomment-1668573583)
* Update the Haskell installation action to remove the deprecation warning in the CI.

I don't think this PR closes #375 because it doesn't have an Agda build cache, so building `agda-categories` is still completely fresh for each CI run, so we can have a look at that too.

To audit this solution, please have a look at my fork's [CI runs](https://github.com/balacij/agda-categories/actions), specifically the earliest 2's build times and logs. The first one is a completely fresh run without a cache, and the second one has a cache. There's a ~20min time cut with the cache hit, but I would hope that it can be brought down further by adding the Agda build cache to the CI cache too.

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -99,7 +99,7 @@ jobs:
 
       - name: Install ghc and cabal
         if: steps.cache-cabal.outputs.cache-hit != 'true'
-        uses: haskell/actions/setup@v2
+        uses: haskell-actions/setup@v2
         with:
           ghc-version: ${{ env.GHC_VERSION }}
           cabal-version: ${{ env.CABAL_VERSION }}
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):

balacij:
It seems that GitHub doesn't overwrite caches if hits occurred. On my fork, this wasn't an issue (because I have no caches!), but on this repo, we probably should have [evicted all caches](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#deleting-cache-entries) before merging this in. My bad!

PR #407, opened 2023-12-28
change definition of function sets equality to not build-in cong (

JacquesCarette:
see agda-stdlib issue #1467 for further information. Fix everything that needed fixed because of that - a lot!

Amazing how many things actually depend on the equivalence of function setoids. Which, by extension, just how much 'set theory' there is in category theory.


JacquesCarette:
Sorry, I screwed this up and made the changes on master, tried to fix it by going to a branch, and failed.

PR #408, opened 2024-01-10
Right adjoint to Functor.Slice.Free

Taneb:
Supercedes #370

I started over with a better understanding of the problem space, on top of the cleaned up version of the module. Also, I'm making this branch in the `agda` repo rather than my own fork so it's easier for other contributors to make commits.

- [X] Functor
- [x] Adjoint
- [ ] Corresponding functor and adjoint for base changes

Even with what I have so far there's a lot of potential for simplifying the proofs.

An insight I've had is that the proof passed in to universal constructions (and other ways to construct a morphism or object from a proof) should generally be `abstract`. for compilation times. Other proofs don't matter nearly as much in this respect. This is because we generally have many many copies of the morphism term in other terms and types.


Taneb:
This isn't complete but it's at a point where I'd like some other eyes on it if anyone wants to take a look

src/Categories/Adjoint/Instance/Slice.agda
```diff
@@ -44,3 +52,94 @@ module _ {A : Obj} (product : ∀ {X} → Product A X) where
...
+  open CartesianClosed ccc
+  open Cartesian cartesian
+  open Terminal terminal
+  open BinaryProducts products
+
+  Free⊣Coforgetful : Free {A = A} product ⊣ Coforgetful ccc pullback
+  Free⊣Coforgetful = record
+    { unit = ntHelper record
+      { η = λ X → p.universal (sliceobj π₁) (λ-unique₂′ (unit-pb X))
+      ; commute = λ {S} {T} f → p.unique-diagram (sliceobj π₁) !-unique₂ $ begin
```
JacquesCarette, line 67: For these longer proofs, best to put them either in `where` or named in a private block. Purely for efficiency.

src/Categories/Adjoint/Instance/Slice.agda
```diff
@@ -44,3 +52,94 @@ module _ {A : Obj} (product : ∀ {X} → Product A X) where
...
+      swap ∘ swap                                                                        ≈⟨ swap∘swap ⟩
+      id                                                                                 ∎
+    ; zag = λ {X} → p.unique-diagram X !-unique₂ $ begin
+      p.p₂ X ∘ p.universal X _ ∘ p.universal (sliceobj π₁) _                                                  ≈⟨ pullˡ (p.p₂∘universal≈h₂ X) ⟩
+      λg ((eval′ ∘ first (p.p₂ X) ∘ swap) ∘ eval′ ∘ first (p.p₂ (sliceobj π₁))) ∘ p.universal (sliceobj π₁) _ ≈˘⟨ pullˡ (subst ○ λ-cong assoc) ⟩
+      λg ((eval′ ∘ first (p.p₂ X) ∘ swap) ∘ eval′) ∘ p.p₂ (sliceobj π₁) ∘ p.universal (sliceobj π₁) _         ≈⟨ refl⟩∘⟨ p.p₂∘universal≈h₂ (sliceobj π₁) ⟩
+      λg ((eval′ ∘ first (p.p₂ X) ∘ swap) ∘ eval′) ∘ λg swap                                                  ≈⟨ subst ⟩
+      λg (((eval′ ∘ first (p.p₂ X) ∘ swap) ∘ eval′) ∘ first (λg swap))                                        ≈⟨ λ-cong (pullʳ β′) ⟩
+      λg ((eval′ ∘ first (p.p₂ X) ∘ swap) ∘ swap)                                                             ≈⟨ λ-cong (pull-last swap∘swap) ⟩
```
JacquesCarette, line 102: cancel-r ?

src/Categories/Functor/Slice.agda
```diff
@@ -96,3 +102,58 @@ module _ {A : Obj} where
...
+  -- e.g. we only need exponential objects with A as domain
+  -- I don't think we need all products but I don't have a clear idea of what products we do need
+  module _ (ccc : CartesianClosed) (pullback : ∀ {X} {Y} {Z} (h : X ⇒ Z) (i : Y ⇒ Z) → Pullback h i) where
+
+    open CartesianClosed ccc
+    open Cartesian cartesian
+    open Terminal terminal
+    open BinaryProducts products
+
+    -- Needs better name!
```
JacquesCarette, line 115: Yes, it does. Ask on the category theory Zulip? It feels like it ought to have a name. (Does the 1lab have this yet?)

JacquesCarette (changes requested):
It's looking really good. Tiny tweaks to do in one spot, otherwise, it's just comments.

src/Categories/Functor/Slice.agda
Taneb: 1lab calls this functor [`exponentiable→product`](https://1lab.dev/Cat.Diagram.Exponential.html#12821), and what we call `Free` they call [`constant-family`](https://1lab.dev/Cat.Instances.Slice.html#22029). I can't see a counterpart to our `Forgetful` but I've asked in their Discord.

I would like to, following _Polynomial Functors and Polynomial Monads_ (the paper I'm slowly working towards formalize here), call `Forgetful` `Σ`, `Free` `Δ`, and `Coforgetful` `Π`, but those names are even _less_ indicative of their function than what we currently have. _Aspects of Topoi_ uses the same names for `Forgetful` and `Coforgetful`, but calls `Free` `×`.

JacquesCarette: Yes, those single-letter names are even worse. The 1lab's names don't seem so bad?

I could try to summon @TOTBWF to see if he has opinions.

TOTBWF: I have been SUMMONED ✨🧊 🧙✨ 

To get a good idea for names, it's good to understand exactly what is going on with slices. The intuition here is that a slice `f : X -> I` is how we can encode an I-indexed family of objects X_i without any sort of classifying object like `Type`. In places like `Set`, we can move between the two by taking the fibres of `f`, but in a general category we cannot do this. 

As such, base change `pullback-functorial` should really be thought of as a functor on _slices_ `C/B -> C/A` for a morphism `A -> B`: from the perspective of families, this restricts a family `f : X -> B` to a new family `P -> A` along `f` potentially duplicating or deleting fibres. In other words, we take the fibre product: IE, the pullback. By composing with the projection out of the slice, we've essentially "forgotten" the indexing, and are just looking at the total space of the family.

From this perspective, `Free` isn't really the free slice, but rather _the constant family_ on `I`: each fibre over is a copy of `X`. Furthermore, `Forgetful` doesn't really "forget" a slice: it returns the total space of a family by forgetting the indexing.

`Coforgetful` is the trickiest: the bottom leg of the pullback is an obfuscation of the identity morphism, and the right leg is more or less postcomposition by the projection out of the family `X -> I`. Therefore, this pullback consists of sections of the family `X -> I` internalized as a exponential objects.

With all that background out of the way, here are my suggestions:
- Make `pullback-functorial` a functor on slices, and rename it to `Base-change` or `Reindex`
- `Free` becomes `Constant-family`
- `Forgteful` becomes `Total-space` or `Total`
- `Coforgetful` becomes `Internal-sections` or `Sections`

PS: this is some really great work!

Taneb: I like these names a lot! Thanks for the explanation.

While switching to them, I noticed that `pullback-functorial` (after @TOTBWF's suggestion to make it less forgetful) looks a lot like `Categories.Functor.Slice.BaseChange.BaseChange*`. I haven't quite been able to convince myself that they _are_ the same thing, because there's a lot of redirection and it's a little confusing to get my head around, but if they are, I'd like to remove `pullback-functorial` entirely.

Taneb:
Looks like I've broken `Categories.Category.CartesianClosed.Locally.Properties`, a module which I had honestly forgotten about, but I think I should be able to entirely replace that with `Categories.Functor.Slice.BaseChange`, with an actual proof of adjointness soon. It's making me wonder if there's a simpler implementation of `InternalSections`, though

src/Categories/Category/BinaryProducts.agda
```diff
@@ -94,6 +94,12 @@ record BinaryProducts : Set (levelOfTerm 𝒞) where
   ⁂-cong₂ : f ≈ g → h ≈ i → f ⁂ h ≈ g ⁂ i
   ⁂-cong₂ = [ product ⇒ product ]×-cong₂
 
+  first-cong : f ≈ g → f ⁂ h ≈ g ⁂ h
```
JacquesCarette, line 97: For consistency, I think these ought to be called `⁂-cong` with l/r superscripts.

JacquesCarette (commented):
This is looking really nice!

src/Categories/Category/BinaryProducts.agda
Taneb: I called them that because we do define `first` and `second` rather than `⁂ˡ` and `⁂ʳ` or anything. I'm not sure why I didn't use those in the type signature

JacquesCarette: Good point, I didn't remember that (nor didn't see them in the diffs to remind me). Right, no need to change them.

Taneb:
I'd like to finish this off... but I've gotten lost with what was going on with it :(

JacquesCarette:
Anything I could do to help?

Taneb:
Remaining considerations:

* Define the next base change functor (composition of `slice⇒slice-slice` and `InternalSections`)
* Define the base change adunction
* Can the proofs be simplified? They're huge and scary
* Can the environment of `InternalSections` be simplified? We don't need all products
* Is `BaseChange⁎` in `Categories.Category.CartesianClosed.Locally` the same thing as what we're defining in this PR? Is it a viable way to simplify the definitions?

Taneb:
I'm not happy with this and I don't think I'm going to become happy with this from where it is now. I'm going to try and take out some small nice stuff, get that merged, and then consider what I want to do for this

Taneb:
Revisiting this a bit. This PR is hitting something that makes Agda incredibly slow, especially when some things are holes.

PR #410, opened 2024-01-25
Properties of distributive categories

Reijix:
I've added some properties of the distribution isos in `Categories.Category.Distributive` that I am using in my personal project.

Im very open for naming suggestions for these properties, the current names might not be that good.

I also thought about putting them in `Categories.Category.Distributive.Properties` but it seems to me that as soon as you are working with distributive categories you will need (some of) these helpers.

### DIFF
--- a/src/Categories/Category/Distributive.agda
+++ b/src/Categories/Category/Distributive.agda
@@ -1,14 +1,12 @@
 {-# OPTIONS --without-K --safe #-}
 
-open import Level
-open import Categories.Category.Core
+open import Level using (levelOfTerm)
+open import Categories.Category.Core using (Category)
 open import Categories.Category.Cartesian using (Cartesian)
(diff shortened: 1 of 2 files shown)


JacquesCarette (changes requested):
While I agree that these are often needed as soon as you deal with Distributive categories, for scalability we've found that putting a lot of properties in the defining record tends to really bloat things up further down the line. So even though it is less convenient, putting them in `Categories.Category.Distributive.Properties` would scale better.

Another possibility is to define `Distributive` in ...`Distributive.Core` (and open it public in this file) and then leave all of these definitions here.  

Either of those would be fine with me.

JacquesCarette:
Should have said: really nice work. I do want to pull this in!

Reijix:
Good point, I moved the properties to `Categories.Category.Distributive.Properties`.

JacquesCarette (approved):

PR #411, opened 2024-02-02
Use definitions from stdlib

Reijix:
The stdlib now has these definitions so I added them.


JacquesCarette (approved):
Wonderful, thank you.

PR #413, opened 2024-03-14
Rename and expand Assoc4 reasoning combinators

Taneb:
I haven't hunted down every possible place where the new combinators could be used. I'm sure there's plenty.

This naming scheme isn't perfect but I think it's a lot better than what was there before


JacquesCarette (changes requested):
I wonder if assoc-1->4 (with proper unicode) instead of `assoc²αδ` would be better still? It's cute that you're asking the reader to remember the alphabetical order of the Greek alphabet... and isn't too high a bar for people doing category theory!

Definitely needed: a list of all 5 associations in the documentation. I don't want to have people be forced to reconstruct that in their heads, but just be able to look it up.

Otherwise, this is great!

JacquesCarette (approved):

Issue #419, opened 2024-05-26
Unbundling limit-y things


4e554c4c:
Something I've noticed, especially when dealing with limits, colimits and ends is that you occasionally want to state that a particular thing *is* the limit (/colimit/end) of a diagram, instead of merely stating that an end exists. This exists for the very specific limits `Terminal` and `Initial` (in the form of `IsTerminal` and `IsInitial`) but I think it would make sense for all limit-y things.

This is especially important when converting between different kinds of limit-y objects. E.g. limits -> ends, initial objects -> colimits, etc.

4e554c4c:
For an example of why this might be important, see a recent commit of mine that proves that initial objects are strict initial in a cartesian closed category: https://github.com/4e554c4c/agda-categories/commit/319c9f47c25debfba939e5afa78004c02e8e558c#diff-0261bd0924a764ad5ad50b05434d00e33a2929173a1477af974fd091e90cd941R92

In particular this proof relies on the apex of a colimit to be (definitionally equal to) a particular thing, but cannot state this assumption.

JacquesCarette:
Yes, absolutely. It so happens that it wasn't much needed up to now, but as we want to 'say more things', it very much becomes needed. So I'd welcome PRs that do this.

PR #422, opened 2024-06-14
End calculus

4e554c4c:
This PR is a collection of theorems and refactorings that were developed during my master's thesis.

This includes
* Initial objects
  * are colimits
* Strict initial objects
  * Initial objects in a Cartesian category are strict
* The category of twisted arrows
  * has a forgetful functor
  * `Twist (F ∘ J)` is iso to `F ∘ Twist J`
* End calculus
  * Ends are preserved by continuous functors
  * The setoid of natural transformations is an end (given size constraints)
  * Parameterized ends  (`EndF`)
    * are ends 
    * are preserved by continuous functors
  * The end of a parameterized end is an end in the product category (Fubini)

Many definitions could be simplified if #419 is resolved.

## Considerations

`--lossy-unification` is currently necessary for `Categories.Diagram.End.Fubini`. I was hoping that refactoring `EndF` to use an opaque `end-η` could fix this, but it does not.

### DIFF
--- a/src/Categories/Category/Construction/TwistedArrow.agda
+++ b/src/Categories/Category/Construction/TwistedArrow.agda
@@ -1,27 +1,24 @@
 {-# OPTIONS --without-K --safe #-}
-open import Data.Product using (_,_; _×_; map; zip)
-open import Function.Base using (_$_; flip)
-open import Level
-open import Relation.Binary.Core using (Rel)
 
 open import Categories.Category using (Category; module Definitions)
-open import Categories.Category.Product renaming (Product to _×ᶜ_)
-open import Categories.Functor
-
-import Categories.Morphism as M
-import Categories.Morphism.Reasoning as MR
 
 -- Definition of the "Twisted Arrow" Category of a Category 𝒞
 module Categories.Category.Construction.TwistedArrow {o ℓ e} (𝒞 : Category o ℓ e) where
 
+open import Level
+open import Data.Product using (_,_; _×_; map; zip)
+open import Function.Base using (_$_; flip)
+open import Relation.Binary.Core using (Rel)
+
+open import Categories.Category.Product renaming (Product to _×ᶜ_)
+open import Categories.Functor using (Functor)
+open import Categories.Morphism.Reasoning 𝒞 using (pullˡ; pullʳ )
+
 private
   open module 𝒞 = Category 𝒞
-
-
-open M 𝒞
 open Definitions 𝒞
-open MR 𝒞
 open HomReasoning
+
 
 private
   variable
@@ -78,14 +75,12 @@ TwistedArrow = record
 
 -- Consider TwistedArrow as the comma category * / Hom[C][-,-]
 -- We have the codomain functor TwistedArrow → C.op × C
-
-module _ where
-  open Morphism
-  open Morphism⇒
-  open Functor
-  Forget : Functor TwistedArrow (𝒞.op ×ᶜ 𝒞)
-  Forget .F₀ x = dom x , cod x
-  Forget .F₁ f = dom⇐ f , cod⇒ f
-  Forget .identity = Equiv.refl , Equiv.refl
-  Forget .homomorphism = Equiv.refl , Equiv.refl
-  Forget .F-resp-≈ e = e
+open Functor
+Codomain : Functor TwistedArrow (𝒞.op ×ᶜ 𝒞)
+Codomain .F₀ x = dom , cod
+  where open Morphism x
+Codomain .F₁ f = dom⇐ , cod⇒
+  where open Morphism⇒ f
+Codomain .identity = Equiv.refl , Equiv.refl
+Codomain .homomorphism = Equiv.refl , Equiv.refl
+Codomain .F-resp-≈ e = e
--- a/src/Categories/Category/Construction/TwistedArrow.agda
+++ b/src/Categories/Category/Construction/TwistedArrow.agda
@@ -1,20 +1,26 @@
 {-# OPTIONS --without-K --safe #-}
+open import Data.Product using (_,_; _×_; map; zip)
+open import Function.Base using (_$_; flip)
+open import Level
+open import Relation.Binary.Core using (Rel)
+
 open import Categories.Category using (Category; module Definitions)
+open import Categories.Category.Product renaming (Product to _×ᶜ_)
+open import Categories.Functor
+
+import Categories.Morphism as M
+import Categories.Morphism.Reasoning as MR
 
 -- Definition of the "Twisted Arrow" Category of a Category 𝒞
 module Categories.Category.Construction.TwistedArrow {o ℓ e} (𝒞 : Category o ℓ e) where
 
-open import Level
-open import Data.Product using (_,_; _×_; map; zip)
-open import Function.Base using (_$_; flip)
-open import Relation.Binary.Core using (Rel)
+private
+  open module 𝒞 = Category 𝒞
 
-import Categories.Morphism as M
+
 open M 𝒞
-open import Categories.Morphism.Reasoning 𝒞
-
-open Category 𝒞
 open Definitions 𝒞
+open MR 𝒞
 open HomReasoning
 
 private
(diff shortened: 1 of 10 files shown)


src/Categories/Diagram/End/Fubini.agda
```diff
@@ -0,0 +1,206 @@
...
+  -- following MacLane §IX.8
+  private module _ (ρ : Wedge ∫F) where
+    private
+      open module ρ = Wedge ρ using () renaming (E to x)
+    -- first, our wedge gives us a wedge in the product
+    ξ : Wedge F
+    ξ .apex = x
+    ξ .dinatural = dtHelper record
+      { α = λ { (c , p) → ω.dinatural.α (p , p) c ∘ ρ.dinatural.α p }
+      ; commute = λ { {c , p} {d , q} (f , s) → begin
```
JacquesCarette, line 76: Please pull out this inline proof of `commute` into a `where` clause below the record. You can then take advantage of this to declare some short-forms for things that are repeated a lot (and long) like `ω.dinatural.α` and `ρ.dinatural.α p` and likely more.  Even just naming `(C.id , P.id)` would save a lot of horizontal space.

src/Categories/Diagram/End/Fubini.agda
```diff
@@ -0,0 +1,206 @@
...
+    ρ .dinatural = dtHelper record
+      { α = λ p → ω.factor (p , p) record
+        { E = x
+        ; dinatural = dtHelper record
+          { α = λ c → ξ.dinatural.α (c , p)
+          ; commute = λ f → ξ.dinatural.commute (f , P.id)
+          }
+        }
+          -- end-η-commute ⦃ ω (?) ⦄ ⦃ ω (?) ⦄ (curry.₀.₁ (F ′) (?)) ?
+      ; commute = λ {p} {q} f → ω.unique′ (p , q) λ {c} → begin
```
JacquesCarette, line 131: Same comment as above here

src/Categories/Diagram/End/Fubini.agda
```diff
@@ -0,0 +1,206 @@
...
+  -- Thus the product end need not exist.
+  module _ {{eᵢ : ∫ (⨏ (F ′) {ω})}} where
+    private
+      module eᵢ = ∫ eᵢ
+    eₚ : ∫ F
+    eₚ .∫.wedge = ξ eᵢ.wedge
+    eₚ .∫.factor W = eᵢ.factor (ρ W)
+    eₚ .∫.universal {W} {c , p} = begin
+      ξ eᵢ.wedge .dinatural.α (c , p) ∘ eᵢ.factor (ρ W)               ≈⟨ Equiv.refl ⟩
+      (ω.dinatural.α (p , p) c ∘ eᵢ.dinatural.α p)  ∘ eᵢ.factor (ρ W) ≈⟨ assoc ⟩
```
JacquesCarette, line 164: these two lines are a push/pull (I always forget).

src/Categories/Functor/Bifunctor/Properties.agda
```diff
@@ -6,8 +6,12 @@ open import Level
 open import Data.Product using (Σ; _,_)
 
 open import Categories.Category
+open import Categories.Category.Product
```
JacquesCarette, line 9: Why all the additional imports? They don't seem to be needed / used !

src/Categories/Functor/Instance/Twisted.agda
```diff
@@ -1,62 +1,39 @@
...
-    ; commute = λ f → commute nt (dom⇐ f , cod⇒ f)
-    }
-    where
-    open NaturalTransformation
+  module C = Category C
+  module D = Category D
+
+open Morphism
+open Morphism⇒
+-- precomposition with the forgetful functor
```
JacquesCarette, line 27: Nice!!!

src/Categories/Diagram/End/Parameterized.agda
```diff
@@ -0,0 +1,220 @@
...
+    o ℓ e : Level
+    P C D E : Category o ℓ e
+
+-- we use MacLane's notation for paramaterized ends (CWM §IX.5)
+_♯ : Functor (P ×ᶜ (Category.op C ×ᶜ C)) D → Functor (Category.op C ×ᶜ C) (Functors P D)
+F ♯ = curry.₀ F.flip
+  where module F = Bifunctor F
+
+end-η♯ : {F G : Functor (P ×ᶜ (Category.op C ×ᶜ C)) D} (η : NaturalTransformation F G)
+         ⦃ ef : ∫ (F ♯) ⦄ ⦃ eg : ∫ (G ♯) ⦄ → NaturalTransformation (∫.E ef) (∫.E eg)
```
JacquesCarette, line 43: please don't use instance arguments. I'm fine with `opaque` and `unfolding`, but not these. Things are slow enough as it is, let's not make it worse.

JacquesCarette (changes requested):

JacquesCarette:
Re: `--lossy-unification`. Is this because otherwise it takes way too long? At this this is cause by Agda reconstructing some implicit arguments "the hard way" even though, to use, they are obvious. I've been able to cut some typechecking time down a lot by supplying just the right implicits. One big clue is if you ask Agda for the type of various things and stuff is eta-expanded. This definitely happens in the presence of things that compute, such as `×ᶜ`.

src/Categories/Category/CartesianClosed/Properties.agda
```diff
@@ -1,53 +1,99 @@
 {-# OPTIONS --without-K --safe #-}
 
-module Categories.Category.CartesianClosed.Properties where
```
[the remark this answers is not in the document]
4e554c4c: I couldn't really understand the style used. this makes sense! I'll fix the rest to match.

src/Categories/Category/Construction/TwistedArrow.agda
```diff
@@ -68,3 +74,18 @@ TwistedArrow = record
     cod⇒ m₁ ∘ (cod⇒ m₂ ∘ Morphism.arr A) ∘ (dom⇐ m₂ ∘ dom⇐ m₁) ≈⟨ refl⟩∘⟨ (pullˡ assoc) ⟩
     cod⇒ m₁ ∘ (cod⇒ m₂ ∘ Morphism.arr A ∘ dom⇐ m₂) ∘ dom⇐ m₁   ≈⟨ (refl⟩∘⟨ square m₂ ⟩∘⟨refl) ⟩
     cod⇒ m₁ ∘ Morphism.arr B ∘ dom⇐ m₁ ∎
+
+
+-- Consider TwistedArrow as the comma category * / Hom[C][-,-]
+-- We have the codomain functor TwistedArrow → C.op × C
```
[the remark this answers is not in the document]
4e554c4c: The domain functor here is trivial, so the category only has one forgetful functor that makes any sense. But I'm happy to change to Codomain

4e554c4c:
> Re: `--lossy-unification`. Is this because otherwise it takes way too long?

Yes, but I don't think this is all about reconstructing implicit arguments. It is also comparing terms. I reference #419 because one of the things that needs to be compared often is apices of ends

src/Categories/Diagram/End/Parameterized.agda
4e554c4c: I'm not sure if the instances are making things slow here.

They make a lot of definitions nicer. For example, in `Properties.agda` there is the type

```
end-η-resp-∘ : (α : NaturalTransformation F G) (β : NaturalTransformation G H) {{ef : ∫ F}} {{eg : ∫ G}} {{eh : ∫ H}} → end-η (β ∘ᵥ α) ≈ end-η β ∘ end-η α
```

Stating each end here for `end-η` complicates the definition a lot, and I don't think that it makes things nicer. Using implicit arguments isn't possible either, since the only portion of the end that is solved uniquely is the apex.

4e554c4c:
I can't resolve all of the `<greek letter>.dinatural.α blah` noise in the fubini proofs. I can move give names to some of them in `where` clauses if I define the ends using copatterns (as is probably more reasonable anyway). Really the main solution to this would be to resolve #7156 and allow for "applying" wedges (or ends) to an object to get the appropriate map.

4e554c4c:
@JacquesCarette  whenever you have the time, I would appreciate another look. I haven't changed everything that you suggested, but I left comments wherever I did not.

4e554c4c:
> ...would it be possible to split this into several other PRs? It's very big and scary to look at for me (this isn't a requirement but I'd appreciate it)

I can't split off too much, since github PRs can't depend on eachother. I split off some misc changes  into #423 and #424 (will probably add to both of these later)

JacquesCarette:
Sorry for the slowness - was on vacation all last week.

JacquesCarette (approved):

PR #423, opened 2024-06-25
Add strict initial objects

4e554c4c:
This commit add strict initial object, and proves that any initial object in a CCC is strict. There is also a refactoring of the empty diagram.


src/Categories/Category/Construction/Empty.agda
```diff
@@ -0,0 +1,13 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Category.Construction.Empty {o ℓ e}  where
```
Taneb, line 3: I think this already exists as `Categories.Category.Instance.Zero` (although your implementation is a lot tidier!)

4e554c4c: I see! I wonder why everything was using `liftC (Discrete 0)` then.

any preference on a name to use for both of them?

Taneb: Please stick with the existing name

4e554c4c: ah, there is also `Categories.Category.Instance.EmptySet`
quite a few of these.

Taneb: `Categories.Category.Instance.EmptySet` is certainly in the wrong place, but it's not a duplicate: it defines the initial objects for the `Sets` and `Setoids` categories

src/Categories/Object/StrictInitial.agda
```diff
@@ -0,0 +1,30 @@
+{-# OPTIONS --without-K --safe #-}
+open import Categories.Category
```
JacquesCarette, line 2: `.Core using (Category)`

JacquesCarette (changes requested):
Beautiful refactoring! One minor thing to fix before I merge.

JacquesCarette:
On the question

> I wonder why everything was using liftC (Discrete 0) then

This is because `Data.Empty.Polymorphic` did not yet exist when this was created. I contributed those because of this lack, but never had the chance to come back to `agda-categories` to refactor!

JacquesCarette (approved):

Issue #334, opened 2022-01-03
Suggested improvements for `IsPullback`


JacquesCarette:
(from @sergey-goncharov ):
I would suggest:

1. move the "unique" field to the bottom. It is the most sophisticated property, and must logically be 
the last one (this should more generally be taken as a rule for all limits and colimits).

2. I suggest to replace 'unique' with what is currently derived as 'unique-diagram'. This would make 
everything symmetric, and I would expect, in most cases, it would be easier to prove that something is a 
pullback.

As a bonus, uniqueness will no longer depend on 'eq : f ∘ h₁ ≈ g ∘ h₂', which is really not needed for 
it, but which gets in way.

TOTBWF:
These all seem like great changes!

When we do this, we should probably update `Pushout` as well.

jacquescomeaux:
This seems pretty straightforward. Are these changes still wanted?

sergey-goncharov:
> This seems pretty straightforward. Are these changes still wanted?

I think so.

JacquesCarette:
Some low-hanging fruit just need people with spare cycles!

PR #431, opened 2024-08-06
Implement IsPullback suggestions

jacquescomeaux:
This pull request is intended to address issue #334.

Tasks:
- [x] Move `unique` field to bottom
- [x] Replace `unique` with `unique-diagram`
- [x] Reconsider naming

### DIFF
--- a/src/Categories/Category/Instance/Properties/Setoids/Extensive.agda
+++ b/src/Categories/Category/Instance/Properties/Setoids/Extensive.agda
@@ -3,6 +3,7 @@
 module Categories.Category.Instance.Properties.Setoids.Extensive where
 
 open import Level using (Level)
+open import Data.Empty renaming (⊥ to ⊥′)
 open import Data.Product using (_,_)
 open import Data.Sum.Base as Sum using (_⊎_; inj₁; inj₂)
 open import Data.Sum.Relation.Binary.Pointwise using (inj₁; inj₂; _⊎ₛ_; drop-inj₁; drop-inj₂)
@@ -48,7 +49,10 @@ Setoids-Extensive ℓ = record
            }
         ; p₁∘universal≈h₁ = λ {_ _ _ eq x} → conflict A B (eq {x})
         ; p₂∘universal≈h₂ = λ {_ _ _ eq y} → conflict A B (eq {y})
-        ; unique = λ {_} {_} {_} {_} {eq} _ _ {x} → conflict A B (eq {x})
+        ; unique-diagram = λ {X} {h} {i} eq₁ eq₂ {x} →
+            let fromBottom : Level.Lift ℓ ⊥′ → [ ⊥ ][ h ⟨$⟩ x ≈ i ⟨$⟩ x ]
+                fromBottom = λ ()
+            in fromBottom (h ⟨$⟩ x)
         }
    }
      where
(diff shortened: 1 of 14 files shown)


src/Categories/Category/Instance/Properties/Setoids/Extensive.agda
```diff
@@ -3,6 +3,7 @@
 module Categories.Category.Instance.Properties.Setoids.Extensive where
 
 open import Level using (Level)
+open import Data.Empty renaming (⊥ to ⊥′)
```
JacquesCarette, line 6: Use `Data.Empty.Polymorphic` instead and you won't have to `Lift` below

src/Categories/Category/Instance/Properties/Setoids/Extensive.agda
```diff
@@ -46,9 +47,12 @@ Setoids-Extensive ℓ = record
            { to = λ z → conflict A B (eq {z})
            ; cong = λ {x} _ → conflict A B (eq {x})
            }
-        ; unique = λ {_} {_} {_} {_} {eq} _ _ {x} → conflict A B (eq {x})
         ; p₁∘universal≈h₁ = λ {_ _ _ eq x} → conflict A B (eq {x})
         ; p₂∘universal≈h₂ = λ {_ _ _ eq y} → conflict A B (eq {y})
+        ; unique-diagram = λ {X} {h} {i} eq₁ eq₂ {x} →
```
JacquesCarette, line 52: The previous proof was quite a bit more succinct, any chance to re-capture that here too?

src/Categories/Category/Instance/Properties/Setoids/Limits/Canonical.agda
```diff
@@ -58,9 +58,9 @@ pullback _ _ {X = X} {Y = Y} {Z = Z} f g = record
           }
         ; cong = < cong h₁ , cong h₂ >
         }
-      ; unique = λ eq₁ eq₂ → eq₁ , eq₂
       ; p₁∘universal≈h₁ = X.refl
       ; p₂∘universal≈h₂ = Y.refl
+      ; unique-diagram = λ eq₁ eq₂ → eq₁ , eq₂
```
JacquesCarette, line 63: would `_,_` do?

JacquesCarette (commented):
Coming along nicely - a few comments on the current state.

src/Categories/Category/Instance/Properties/Setoids/Limits/Canonical.agda
jacquescomeaux: It doesn't seem like that will work, because there is an implicit argument after eq2

src/Categories/Category/Instance/Properties/Setoids/Extensive.agda
jacquescomeaux: Thanks, that makes things a little nice below

src/Categories/Category/Instance/Properties/Setoids/Extensive.agda
jacquescomeaux: Got it back to one line

jacquescomeaux:
What about naming? Is `unique-diagram` the right name for the property now included as a record field?

JacquesCarette:
Thanks for the updates (and sorry on my slowness!).  On name: `unique-diagram` seems super generic (as was unique...) but if we're going to change it, then maybe `unique-pb-square` would be more indicative?

jacquescomeaux:
I don't really think the name needs to change. I just wanted to make sure there wasn't an obviously-better alternative before I consider the PR done.

JacquesCarette:
Sure, we can go with that and see if people complain.

JacquesCarette (approved):

PR #438, opened 2024-10-21
Bump standard library dependency to 2.1

### LINKED ISSUE #437: Support for stdlib 2.1?
stdlib 2.1 is released recently,
- https://github.com/agda/agda-stdlib/releases/tag/v2.1

and so I expect categories also should support it.

Taneb:
Closes #437 

No code changes are necessary but i think this unlocks `Function.Relation.Binary.Equality` for us: do we want to alter `Categories.Category.Instance.Setoids` to use this at the same time?

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -46,7 +46,7 @@ on:
 
 env:
   AGDA_COMMIT: tags/v2.6.4
-  STDLIB_VERSION: "2.0"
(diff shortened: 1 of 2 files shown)


JacquesCarette:
If you have the time, that would be great. Otherwise this can be merged now (and a release will have to be made) and that PR can come later.

JacquesCarette (approved):

JacquesCarette (changes requested):
Need to change the packaged `libraries` that the CI uses too - build is failing.

PR #443, opened 2024-12-03
Category and Cartesian instances for polymorphic predicates

AJChapman:
This is a new category in which the objects are predicates (wrapped in a `PRED` record implicitly containing the underlying type and the predicate) and the morphisms map between these objects. This differs from the `Preds` instance in Categories.Category.Instance.Preds in that the morphisms can change the underlying type as well as the predicate. I'm tentatively referring to this as 'polymorphic' to distinguish it from the existing instance.

I'm not certain that this fits into this library, and if it does then maybe I need to break it apart into separate files for `PRED`, `Preds` and `PredsCartesian`? I'd be happy to adapt it if you're interested, or just to get feedback on any of it as I'm fairly new to Agda and dependent types.

### DIFF
--- a/src/Categories/Category/Instance/Preds/Polymorphic.agda
+++ b/src/Categories/Category/Instance/Preds/Polymorphic.agda
@@ -1,0 +1,226 @@
+module Categories.Category.Instance.Preds.Polymorphic {o} where
+
+{-
+A Category in which Predicates are the Objects
+
+Note that this is different from the category defined in `Categories.Category.Instance.Preds`, because in that category the objects are always predicates on the same type.
+Here we are making a category in which the objects are all predicates, but not necessarily of the same type.
+So this category actually generalises the `Sets` category, since one possible predicate of any type is the universal predicate, allowing all elements of the type.
+
+This idea is drawn from Conal Elliott's paper Timely Computation.
+-}
+
+open import Level using (0ℓ; Level; suc; _⊔_)
+open import Function using (id; _∘_)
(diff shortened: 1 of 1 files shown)


JacquesCarette:
Note that I'm away at a conference right now, so won't be able to review in detail for another few days.

Because PRED is essentially isomorphic to 'containers' which is also known as 'polynomial functors', we definitely want this. Though doing it atop 'containers' is likely the way to go.

TOTBWF:
A better name for this would be `Fam(Set)`; it is precisely the total category of the [family fibration](https://1lab.dev/Cat.Displayed.Instances.Family.html#the-family-fibration) on the category of sets.

Containers/Poly would have the maps going "backwards", EG:
```agda
record _⇒_ (A B : PRED) : Set o where
  constructor mk⇒
  open PRED
  field
    {f} : U A → U B
    imp : ∀ {u} → P B (f u) → P A u
```

TOTBWF:
It might be worth doing `Fam` in it's generality though; this would give you `Fam(Set)`, `Fam(Setoids)`, and `Fam(Setoids^op)` all in one go.

JacquesCarette:
Thanks for catching that I missed the orientation reversal. And yes, it probably is worth doing `Fam` instead. Do you know what that means @AJChapman  ?

AJChapman:
I don't yet know what it means. My category theory knowledge is quite limited. I'll read up on what a 'family fibration' is at the 1lab link @TOTBWF gave. I see that 1lab already has this implemented, so I can use their code as a guide.

PR #446, opened 2024-12-09
Add predicate form for pushouts

jacquescomeaux:


JacquesCarette (approved):
Hopefully `Pullback` is done in the same style too?

Issue #417, opened 2024-05-21
Contribution of lemmas about colimits


t-wissmann:
Hi,

thank you all for agda-categories :-) I spent almost entire last year on an initial F-algebra construction, which builds the initial algebra from finite, recursive F-coalgebras ("Initial Algebras Unchained", accepted at LICS'24, [preprint](https://arxiv.org/abs/2405.09504), [formalization](https://git8.cs.fau.de/software/initial-algebras-unchained), [agda html](https://arxiv.org/src/2405.09504v2/anc/index.html)). For the construction, I had written a couple of lemmas (mostly about colimits) which I find helpful for a general audience. Before flooding you with PRs, I'd like to double check with you, which ones you find useful:

* `singleton-hom 𝒞 X Y` [(doc)](https://arxiv.org/src/2405.09504v2/anc/Helper-Definitions.html#singleton-hom): the property that the hom set 𝒞(X,Y) is singleton, with notation
  `_[_=∃!=>_]` (𝒞 [ X =∃!=> Y ])

* For a diagram `F : Functor J C` and another functor `G : Functor C D`, one obtains a `Functor (Cocones F) (Cocones (G ∘F F))` ([doc](https://arxiv.org/src/2405.09504v2/anc/Helper-Definitions.html#1788)). This functor just does `F₀ = F-map-Coconeˡ G` and `F₁ = F-map-Cocone⇒ˡ G` (from Categories.Diagram.Cocone.Properties), but it's very useful to have a functor for the next lemma (with different identifiers, sorry!):

* If a functor `F` sends a colimit `C₁ : Colim J` to a colimitting cocone `FC₁ : Colim (F ∘F J)`, then every colimitting J-cocone is sent to a colimitting cocone ([doc](https://arxiv.org/src/2405.09504v2/anc/Colimit-Lemmas.html#2657)). On paper the proof is: colimits are unique up to isomorphism and functors preserve isomorphism :-)

* The colimit injections form a jointly-epic family ([doc](https://arxiv.org/src/2405.09504v2/anc/Colimit-Lemmas.html#jointly-epic))

* For a diagram in a full subcategory of a category 𝒞: If the colimit in 𝒞 exists and if the coapex is isormorphic to some object in the subcategory, then this yields the colimit in the subcategory ([doc](https://arxiv.org/src/2405.09504v2/anc/Colimit-Lemmas.html#FullSub-Colimit))

[...]

PR #448, opened 2025-01-27
Add functors between (co)cone categories

t-wissmann:
This adds three functors between cone categories (and cocone categories, respectively). All F₀ and F₁ fields are maps that already have been defined in Categories.Diagram.Co(co)ne.Properties, and thus the Functor definition is places there as well. Unfortunately, using the definition of Cones (and Cocones) requires the import of
Categories.Category.Construction.Cones (and Cocones) in the Cones.Properties module.

For `F : Functor J C`, the new functors are:

  - Every `G : Functor C D` lifts to a `Functor (Cones F) (Cones (G ∘F F))` and `Functor (Cocones F) (Cocones (G ∘F F))` (from #417)
  - Every `G : Functor J′ J` lifts to a `Functor (Cones F) (Cones (F ∘F G))` and `Functor (Cocones F) (Cocones (F ∘F G))`
  - Every `G : Functor J C` and `α : NaturalTransformation F G` extends to a `Functor (Cones F) (Cones G)` and `Functor (Cocones G) (Cocones F)`

In Cocones.Properties, I've restricted the import of Cone.Properties to the four maps that are actually used.

### DIFF
--- a/src/Categories/Diagram/Cocone/Properties.agda
+++ b/src/Categories/Diagram/Cocone/Properties.agda
@@ -8,8 +8,9 @@ open import Categories.Category
 open import Categories.Functor
 open import Categories.Functor.Properties
 open import Categories.NaturalTransformation
-open import Categories.Diagram.Cone.Properties
+open import Categories.Diagram.Cone.Properties using (F-map-Coneʳ; F-map-Cone⇒ʳ; nat-map-Cone; nat-map-Cone⇒)
 open import Categories.Diagram.Duality
+open import Categories.Category.Construction.Cocones using (Cocones)
 
 import Categories.Diagram.Cocone as Coc
 import Categories.Morphism.Reasoning as MR
@@ -45,6 +46,15 @@ module _ {F : Functor J C} (G : Functor C D) where
     }
     where open CF.Cocone⇒ f
 
+  mapˡ : Functor (Cocones F) (Cocones (G ∘F F))
+  mapˡ = record
+    { F₀ = F-map-Coconeˡ
+    ; F₁ =  F-map-Cocone⇒ˡ
+    ; identity = G.identity
+    ; homomorphism = G.homomorphism
+    ; F-resp-≈ = G.F-resp-≈
+    }
+
 module _ {F : Functor J C} (G : Functor J′ J) where
   private
     module C   = Category C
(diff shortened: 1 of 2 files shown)


t-wissmann:
Beside the location of the above functors, I also wasn't to certain about the naming scheme for them. I simply took the existing names and replaced `Coneˡ` and `Coconeˡ⇒` etc by `Functor`, obtaining:
* `F-map-Functorˡ : Functor (Cocones F) (Cocones (G ∘F F))`
* `F-map-Functorʳ : Functor (Cocones F) (Cocones (F ∘F G))`
* `nat-map-Functor : Functor (Cocones G) (Cocones F)`

src/Categories/Diagram/Cocone/Properties.agda
```diff
@@ -45,6 +46,15 @@ module _ {F : Functor J C} (G : Functor C D) where
     }
     where open CF.Cocone⇒ f
 
+  F-map-Functorˡ : Functor (Cocones F) (Cocones (G ∘F F))
+  F-map-Functorˡ = record
+    { F₀ = F-map-Coconeˡ
+    ; F₁ =  F-map-Cocone⇒ˡ
+    ; identity = λ {A} → G.identity
+    ; homomorphism = λ {X} {Y} {Z} {f} {g} → G.homomorphism
```
JacquesCarette, line 54: here (and everywhere in this PR), can you delete implicits if/when they are not needed please?

JacquesCarette (changes requested):
one nitpick, and this is good to go, at least code-wise.

JacquesCarette:
On names: I wonder if `mapˡ` and `nat-map` would be sufficient? Certainly the `F-` prefix seems redundant, no?

src/Categories/Diagram/Cocone/Properties.agda
t-wissmann: Sure! I've removed all implicit arguments from the lambdas in my commit.

t-wissmann:
> On names: I wonder if `mapˡ` and `nat-map` would be sufficient? Certainly the `F-` prefix seems redundant, no?

I agree; maybe it would still be preferrable to include `Cones` etc in the name such that no name conflict arises (so far, `Cocones.Properties` and `Cones.Properties` don't have any conflicting names). So maybe: `map-Conesˡ`, `map-Coconesˡ`, `nat-map-Cones`?

JacquesCarette:
I admit that I was thinking of the usage as having `open import Categories.Diagram.Cones as Cones` in the header, so that having `Cones` in the function name doubles things up.

t-wissmann:
I've just renamed the functions to `mapˡ`, `nat-map` etc. :-)

JacquesCarette (approved):

t-wissmann:
Thanks for approving this PR! Is there anything left that I can (or need to) do?

JacquesCarette:
No, I just needed to let the checks run to completion before I could merge (and then I got insanely busy). Merging now.

PR #449, opened 2025-02-06
Pullback properties

frederikgebert:
I proved the [pasting law for pullbacks](https://ncatlab.org/nlab/show/pasting+law+for+pullbacks) and added it to Pullback/Properties.agda.
I'm new to agda, so any feedback is much appreciated. I hope this fits into this library!

### DIFF
--- a/src/Categories/Diagram/Pullback/Properties.agda
+++ b/src/Categories/Diagram/Pullback/Properties.agda
@@ -204,88 +204,81 @@ module PullbackPastingLaw {A B C D E F : Obj}
   (ABDE : i ∘ f ≈ k ∘ h) (BCEF : j ∘ g ≈ l ∘ i) (pbᵣ : IsPullback g i j l) where
   
   open IsPullback using (p₁∘universal≈h₁; p₂∘universal≈h₂; universal; unique-diagram)
-  
-  ACDF : j ∘ (g ∘ f) ≈ (l ∘ k) ∘ h
-  ACDF = begin
-    j ∘ g ∘ f   ≈⟨ pullˡ BCEF ⟩ 
-    (l ∘ i) ∘ f ≈⟨ assoc ⟩
-    l ∘ (i ∘ f) ≈⟨ pushʳ ABDE ⟩ 
-    (l ∘ k) ∘ h ∎
 
   leftPullback⇒bigPullback : IsPullback f h i k → IsPullback (g ∘ f) h j (l ∘ k)
   leftPullback⇒bigPullback pbₗ = record
-     { commute = ACDF
-     ; universal = universalb
-     ; p₁∘universal≈h₁ = [g∘f]∘universalb≈h₁
-     ; p₂∘universal≈h₂ = p₂∘universal≈h₂ pbₗ
-     ; unique-diagram = unique-diagramb
-     } where
-        j∘h₁≈l∘k∘h₂ : {H : Obj} {h₁ : H ⇒ C} {h₂ : H ⇒ D} → j ∘ h₁ ≈ (l ∘ k) ∘ h₂ → j ∘ h₁ ≈ l ∘ (k ∘ h₂)
-        j∘h₁≈l∘k∘h₂ eq = begin
-          j ∘ _       ≈⟨ eq ⟩
-          (l ∘ k) ∘ _ ≈⟨ assoc ⟩
-          l ∘ (k ∘ _) ∎
+    { commute = ACDF
+    ; universal = universalb
+    ; p₁∘universal≈h₁ = [g∘f]∘universalb≈h₁
+    ; p₂∘universal≈h₂ = p₂∘universal≈h₂ pbₗ
+    ; unique-diagram = unique-diagramb
+    } where
+      ACDF : j ∘ (g ∘ f) ≈ (l ∘ k) ∘ h
+      ACDF = begin
+        j ∘ g ∘ f   ≈⟨ extendʳ BCEF ⟩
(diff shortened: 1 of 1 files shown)


src/Categories/Diagram/Pullback/Properties.agda
```diff
@@ -189,3 +189,130 @@ module IsoPb {X Y Z} {f : X ⇒ Z} {g : Y ⇒ Z} (pull₀ pull₁ : Pullback f g
...
+
+-- pasting law for pullbacks:
+-- in a commutative diagram of the form
+-- A -> B -> C
+-- |    |    |
+-- D -> E -> F
+-- if the right square (BCEF) is a pullback,
+-- then the left square (ABDE) is a pullback
+-- iff the big square (ACDF) is a pullback.
+module PullbackPastingLaw {A B C D E F : Obj} (f : A ⇒ B) (g : B ⇒ C) (h : A ⇒ D) (i : B ⇒ E) (j : C ⇒ F) (k : D ⇒ E) (l : E ⇒ F) (ABDE : i ∘ f ≈ k ∘ h) (BCEF : j ∘ g ≈ l ∘ i) (pbᵣ : IsPullback g i j l) where
```
JacquesCarette, line 202: You can probably make f ... l implicit; certainly g, i, j, l can be,  Also, line is too long, please cut; I think putting the morphisms on a new line, then the rest again on a new one would make sense.

src/Categories/Diagram/Pullback/Properties.agda
```diff
@@ -189,3 +189,130 @@ module IsoPb {X Y Z} {f : X ⇒ Z} {g : Y ⇒ Z} (pull₀ pull₁ : Pullback f g
...
+  open IsPullback using (p₁∘universal≈h₁; p₂∘universal≈h₂; universal; unique-diagram)
+  
+  ACDF : j ∘ (g ∘ f) ≈ (l ∘ k) ∘ h
+  ACDF = begin
+    j ∘ g ∘ f ≈⟨ sym-assoc ⟩
```
JacquesCarette, line 207: The push and pull combinators would help shorten this proof to 3 lines down from 5.

src/Categories/Diagram/Pullback/Properties.agda
```diff
@@ -189,3 +189,130 @@ module IsoPb {X Y Z} {f : X ⇒ Z} {g : Y ⇒ Z} (pull₀ pull₁ : Pullback f g
...
+  leftPullback⇒bigPullback : IsPullback f h i k → IsPullback (g ∘ f) h j (l ∘ k)
+  leftPullback⇒bigPullback pbₗ = record
+     { commute = ACDF
+     ; universal = universalb
+     ; p₁∘universal≈h₁ = [g∘f]∘universalb≈h₁
+     ; p₂∘universal≈h₂ = h∘universalb≈h₂
+     ; unique-diagram = unique-diagramb
+     } where
+        j∘h₁≈l∘k∘h₂ : {H : Obj} {h₁ : H ⇒ C} {h₂ : H ⇒ D} → j ∘ h₁ ≈ (l ∘ k) ∘ h₂ → j ∘ h₁ ≈ l ∘ (k ∘ h₂)
+        j∘h₁≈l∘k∘h₂ {H} {h₁} {h₂} eq = begin
```
JacquesCarette, line 223: try to line up the equations please.

src/Categories/Diagram/Pullback/Properties.agda
```diff
@@ -189,3 +189,130 @@ module IsoPb {X Y Z} {f : X ⇒ Z} {g : Y ⇒ Z} (pull₀ pull₁ : Pullback f g
...
+          j ∘ h₁ ≈⟨ eq ⟩
+          (l ∘ k) ∘ h₂ ≈⟨ assoc ⟩
+          l ∘ (k ∘ h₂) ∎
+                  
+        universalᵣ[h₁,k∘h₂] : {H : Obj} {h₁ : H ⇒ C} {h₂ : H ⇒ D} → j ∘ h₁ ≈ (l ∘ k) ∘ h₂ → H ⇒ B
+        universalᵣ[h₁,k∘h₂] {H} {h₁} {h₂} eq = universal pbᵣ {H} {h₁} {k ∘ h₂} (j∘h₁≈l∘k∘h₂ eq)
```
JacquesCarette, line 229: in the definition (line 229, not 228) can you figure out which of the implicits can be omitted, if any? You can use named bindings to only give the ones necessary.  Here and through the rest of the file please.

JacquesCarette (changes requested):
This is generally very nice. Just a few minor comments.

There might be a second round, as some thing might become more obvious once these clean-ups are down.

frederikgebert:
Thank you for your feedback. I removed all implicit arguments and everything still works. I'm very impressed by Agda's ability to fill in those arguments! I also implemented the other changes you proposed. Everything is so much cleaner now.. :O
I will go through the proof again in the next days to see if I can shorten it by removing unnecessary definitions.

JacquesCarette:
Don't go overboard on shortening. Sometimes some intermediate definitions are useful for conceptual clarity. But in general, yes, shorter is better.

frederikgebert:
I removed some definitions that seemed redundant and moved intermediate equations to the definitions that use them. This increases clarity because the typing judgements are much shorter.
I also removed some lines by using pull, push and extend reasoning.
Sorry for basically rewriting the whole proof but I think it makes more sense now.

JacquesCarette (approved):

PR #451, opened 2025-02-21
Copy over a couple of properties from ends to coends

Taneb:


Taneb:
I feel like these should be moved into new modules `Categories.Diagram.{Coe,E}nd.{Col,L}imit`. Took me way too long to realise that this existed for `End`s after I'd been trying to write them for `Coend`s.

JacquesCarette (approved):

JacquesCarette:
I'd be fine with such a move. Do you want to do that in this PR or another one?

Taneb:
Another, I think

PR #452, opened 2025-02-22
Factor (co)end (co)limits into their own modules

Taneb:
Doing this has made it clear just how much more we have for ends than for coends... I'll save that for another PR though


JacquesCarette (approved):

JacquesCarette:
That might have been my 'fault' - I was going through Loregian's book and ran out of steam at some point.

JacquesCarette:
Hmm, the *code* looks good, but the CI isn't happy...

Taneb:
CI failure is I need to change the imports in `Categories.Diagram.End.Parameterized`. Fix incoming...

JacquesCarette (approved):

PR #453, opened 2025-02-24
Bring End and Coend properties to parity

Taneb:


src/Categories/Diagram/End/Properties.agda
```diff
@@ -101,17 +128,19 @@ module _ {F : Bifunctor (Category.op C) C D} (ω₁ ω₂ : ∫ F) where
...
-  open Iso
   open MR D
-  open HomReasoning
 
   end-unique : ∫.E ω₁ ≅ ∫.E ω₂
-  end-unique .to = ω₁.factor ω₂.wedge
-  end-unique .from = ω₂.factor ω₁.wedge
-  end-unique .iso .isoʳ = ω₂.unique′ $ pullˡ ω₂.universal ○ ω₁.universal ○ ⟺ identityʳ
-  end-unique .iso .isoˡ = ω₁.unique′ $ pullˡ ω₁.universal ○ ω₂.universal ○ ⟺ identityʳ
+  end-unique = record
```
JacquesCarette, line 136: Note that I'm fine with copatterns in `agda-categories`. So don't feel you have to switch those. No need to revert this change though.

JacquesCarette (commented):
Looking promising.

PR #456, opened 2025-02-26
liftC C is equivalent to C

Taneb:


src/Categories/Category/Lift/Properties.agda
```diff
@@ -0,0 +1,39 @@
...
+open import Categories.Category.Core
+open import Categories.Category.Equivalence
+open import Categories.Category.Lift
+open import Categories.NaturalTransformation.NaturalIsomorphism
+import Categories.Morphism.Reasoning.Core as MR
+
+unliftF-liftF-weakInverse : ∀ {o ℓ e} o′ ℓ′ e′ (C : Category o ℓ e) → WeakInverse (unliftF o′ ℓ′ e′ C) (liftF o′ ℓ′ e′ C)
+unliftF-liftF-weakInverse o′ ℓ′ e′ C = record
+  { F∘G≈id = niHelper record
+    { η = λ X → id
```
JacquesCarette, line 16: I think it is worth not naming the arguments in these cases (and below), i.e. `λ _ →`. Though `const` seems a bit much?

JacquesCarette (changes requested):

src/Categories/Category/Lift/Properties.agda
Taneb: I didn't check, but I suspect that using `const` might actually be a type error - the variables are still used implicitly, but `const`'s type prevents them from being in scope

JacquesCarette: you're most likely right about `const`

JacquesCarette (approved):

PR #458, opened 2025-02-27
Locally Graded Categories

TOTBWF:
This PR adds a definition of [locally graded categories](https://ncatlab.org/nlab/show/locally%20graded%20category) which are probably the "correct" notion of displayed category in E-category theory. They avoid the normal transport hell that comes with displayed setoids by directly axiomatizing a sort of "directed transport" operation along a morphism in a bicategory, which allows for all of the displayed equations to be turned into simple fibrewise ones.

### DIFF
--- a/src/Categories/LocallyGraded.agda
+++ b/src/Categories/LocallyGraded.agda
@@ -1,0 +1,94 @@
+{-# OPTIONS --without-K --safe #-}
+module Categories.LocallyGraded where
+
+open import Level
+open import Data.Unit using (tt)
+open import Data.Product using (_,_)
+
+open import Categories.Bicategory
+open import Categories.NaturalTransformation.NaturalIsomorphism using (NaturalIsomorphism)
+
+import Categories.Bicategory.Extras
+
+-- Locally Graded categories, generalized to gradation via a bicategory.
+--
+-- Locally graded categories are a joint generalization of displayed categories
+-- and enriched categories: display can be obtained by picking a discrete
+-- bicategory as your gradition, and enrichment can be obtained by asking
+-- for representing objects of hom setoids.
+record LocallyGraded
+  {o ℓ e t}
+  (B : Bicategory o ℓ e t)
+  (o' ℓ' e' : Level)
+  : Set (o ⊔ ℓ ⊔ e ⊔ t ⊔ suc o' ⊔ suc ℓ' ⊔ suc e') where
+  open Categories.Bicategory.Extras B
+  open Shorthands
+
+  infix 4 _⇒[_]_ _≈ᵥ_
+  infixr 11 _∘'_
+  infix 12 _⟨_⟩
+
+  field
+    -- Displayed objects and displayed morphisms over 1-cells.
+    Obj[_] : Obj → Set o'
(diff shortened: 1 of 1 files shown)


src/Categories/LocallyGraded.agda
```diff
@@ -0,0 +1,94 @@
...
+  field
+    -- Displayed objects and displayed morphisms over 1-cells.
+    Obj[_] : Obj → Set o'
+    _⇒[_]_ : ∀ {X Y}  → Obj[ X ] → X ⇒₁ Y → Obj[ Y ] → Set ℓ'
+
+    -- We opt to only require fibrewise setoids, as opposed to a displayed setoid.
+    _≈ᵥ_ : ∀ {X Y X' Y'} {f : X ⇒₁ Y} → X' ⇒[ f ] Y' → X' ⇒[ f ] Y' → Set e'
+
+    -- Directed transport along 2-cells in the grading category.
+    _⟨_⟩ : ∀ {X Y X' Y'} {f g : X ⇒₁ Y} → X' ⇒[ g ] Y' → f ⇒₂ g → X' ⇒[ f ] Y'
```
JacquesCarette, line 40: why is this contravariant?

src/Categories/LocallyGraded.agda
```diff
@@ -0,0 +1,94 @@
...
+open import Categories.Bicategory
+open import Categories.NaturalTransformation.NaturalIsomorphism using (NaturalIsomorphism)
+
+import Categories.Bicategory.Extras
+
+-- Locally Graded categories, generalized to gradation via a bicategory.
+--
+-- Locally graded categories are a joint generalization of displayed categories
+-- and enriched categories: display can be obtained by picking a discrete
+-- bicategory as your gradition, and enrichment can be obtained by asking
```
JacquesCarette, line 17: gradition -> gradation

JacquesCarette (changes requested):
So while this is nice, it's not clear that this really does behave better. I'd like to see more kit before pulling it in. For example, do these form a (Bi)Category for fixed base B?

src/Categories/LocallyGraded.agda
TOTBWF: I think of the transport as a sort of substitution operation, so it feels more natural to be contravariant. Also has the nice benefit of making the identity laws line up with the directions of the unitors/associators that you'd expect.
Ultimately doesn't matter too much though, as you can get the opposite variance by grading over co.

Taneb:
Glad to see someone who knows what they're doing (i.e. not me) doing this!

Is it missing that `_≈ᵥ_` is an equivalence relation?

TOTBWF:
Yes it absolutely is; haven't touched setoids for a minute obviously :) Also needs congruence lemmas everywhere.

src/Categories/LocallyGraded.agda
JacquesCarette: I definitely see the laws lining up, which is nice. It just seems weird that `f ⇒₂ g` is used to transform a `X' ⇒[ g ] Y'` to a `X' ⇒[ f ] Y'`. That 2-cell might not be invertible. Calling it a "sort of substitution operation" only mildly helps.

PR #463, opened 2025-05-02
Fix typo in product of functors distributive law

jacquescomeaux:
The domain of the first functor is preventing this from being as general as it should be. I noticed this when I tried to use this property and agda couldn't infer the unused implicit argument.


JacquesCarette (approved):
Nicely spotted! Thanks.

PR #464, opened 2025-05-05
Minor reasoning combinator updates

jacquescomeaux:
Hi, a few more tweaks from using the library. Having the first argument of `push-center` be of the form `c≈a∘b` would make it consistent with the other "pushes" in that module. Also, added `merge₁ʳ` and `merge₁ˡ` to complement `merge₂ʳ` and `merge₂ˡ`.


JacquesCarette:
It would be nice to also see these used in the library - but that can wait for another PR.

PR #465, opened 2025-05-15
agda 2.8.0

andreasabel:
To prepare for use with Agda 2.8.0, nothing much is needed to be done.
Agda has an improved check for the `UselessPrivate` so it found some:

- **Remove some useless `private`s that Agda-2.8.0 warns about**
  
When I saved a file my editor removed trailing whitespace.  So I did this systematically with `fix-whitespace $(find -name "*.agda")`.

- **Remove trailing whitespace in `.agda` files**
  
If you don't like this cleanup, I can remove this commit again.

Btw, your CI is on some older GHC.  This could be bumped.
It is also on Agda-2.6.4 when we already have 2.7.0.1 for quite a while...


JacquesCarette:
Thanks for the fixes, and all the remarks about the CI. All of these should be fixed indeed. [I need some co-maintainers, I'm having a hard time keeping up during teaching terms.]

JacquesCarette (approved):

Issue #467, opened 2025-05-15
Bump CI version up to more modern GHC and Agda


JacquesCarette:
@andreasabel mentions
> Btw, your CI is on some older GHC. This could be bumped.
> It is also on Agda-2.6.4 when we already have 2.7.0.1 for quite a while...

in PR #465 which ought to be done.

andreasabel:
Bumped Agda to 2.7.0.1 (and dropped GHC) in #468.

Issue #466, opened 2025-05-15
Speed up CI by installing Agda via setup-agda


andreasabel:
The CI could be sped up significantly by installing binary Agda via the [setup-agda action](https://github.com/wenkokke/setup-agda), since you are using a released version of Agda rather than `master` or a branch.

Significant time would be saved by
- not having to install GHC and Cabal
- not having to build Agda from source

The problem with the current CI is that Github deletes caches rather quickly, e.g. for my current run no caches were present, when the previous run was not soooo long ago:
- My run: 2025-05-15 https://github.com/agda/agda-categories/actions/runs/15049373926
- Previous run: 2025-05-05 "last week" https://github.com/agda/agda-categories/actions/runs/14849398474

A cache does not survive 10 days.

If you want, I can update CI to use setup-agda, please let me know.

JacquesCarette:
Agreed - and yes please!

andreasabel:
Closed in #468.

PR #468, opened 2025-05-15
Use setup-agda action in CI; simplify caching of _build

andreasabel:
Closes:
- #466 
- #467


andreasabel:
Agda now installs in 11sec.  So worst case CI runtime is cut by 50% now (from 40 min to 20 min).

JacquesCarette (approved):
Superb!

PR #478, opened 2025-06-30
Bimodule

tillrampe:
This pull request defines bimodules between monads and bimodulehomomorphisms (cf. Shulman - Framed Bicategories and Monoidal Fibrations, Ch. 11). This is part of a larger project constructing a bicategory of monads and bimodules.

### DIFF
--- a/src/Categories/Bicategory/Monad/Bimodule.agda
+++ b/src/Categories/Bicategory/Monad/Bimodule.agda
@@ -10,24 +10,22 @@ import Categories.Bicategory.Extras as Bicat
 open Bicat 𝒞
 
 record Bimodule (M₁ M₂ : Monad 𝒞) : Set (o ⊔ ℓ ⊔ e) where
-  open Monad M₁ renaming (C to C₁; T to T₁; μ to μ₁; η to η₁)
-  open Monad M₂ renaming (C to C₂; T to T₂; μ to μ₂; η to η₂)
-
+  open Monad
   field
-    F       : C₁ ⇒₁ C₂
-    actionˡ : F ∘₁ T₁ ⇒₂ F
-    actionʳ : T₂ ∘₁ F ⇒₂ F
+    F       : C M₁ ⇒₁ C M₂
+    actionˡ : F ∘₁ T M₁ ⇒₂ F
+    actionʳ : T M₂ ∘₁ F ⇒₂ F
 
-    assoc     : actionʳ ∘ᵥ (T₂ ▷ actionˡ) ∘ᵥ associator.from ≈ actionˡ ∘ᵥ (actionʳ ◁ T₁)
-    sym-assoc : actionˡ ∘ᵥ (actionʳ ◁ T₁) ∘ᵥ associator.to ≈ actionʳ ∘ᵥ (T₂ ▷ actionˡ)
+    assoc     : actionʳ ∘ᵥ (T M₂ ▷ actionˡ) ∘ᵥ associator.from ≈ actionˡ ∘ᵥ (actionʳ ◁ T M₁)
+    sym-assoc : actionˡ ∘ᵥ (actionʳ ◁ T M₁) ∘ᵥ associator.to ≈ actionʳ ∘ᵥ (T M₂ ▷ actionˡ)
 
-    assoc-actionˡ     : actionˡ ∘ᵥ (F ▷ μ₁) ∘ᵥ associator.from ≈ actionˡ ∘ᵥ (actionˡ ◁ T₁)
-    sym-assoc-actionˡ : actionˡ ∘ᵥ (actionˡ ◁ T₁) ∘ᵥ associator.to ≈ actionˡ ∘ᵥ (F ▷ μ₁)
-    assoc-actionʳ     : actionʳ ∘ᵥ (μ₂ ◁ F) ∘ᵥ associator.to ≈ actionʳ ∘ᵥ (T₂ ▷ actionʳ)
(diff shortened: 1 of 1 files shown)


src/Categories/Bicategory/Monad/Bimodule.agda
```diff
@@ -0,0 +1,99 @@
+{-# OPTIONS --without-K --safe --lossy-unification #-}
```
JacquesCarette, line 1: is `--lossy-unification` (still) crucial?

src/Categories/Bicategory/Monad/Bimodule.agda
```diff
@@ -0,0 +1,99 @@
...
+open Bicat 𝒞
+
+record Bimodule (M₁ M₂ : Monad 𝒞) : Set (o ⊔ ℓ ⊔ e) where
+  open Monad using (C; T; μ; η)
+  field
+    F       : C M₁ ⇒₁ C M₂
+    actionˡ : F ∘₁ T M₁ ⇒₂ F
+    actionʳ : T M₂ ∘₁ F ⇒₂ F
+
+    assoc     : actionʳ ∘ᵥ (T M₂ ▷ actionˡ) ∘ᵥ associator.from ≈ actionˡ ∘ᵥ (actionʳ ◁ T M₁)
```
JacquesCarette, line 19: Since you use `Extras` you might as well use `Shorthands` too.

src/Categories/Bicategory/Monad/Bimodule.agda
```diff
@@ -0,0 +1,99 @@
...
+    sym-assoc : actionˡ ∘ᵥ (actionʳ ◁ T M₁) ∘ᵥ associator.to ≈ actionʳ ∘ᵥ (T M₂ ▷ actionˡ)
```
JacquesCarette, line 20: Since you use this version that is "definitionally symmetric", it might be useful to have a `BimoduleHelper` that lets users elide the items that are derivable.

src/Categories/Bicategory/Monad/Bimodule.agda
```diff
@@ -0,0 +1,99 @@
...
+  ; sym-assoc-actionˡ = sym-assoc
+  ; assoc-actionʳ = sym-assoc
+  ; sym-assoc-actionʳ = assoc
+  ; identityˡ = identityˡ
+  ; identityʳ = identityʳ
+  }
+  where
+    open Monad M
+
+record Bimodulehomomorphism {M₁ M₂ : Monad 𝒞} (B₁ B₂ : Bimodule M₁ M₂) : Set (ℓ ⊔ e) where
```
JacquesCarette, line 47: I know that for smaller things, the objects and morphisms are all defined in the same file, but I think that for `Bimodule`, it would make sense to split this into `Bimodule/Homomorphism.agda`. It could also help make typechecking faster.

src/Categories/Bicategory/Monad/Bimodule.agda
```diff
@@ -0,0 +1,99 @@
...
+  where
+    open Monad using (C; T)
+    open Bimodule B using (actionˡ; actionʳ)
+    open hom.HomReasoning
+
+bimodule-hom-∘ : {M₁ M₂ : Monad 𝒞} → {B₁ B₂ B₃ : Bimodule M₁ M₂}
+               → Bimodulehomomorphism B₂ B₃ → Bimodulehomomorphism B₁ B₂ → Bimodulehomomorphism B₁ B₃
+bimodule-hom-∘ {M₁} {M₂} {B₁} {B₂} {B₃} g f = record
+  { α = α g ∘ᵥ α f
+  ; linearˡ = begin
```
JacquesCarette, line 78: Please move both of these proofs to the `where` clause. Having these inline will cause a lot of noise in downstream use, as Agda is too eager to eta-expand.

src/Categories/Bicategory/Monad/Bimodule.agda
```diff
@@ -0,0 +1,99 @@
...
+      actionˡ B₃ ∘ᵥ (α g ∘ᵥ α f) ◁ T M₁            ≈⟨ refl⟩∘⟨ ⟺ ∘ᵥ-distr-◁ ⟩
+      actionˡ B₃ ∘ᵥ (α g ◁ T M₁) ∘ᵥ (α f ◁ T M₁)   ≈⟨ sym-assoc₂ ⟩
```
JacquesCarette, line 80: for a future PR: this sure looks like bicategorical versions of `push` and `pull`. Should be abstracted out!

JacquesCarette (changes requested):

src/Categories/Bicategory/Monad/Bimodule.agda
tillrampe: What is the recommended way to get `push` and `pull` into scope?

JacquesCarette:
```agda
import Categories.Morphism.Reasoning as MR
open MR C
```
where C is the 1-category of interest. Is the 'vertical' category clearly visible here?

My comment here is that there should be a `Category.Bicategory.Reasoning` that serves the same purpose. Right now, there is a bunch of stuff in `Category.Bicategory.Extras` that probably belongs in such a (new) module.

tillrampe: There is a  private module `MR'` defined in `Category.Bicategory.Extras`. Do you want this but public?

JacquesCarette: Something like that, yes.

JacquesCarette (approved):
Wonderful!

PR #479, opened 2025-07-03
1-Category of Bimodules

tillrampe:
This pull request constructs a 1-category of bimodules and bimodule homomorphism given two monads in a bicategory (cf. Shulman - Framed Bicategories and Monoidal Fibrations, Ch. 11). This is part of a larger project constructing a bicategory of monads and bimodules.

### DIFF
--- a/src/Categories/Category/Construction/Bimodules/Properties.agda
+++ b/src/Categories/Category/Construction/Bimodules/Properties.agda
@@ -43,6 +43,7 @@ module Bimodulehom-isIso {B₁ B₂ : Obj {C = Bimodules}} (f : _⇒_ {C = Bimod
       ; linearʳ = linearʳ⁻¹
       }
     ; iso = record
+    -- Cannot be delta reduced because of size issues
       { isoˡ = IsIso.isoˡ αisIso
       ; isoʳ = IsIso.isoʳ αisIso
       }


src/Categories/Category/Construction/Bimodules/Properties.agda
tillrampe, line 1: The main purpose of this file is to show that a bimodule homomorphism is invertible if its underlying 2-cell is invertible. As this requires considering two different categories plus the ambient bicategory, there is a big potential for overloaded variable names. I'm employing using commands to avoid overloads, but don't rename. If you see a better way to strike a balance between being clear and being concise please let me know.

src/Categories/Category/Construction/Bimodules.agda
```diff
@@ -0,0 +1,35 @@
...
+  ; id = id-bimodule-hom
+  ; _∘_ = bimodule-hom-∘
+  ; assoc = assoc₂
+  ; sym-assoc = sym-assoc₂
+  ; identityˡ = identity₂ˡ
+  ; identityʳ = identity₂ʳ
+  ; identity² = identity₂²
+  ; equiv = record
+    { refl = hom.Equiv.refl
+    ; sym = hom.Equiv.sym
```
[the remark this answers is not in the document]
JacquesCarette: A one liner comment (like "must be delta expanded to typecheck as at different types") would be good.

src/Categories/Category/Construction/Bimodules/Properties.agda
tillrampe: I prefer mixfix notation for `_≅_` and `_⇒_`, and that is not possible unless you open the modules.

JacquesCarette (approved):
As the last thing discussed was quite cosmetic (in a way that is not covered by the library's style guide), I'm not going to hold up this PR because of that.

PR #490, opened 2025-10-03
Make CI fetch stdlib 2.3

Taneb:


JacquesCarette (dismissed):

JacquesCarette:
The reason this is failing is that @wenkokke 's setup-agda has not (yet) been updated to know agda-2.8.0 and stdlib-2.3 exist. I guess I'll work on that.

PR #491, opened 2025-10-30
Use glueTrianglesʳ in definition of _∘_ for slice categories

Taneb:
It has a comment saying it's basically composition in the over category, so it might as well be composition in the over category


Taneb:
@JacquesCarette I've rebased on master so it gets through CI

PR #492, opened 2025-11-03
Clean up monoidal functor properties

jacquescomeaux:
I was intending to improve typechecking time, but I think I ended up making more progress on the "lines of Haskell compiled to" metric.

The biggest improvements came from moving equational proofs out of records fields and into where blocks, but I also noticed that some of my changes (not sure which) to make the proofs more readable also reduced the size of the Haskell.

So far `Categories.Functor.Monoidal.Properties` is down from 242362 lines of Haskell to 95295 lines. I'm not quite done updating that file yet. I plan to update `Categories.Functor.Monoidal.Construction.Product` as part of this PR as well.

I also added opposite braided and symmetric monoidal categorites, and an `op` field to the bundled forms. This does make accessing the `op` of the underlying category ever so slightly harder.

I changed the definition of `flip-bifunctor` for better definitional equality, and added a way to get the oplax structure of a strong monoidal functor.

### DIFF
--- a/src/Categories/Category/Instance/Properties/Cats.agda
+++ b/src/Categories/Category/Instance/Properties/Cats.agda
@@ -16,11 +16,12 @@ open import Categories.Category.Monoidal.Instance.Cats
 open import Categories.Functor using (Functor; _∘F_) renaming (id to idF)
 open import Categories.Functor.Bifunctor using (Bifunctor)
 open import Categories.Functor.Construction.Constant using (constNat)
+import Categories.Morphism.Properties as MP
 import Categories.Morphism.Reasoning as MR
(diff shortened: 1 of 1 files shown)


jacquescomeaux:
Will update this soon. Do copatterns provide "enough insulation"? Or do the proofs still need to be extracted and given a name and type signature when using copatterns?

jacquescomeaux:
I just noticed the `id-iso` construct/lemma that I added to `Categories.Morphism.Properties` in a previous commit was already present in another module (in a less general form), so I removed the old one and saved a few lines of code.

I don't have any more changes to make for this PR, unless there is something specific that still needs to be addressed.

PR #495, opened 2025-11-14
Bump CI to agda-2.8 and stdlib-2.3

andreasabel:


JacquesCarette:
Thanks - I've been meaning to do this, but got stuck on Wen's `setup-agda` not having caught up.

JacquesCarette (approved):
looks good to me, but my YAML-fu is weak.

PR #502, opened 2025-12-12
Add Action monad

### LINKED ISSUE #498: Writer monad
Monad on Setoids constructed by pairing with an element of a monoid

Currently exists in `categories-examples` as `Categories.Examples.Monad.Duration`

Can be generalized to a monad on a monoidal category parameterised by a monoid object. Can be further generalized as a strong monad transformer (I think all monads on (Setoids, ×) are strong)

Taneb:
Closes #498

I'd like to share the commutative diagrams I made while trying to understand how to implement this:

Associativity: 
writer-assoc

Left identity:
writer-identityl

Right identity:
writer-identityr

In the issue I noted that this could be extended to a strong monad transformer - that comes from the fact that any pair of monads with a distributive property compose, which I think should be proven separately.

### DIFF
--- a/src/Categories/Monad/Construction/Action.agda
+++ b/src/Categories/Monad/Construction/Action.agda
@@ -8,8 +8,8 @@
 
 {-# OPTIONS --safe --without-K #-}
 
-open import Categories.Category
-open import Categories.Category.Monoidal
+open import Categories.Category.Core using (Category)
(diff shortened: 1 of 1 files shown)


JacquesCarette (changes requested):
Very nice. Some picky comments.

PR #503, opened 2026-01-06
Fix documentation typos for (0,1)-truncation

jacquescomeaux:
Minor typos. I think I corrected them in the right direction


JacquesCarette (approved):
Well-spotted, thank you.

PR #510, opened 2026-03-11
Add instances for rig categories

ricky136973:
Introduce Sets and PartialFunctions as rig categories.

### DIFF
--- a/src/Categories/Category/Monoidal/Instance/PartialFunctions.agda
+++ b/src/Categories/Category/Monoidal/Instance/PartialFunctions.agda
@@ -1,0 +1,492 @@
+{-# OPTIONS --without-K --safe #-}
+
+module Categories.Category.Monoidal.Instance.PartialFunctions where
+
+open import Data.Maybe using (Maybe; nothing; just; map; zip)
+open import Data.Maybe.Properties using (just-injective)
+open import Data.Product
+  using (_×_; _,_; proj₁; proj₂)
+  renaming (swap to swap×; assocʳ′ to assocʳ×; assocˡ′ to assocˡ×)
+open import Data.Sum
+  using (_⊎_; inj₁; inj₂; [_,_]′)
+  renaming (swap to swap⊎; assocʳ to assocʳ⊎; assocˡ to assocˡ⊎)
+open import Data.Unit using (⊤; tt)
+open import Data.Unit.Polymorphic using () renaming (⊤ to ⊤*; tt to tt*)
+open import Data.Empty.Polymorphic using () renaming (⊥ to ⊥*; ⊥-elim to ⊥*-elim)
+open import Relation.Binary.PropositionalEquality using (_≡_; refl; cong; cong₂; subst₂)
+open import Function using (_∘_; case_of_)
+open import Level
+
+open import Categories.Category.Instance.PartialFunctions using (PartialFunctions)
+open import Categories.Category.Monoidal using (Monoidal)
+open import Categories.Category.Monoidal.Braided using (Braided)
+open import Categories.Category.Monoidal.Symmetric using (Symmetric)
+open import Categories.Category.RigCategory using (RigCategory)
+open import Categories.Functor.Bifunctor using (Bifunctor)
+open import Categories.NaturalTransformation using (NaturalTransformation)
+open import Categories.NaturalTransformation.NaturalIsomorphism using (NaturalIsomorphism)
+
+module Product {o : Level} where
+  private
+    P = PartialFunctions o
+
+  module _ where
+    open Bifunctor
+
+    P⊗ : Bifunctor P P P
+
+    P⊗ .F₀ (X , Y)         = X × Y
+    P⊗ .F₁ (f , g) (x , y) = zip (f x) (g y)
+    P⊗ .identity   (x , y) = refl
+
+    P⊗ .homomorphism {f = f₁ , f₂} (x₁ , x₂)
+      with f₁ x₁ | f₂ x₂
+    ... | just y₁ | just y₂ = refl
+    ... | nothing | nothing = refl
+    P⊗ .homomorphism {g = g₁ , g₂} _ | just y₁ | nothing
+      with g₁ y₁
+    ... | just z₁ = refl
+    ... | nothing = refl
+    P⊗ .homomorphism {g = g₁ , g₂} _ | nothing | just y₂
+      with g₂ y₂
+    ... | just z₂ = refl
+    ... | nothing = refl
+
+    P⊗ .F-resp-≈ {f = f₁ , f₂} {g = g₁ , g₂} (f₁≗g₁ , f₂≗g₂) (x₁ , x₂)
+      with f₁ x₁ in y₁= | g₁ x₁ in z₁=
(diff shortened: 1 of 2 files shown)


src/Categories/Category/Monoidal/Instance/PartialFunctions.agda
```diff
@@ -0,0 +1,492 @@
...
+      { from = just ∘ assocʳ×
+      ; to   = just ∘ assocˡ×
+      ; iso  = record
+        { isoˡ = λ _ → refl
+        ; isoʳ = λ _ → refl
+        }
+      }
+
+    PM⊗ .unitorˡ-commute-from {f = f} (lift tt , x)
+      with f x
```
JacquesCarette: here I definitely would like to see this kind of reasoning abstracted out!

src/Categories/Category/Monoidal/Instance/PartialFunctions.agda
```diff
@@ -0,0 +1,492 @@
...
+
+  module _ where
+    open Bifunctor
+
+    P⊕ : Bifunctor P P P
+
+    P⊕ .F₀ (X , Y)          = X ⊎ Y
+    P⊕ .F₁ (f , g) (inj₁ x) = map inj₁ (f x)
+    P⊕ .F₁ (f , g) (inj₂ y) = map inj₂ (g y)
+    P⊕ .identity   (inj₁ x) = refl
```
JacquesCarette: I would be tempted to use properties of `[_,_]` instead of proving "from scratch".

src/Categories/Category/Monoidal/Instance/Sets.agda
```diff
@@ -69,3 +75,118 @@ module Coproduct {o : Level} where
...
+  Sets-Symmetric = Cocartesian.CocartesianSymmetricMonoidal.+-symmetric S Sets-is
+
+module Rig {o : Level} where
+  private
+    S = Sets o
+    module Sets⊎ = Coproduct
+    module Sets× = Product
+
+  Sets-Rig : RigCategory S Sets⊎.Sets-Symmetric Sets×.Sets-Symmetric
+  Sets-Rig = record
```
JacquesCarette: Sorry to ask this but could you use co-patterns here as well? I know it's not the same as the rest of the file, but it was written before co-patterns were a big thing.

Over time, even better would be to pull out the underlying properties and submit them to `agda-stdlib` and then re-use them here. But I won't block this PR for that!

JacquesCarette (commented):
Very nice! I want to explore some potential changes. But while I say 'Request changes', what I really mean is "let's discuss the design some more". This is definitely a great contribution.

TOTBWF:
I'd maybe rename `PartialFunctions` to something else. In constructive foundations, `A -> Maybe B` describes partial functions whose domain of definition has decidable membership.

src/Categories/Category/Monoidal/Instance/PartialFunctions.agda
```diff
@@ -0,0 +1,492 @@
...
+  private
+    P = PartialFunctions o
+
+  module _ where
+    open Bifunctor
+
+    P⊕ : Bifunctor P P P
+
+    P⊕ .F₀ (X , Y)          = X ⊎ Y
+    P⊕ .F₁ (f , g) (inj₁ x) = map inj₁ (f x)
```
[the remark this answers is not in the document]
ricky136973: Not exactly, but I simplified as `[ map inj₁ ∘ f , map inj₂ ∘ g ]′`. Haven't found a prettier way yet.

src/Categories/Category/Monoidal/Instance/PartialFunctions.agda
ricky136973: Revised as `trans (sym ([,]-∘ just x)) (cong just (map-id x))`. Could do an $\eta$-simplification by extracting the properties of extensional equality from the setoid, but I think it won't help much.

ricky136973:
> I'd maybe rename `PartialFunctions` to something else. In constructive foundations, `A -> Maybe B` describes partial functions whose domain of definition has decidable membership.

I know the more general notion of "partial functions" is $\sum_{P:X\to\Omega}(\sum_{x:X}P(x))\to Y$, and that's also the encoding used in 1lab. I still named it as `PartialFunctions` only because there is already a module `Categories.Category.Instance.PartialFunctions` defining the maybe-functions as a category, and I believe it is better to maintain some sort of consistency of naming here.

JacquesCarette:
I agree that consistency is very important. In the absence of other forces, I would always let consistency win.

In this case: I agree with @TOTBWF that this is a bad name. So I would like to find a better name -- and then *I* will change `Categories.Category.Instance.PartialFunctions` accordingly.

src/Categories/Category/Monoidal/Instance/PartialFunctions.agda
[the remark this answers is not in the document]
JacquesCarette: Do I have access to your fork? I think I would like to experiment with various definitions.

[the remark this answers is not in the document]
ricky136973: Access granted.

src/Categories/Category/Monoidal/Instance/PartialFunctions.agda
```diff
@@ -0,0 +1,492 @@
...
+    open Bifunctor
+
+    P⊗ : Bifunctor P P P
+
+    P⊗ .F₀ (X , Y)         = X × Y
+    P⊗ .F₁ (f , g) (x , y) = zip (f x) (g y)
+    P⊗ .identity   (x , y) = refl
+
+    P⊗ .homomorphism {f = f₁ , f₂} (x₁ , x₂)
+      with f₁ x₁ | f₂ x₂
```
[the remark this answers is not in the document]
ricky136973: I tried to pull the with-clauses out into a where-block, but Agda cannot infer the type of the helper functions if I do not provide the lengthy type signature explicitly.

src/Categories/Category/Monoidal/Instance/PartialFunctions.agda
ricky136973: Similar cases happened to the previous case. The more elegant way could have been like
```agda
PM⊗ .unitorˡ-commute-from {f = f} (lift tt , x) =
  maybe {B = λ fx → ?} (erefl ∘ just) refl (f x)
```
but Agda cannot infer `B`, and the explicit form of it is just as lengthy as in the previous case.

ricky136973:
> I agree that consistency is very important. In the absence of other forces, I would always let consistency win.
> 
> In this case: I agree with @TOTBWF that this is a bad name. So I would like to find a better name -- and then _I_ will change `Categories.Category.Instance.PartialFunctions` accordingly.

Name suggestion: `MaybeFunction`. `DecidablePartialFunction` is too long, `Maybe` is ambiguous.

JacquesCarette:
Apologies for being so slow on this - term is still keeping me down. Hoping to get to this soon.

`DecPartialFunctions` is less of a mouthful. @TOTBWF any other suggestions?

PR #513, opened 2026-04-19
change definition of exponentials

Reijix:
Exponentials are really hard to use at the moment, because [λg](https://agda.github.io/agda-categories/Categories.Object.Exponential.html#792) requires a product to be passed to it.
In CCCs there are helper definitions that alleviate this, but working with a single exponential object in a cartesian category is painful.

Instead, for B^A we should require that all binary products `-×A` exist (thats how exponentials are defined in wikipedia and nlab).

Also, I've moved the product requirement to be a parameter instead of a record field, so that when assuming an exponential object in a cartesian category, the products do not clash.

I've started adjusting the definitions, but since this is a bigger design decision, I'd like some feedback first before I proceed @JacquesCarette.

### DIFF
--- a/src/Categories/Category/BinaryCoproducts.agda
+++ b/src/Categories/Category/BinaryCoproducts.agda
@@ -68,6 +68,7 @@ record BinaryCoproducts : Set (levelOfTerm 𝒞) where
              ; assocˡ∘assocʳ to +-assocˡ∘+-assocʳ
              ; assocʳ∘assocˡ to +-assocʳ∘+-assocˡ
              ; swap         to +-swap
+             ; id⁂id        to id+₁id
              ; first        to +-first
              ; second       to +-second
              ; π₁∘⁂         to +₁∘i₁
--- a/src/Categories/Category/BinaryProducts.agda
+++ b/src/Categories/Category/BinaryProducts.agda
@@ -14,7 +14,7 @@ open HomReasoning
 
 open import Categories.Object.Product 𝒞
 open import Categories.Morphism 𝒞 using (_≅_; module Iso; Mono; Epi)
-open import Categories.Morphism.Reasoning 𝒞 using (pullʳ; pullˡ; elimʳ; cancelˡ; cancelʳ; introˡ; introʳ)
+open import Categories.Morphism.Reasoning 𝒞 using (pullʳ; pullˡ; elimʳ; cancelˡ; cancelʳ; introˡ; introʳ; id-comm)
 open import Categories.Category.Monoidal.Core using (Monoidal)
 
 open import Categories.Functor using (Functor) renaming (id to idF)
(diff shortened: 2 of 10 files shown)


JacquesCarette:
I don't have a strong opinion either way on this. I guess our definition might have been from older sources, and what you're suggesting is a more modern approach. I would be curious to know what 1lab and Max New's development do for this.

Most importantly: the precedent of wikipedia and nlab are enough for me to say "please continue".

Reijix:
> I don't have a strong opinion either way on this. I guess our definition might have been from older sources, and what you're suggesting is a more modern approach. I would be curious to know what 1lab and Max New's development do for this.
> 
> Most importantly: the precedent of wikipedia and nlab are enough for me to say "please continue".

In 1lab they actually require 𝒞 to be cartesian (https://1lab.dev/Cat.Diagram.Exponential.html).
I do think that might be sensible for us too, since definitions like [λ-distrib](https://agda.github.io/agda-categories/Categories.Object.Exponential.html#3212) are still very unwieldy if we only require "products with C" and requiring finite products makes all definitions in Exponential more readable.

What do you think @JacquesCarette?

JacquesCarette:
I would most definitely be happy with a definition of exponential objects defined under a Cartesian assumption -- maybe even as the default definition. We could move the current definition elsewhere, as a somewhat more "general" definition. Happy to kick around potential locations and names.

Reijix:
Hmm I'm not sure what to call the more general version.
We currently do have `Categories.Category.CartesianClosed.Canonical`
where `CartesianClosed` is defined using Exponentials that use finite products.

So one idea would be to just keep the current `Categories.Object.Exponential` and add `Categories.Object.Exponential.Canonical` that requires `Cartesian`. (And adding a comment to `Categories.Object.Exponential` that you probably want the other module in a cartesian category).

JacquesCarette:
(Sorry for the slow reply.) Yes, I like that suggestion!

Reijix:
@JacquesCarette 
I have now adjusted the PR:

- Added `Categories.Object.Exponential.Canonical`
- Use it to simplify `Categories.Category.CartesianClosed`
- Adjust places where CCC is used, most notably, `Category.CartesianClosed.Locally` was hard to adjust, but works now.
- Add a lemma for `id {A} ⁂ id {B} ≈ id` to `BinaryProducts` (and respectively `BinaryCoproducts`)

This simplifies `CartesianClosed` a lot, it removes the redundancy of having two definitions everywhere like `eval` and `eval'`.

Note that these changes make `Categories.Category.CartesianClosed.Canonical` obsolete. In a later PR this could be removed and all places using it could just be adjusted to use the normal CCC instance.

JacquesCarette (approved):
Very nice simplifications indeed!

PR #518, opened 2026-04-27
Refactor cocartesian to better mimic the structure of cartesian

Reijix:
I have implemented the suggestions by @TOTBWF in #509 to fix the issues with Agda inserting qualified names (like `coproducts.[_,_]`) into holes concerning coproducts, from my testing these changes have fixed the issues.

Furthermore, I have split up `Category.Cocartesian` into multiple files, to better mimic the structure of `Category.Cartesian`, for example I have added `Category.BinaryCoproducts` and `Category.Cocartesian.Monoidal`.

I have also adjusted `Category.Cartesian` to open `Terminal` and `BinaryProducts` `public`, as has already been the case in `Category.Cocartesian`.

Finally, I have renamed `!` in Initial to `¡` to better distinguish it from terminal objects, and next I will rename [_⁂_](https://agda.github.io/agda-categories/Categories.Category.BinaryProducts.html#1465) to `_×₁_` to be consistent with `Category.Cocartesian`

I have split my changes into multiple commits, since these are some bigger design decisions that I've changed.

The PR consists of three parts:
- [X] Refactor structure of `Category.Cocartesian`
- [X] Rename universal morphism of `Object.Initital`
- [x] Rename `_⁂_` to `_×₁_`

I can undo parts of it if you prefer.

PR #527, opened 2026-05-29
Rename universal morphism of initial object

Reijix:
As a successor to #521 I have renamed `!` in `IsInitial` to `¡`, since this is how it gets imported most of the time anyway.
This is part of a bigger effort started in #518 to make usage of products and coproducts more streamlined.


JacquesCarette (approved):
Let's see if downstream users scream because of this. But it is still the right thing to do. Thanks.

PR #524, opened 2026-05-17
More properties of first and second

sergey-goncharov:

PR #528, opened 2026-05-29
rename product morphism

Reijix:
As a successor to #521 I have renamed the product morphism `⁂` to `×₁` to make products and coproducts more alike.
This is part of a bigger effort started in https://github.com/agda/agda-categories/pull/518 to make usage of products and coproducts more streamlined.


JacquesCarette:
Can you please resolve the merge conflict?

I'm going to have to think harder on this one. I'm not sure I like the new name -- it especially interacts badly with names that had superscripts before. I do like the general idea. Do you have other suggestions for a name?

Reijix:
To me `×₁` seems to be the only intuitive choice without renaming `+₁` as well.

If I see it correctly, the only place where superscripts are added is `Categories.Category.Product`.
Maybe it was a mistake to rename the functor product to `×₁` in the first place. It does make sense to choose different names here to distinguish the functor product.
So my suggestion would be to keep the functor product as `⁂`.

Reijix:
I have reverted the name of the product functor to `⁂` now, drastically reducing the size of this PR.
Also I've marked the PR as a draft for now, since I didn't re-align proofs yet.

@JacquesCarette if you are still not happy with the new name, I don't mind dropping this PR. I don't want to suggest too many breaking changes.

JacquesCarette:
The merge of a later PR now is causing conflicts here - could you look into that please?

JacquesCarette (dismissed):
I think I'm fine with the remaining renames.

Reijix:
From my end, the PR is ready now.
I have finalized it by aligning all proofs that were misaligned due to the renaming.

Furthermore, I have adjusted the imports in `Category.BinaryCoproducts`:
- changed up the order a little bit, so that the order of the imports reflects the order of the lemmas in `Category.BinaryProducts` (this makes it easier to check that everything that is defined on products, is dualized).
- added the lemmas introduced in #524 to the import list.

JacquesCarette (approved):

PR #533, opened 2026-07-01
Add IsoComma module for "Pullback of functors"

tetrapharmakon:
There doesn't seem to be a module that implement this construction yet.


JacquesCarette (approved):
Nice, thanks.

PR #534, opened 2026-07-02
Add monoidal category coherence solver

LionOfJewdah:
## Summary

This PR adds a monoidal-category coherence solver and a small set of supporting monoidal/braided utilities.

The solver provides:

- `Categories.Tactic.Monoidal.Core`: free structural syntax for monoidal objects and morphisms, normal forms, interpretation, and object and morphism-level coherence for interpreted structural maps.
- `Categories.Tactic.Monoidal`: reflection macros `solve₀` and `solveMonoidal`.
- `Categories.Tactic.Monoidal.Examples`: examples covering object-level and morphism-level use.

It also promotes reusable helper lemmas for monoidal reasoning.

## Motivation

In a personal project, I am working with categories of matrices over commutative rings, especially the complex numbers constructed as the Gaussian extension of the Cauchy reals. In that setting, the matrix category over `ℂ` has a dagger compact structure and is a skeleton of `FdHilb`.

For that development, a large amount of proof work is structural monoidal bookkeeping: associators, unitors, tensor functoriality, and later braidings/symmetries. This PR factors out the plain monoidal coherence layer so those later constructions can avoid carrying long hand-written reassociation proofs.

The motivating trace construction is concrete. In a matrix category, the trace of an endomorphism is the usual sum of diagonal entries. In my downstream project, I have proved that this sum-of-diagonals operation is exactly the Joyal-Street-Verity trace induced by the compact closed structure. This is the standard categorical picture: compact closed categories carry a traced monoidal structure, and the categorical trace recovers the usual matrix trace in finite-dimensional vector spaces and, more generally, free modules over a commutative ring.

This PR is intentionally limited to plain monoidal coherence plus small reusable braided utilities. Follow-up PRs will add braided and symmetric monoidal coherence solvers on top, and then use those solvers in a construction of a traced monoidal structure from compact closed structure.

## Possible Related Contributions

If there is interest, I can also contribute the matrix-category constructions from the downstream project. These include matrix categories over commutative rings and functoriality in the ring: a ring homomorphism `R → S` induces a functor `MatCat R → MatCat S`, which is full when the homomorphism is surjective and faithful when it is injective, with the expected algebraic refinements.

That would be a separate contribution, since it would also require upstreaming the supporting matrix algebra, either to the standard library or to `agda-categories`.

## Follow-Up Work

Planned follow-ups include:

- a braided monoidal coherence solver;
- a symmetric monoidal coherence solver;
- compact-closed helper APIs, including lemmas/accessors showing that in a symmetric monoidal category, left rigidity gives right rigidity and right rigidity gives left rigidity;
- using those APIs to construct a traced monoidal structure from compact closed structure.

## References

- André Joyal, Ross Street, and Dominic Verity, *Traced monoidal categories*, Mathematical Proceedings of the Cambridge Philosophical Society 119:447-468, 1996.
- Peter Selinger, *A survey of graphical languages for monoidal categories*, especially §4.8 on compact closed categories and §5.7 on symmetric traced categories: https://arxiv.org/abs/0908.3347
- Kate Ponto and Michael Shulman, *Traces in symmetric monoidal categories*, especially Examples 3.1-3.2: https://arxiv.org/abs/1107.6032

## Testing

Checked with:

```sh
agda -Werror -i. -isrc --safe --without-K src/Categories/Tactic/Monoidal.agda
agda -Werror -i. -isrc --safe --without-K src/Categories/Tactic/Monoidal/Examples.agda
agda -Werror -i. -isrc --safe --without-K src/Categories/Category/Monoidal/Braided/Properties.agda
agda -Werror -i. -isrc --safe --without-K src/Categories/Category/Monoidal/Interchange/Braided.agda
agda -Werror -i. -isrc --safe --without-K src/Categories/Category/Monoidal/Symmetric/Properties.agda
```


LionOfJewdah:
I wanted to provide a single PR to construct the JSV trace from a compact closed structure but I ended up needing some 3,500 lines of code for that before making the monoidal (and braided/symmetric) coherence solvers, so I thought it would be both easier, more useful, more helpful, and more reusable to build and contribute the solvers as well

Regarding MatCat: It was a fun exercise to prove it's a SemiringCategory/RigCategory on the Kronecker tensor product and Matrix direct sum. I chose objects as the natural numbers and matrices a small record wrapping Fin n -> Fin m -> (Ring.Carrier); this means the tensor functor structures are

```agda
⊗₀ = _*ℕ_ -- on dimensions
⊗₁ = Kronecker product
  I  = 1 -- (Mat 1 1 ≅ R)
  
 ⊗₀ = _+ℕ_ -- on dimensions    
 ⊗₁ = direct sum
  I  = 0 -- (Mat 0 0 is the empty matrix)
```

src/Categories/Category/Monoidal/Symmetric/Properties.agda
```diff
@@ -19,7 +19,7 @@ open Symmetric SM
 
 -- Shorthands for the braiding
 
-open BraidedProperties braided public using (module Shorthands)
+open BraidedProperties braided public using (Obj-⊗-Comm-Monoid; module Shorthands)
```
JacquesCarette: Why add this to the import if it is not needed?

src/Categories/Category/Monoidal/Coherence.agda
```diff
@@ -0,0 +1,206 @@
...
+open TensorIdentity using (id⊗id; whisker-∘; ⊗id-∘)
+open import Categories.Morphism.Reasoning C
+import Categories.Category.Monoidal.Properties M as MonoidalProps
+open MonoidalProps.Kelly's using (coherence₂)
+open import Categories.Tactic.Monoidal using (module Finite)
+open Finite M using (module Solver)
+
+open MonUtil.Shorthands
+
+cancel-middle : ∀ {A B C D}
```
JacquesCarette: This already exists in the reasoning modules.

src/Categories/Category/Monoidal/Coherence.agda
```diff
@@ -0,0 +1,206 @@
...
+
+α-slide : ∀ {A P Q Z} {h : P ⇒ Q} →
+  (id {A} ⊗₁ (h ⊗₁ id {Z})) ∘ α⇒ {A} {P} {Z}
+  ≈ α⇒ {A} {Q} {Z} ∘ ((id {A} ⊗₁ h) ⊗₁ id {Z})
+α-slide = ⟺ assoc-commute-from
+
+α-sweep : ∀ {Y B P Q} {k : P ⇒ Q} →
+  α⇒ {Y} {B} {Q} ∘ (id {Y ⊗₀ B} ⊗₁ k)
+    ≈ (id {Y} ⊗₁ (id {B} ⊗₁ k)) ∘ α⇒ {Y} {B} {P}
+α-sweep {Y} {B} {k = k} =
```
JacquesCarette: We heavily prefer equational reasoning over 'inline' like this.

src/Categories/Category/Monoidal/Coherence.agda
```diff
@@ -0,0 +1,206 @@
...
+  assoc
+  ○ (refl⟩∘⟨ (sym-assoc ○ (from∘to ⟩∘⟨refl) ○ identityˡ))
+
+α-slide : ∀ {A P Q Z} {h : P ⇒ Q} →
+  (id {A} ⊗₁ (h ⊗₁ id {Z})) ∘ α⇒ {A} {P} {Z}
+  ≈ α⇒ {A} {Q} {Z} ∘ ((id {A} ⊗₁ h) ⊗₁ id {Z})
+α-slide = ⟺ assoc-commute-from
+
+α-sweep : ∀ {Y B P Q} {k : P ⇒ Q} →
+  α⇒ {Y} {B} {Q} ∘ (id {Y ⊗₀ B} ⊗₁ k)
```
JacquesCarette: We don't annotate the implicits unless completely necessary (here and everywhere else in this file / PR)

src/Categories/Category/Monoidal/Coherence.agda
```diff
@@ -0,0 +1,206 @@
...
+    α⇒ {A} {B ⊗₀ C} {D ⊗₀ E}
+      ∘ (α⇒ {A} {B} {C} ⊗₁ id {D ⊗₀ E})
+      ∘ α⇒ {(A ⊗₀ B) ⊗₀ C} {D} {E}
+  ≈ (id {A} ⊗₁ α⇐ {B} {C} {D ⊗₀ E})
+      ∘ α⇒ {A} {B} {C ⊗₀ (D ⊗₀ E)}
+      ∘ (id {A ⊗₀ B} ⊗₁ α⇒ {C} {D} {E})
+      ∘ α⇒ {A ⊗₀ B} {C ⊗₀ D} {E}
+      ∘ (α⇒ {A ⊗₀ B} {C} {D} ⊗₁ id {E})
+assoc-pair-coherence {A} {B} {C} {D} {E} =
+  solve ((((x0 ⊗o x1) ⊗o x2) ⊗o x3) ⊗o x4)
```
JacquesCarette: For a variety of reasons, we don't use solvers internally. If a lemma is worth having, it is worth having a full proof that can be inspected.

src/Categories/Category/Monoidal/Reasoning.agda
```diff
@@ -55,6 +55,17 @@ _⟩⊗⟨refl = ⊗-resp-≈ˡ
 ⊗-distrib-over-∘ : ((f ∘ h) ⊗₁ (g ∘ i)) ≈ ((f ⊗₁ g) ∘ (h ⊗₁ i))
 ⊗-distrib-over-∘ = homomorphism
 
+⊗-factor-over-∘ : ((f ⊗₁ g) ∘ (h ⊗₁ i)) ≈ ((f ∘ h) ⊗₁ (g ∘ i))
```
JacquesCarette: Is this really worth naming?

src/Categories/Category/Monoidal/Reasoning.agda
```diff
@@ -134,6 +145,19 @@ split₂ˡ {f = f} {g} {h} = begin
   (id ∘ f) ⊗₁ (g ∘ h)   ≈⟨ ⊗-distrib-over-∘ ⟩
   id ⊗₁ g ∘ f ⊗₁ h      ∎
 
+module TensorIdentity where
+
+  id⊗id : ∀ {A B} → id {A} ⊗₁ id {B} ≈ id
```
JacquesCarette: Feels like this already exists somewhere else?

src/Categories/Category/Monoidal/Reasoning.agda
```diff
@@ -134,6 +145,19 @@ split₂ˡ {f = f} {g} {h} = begin
...
+  id⊗id = Functor.identity ⊗
+
+  whisker-∘ : ∀ {Y P Q Z} {g : Q ⇒ Z} {h : P ⇒ Q} →
```
JacquesCarette: Is this worth naming?

src/Categories/Tactic/Monoidal/Coherence.agda
```diff
@@ -0,0 +1,357 @@
...
+  renaming
+    ( _⇒_ to _⊸_ ; _∘_ to _⊚_ ; _⊗₁_ to _⊗ˢ_
+    ; α⇒  to α⇒ᶠ ; α⇐  to α⇐ᶠ ;  λ⇒  to λ⇒ᶠ ; λ⇐ to λ⇐ᶠ ; ρ⇒ to ρ⇒ᶠ ; ρ⇐ to ρ⇐ᶠ
+    )
+open NormalForm using (assocₙ; assocₙ⁻¹; unitʳₙ; unitʳₙ⁻¹)
+open Kelly's using (coherence₃)
+
+private
+  -- Merge two identity-on-the-left tensors.
+  ⊗id-∘ : ∀ {A B C D} {f : C ⇒ D} {g : B ⇒ C}
```
JacquesCarette: Wasn't this also created in one of the other files

src/Categories/Tactic/Monoidal/Coherence.agda
```diff
@@ -0,0 +1,357 @@
...
+  μ-assocₙ A B C = μ-assoc (nf A) (nf B) (nf C)
+
+-- The co-strictifier: eval (nf X) ⇒ ⟦ X ⟧₀ (the `.to` of `strictify`). Its tensor
+-- case exposes `μ⇒`, so the associator coherence lands on `μ-assoc`.
+κ⁻¹ : (X : Ob) → eval (nf X) ⇒ ⟦ X ⟧₀
+κ⁻¹ X = _≅_.to (strictify X)
+
+-- coercions compose
+substₑ-∘ : {x y z : List Atom} (p : x ≡ y) (q : y ≡ z)
+  → substₑ (Relation.Binary.PropositionalEquality.trans p q) ≈ substₑ q ∘ substₑ p
```
JacquesCarette: Best to import qualified, so as to use a shorter name here.

src/Categories/Tactic/Monoidal/Coherence.agda
```diff
@@ -0,0 +1,357 @@
...
+  ρ⇐ ∘ κ⁻¹ X
+    ∎)
+
+substₑ-loop-refl : {w : List Atom} {p : w ≡ w} → p ≡ refl → substₑ p ≈ id
+substₑ-loop-refl p≡refl = Equiv.reflexive (cong substₑ p≡refl)
+
+loop-trivial-from : ∀ {X} (h : X ⊸ X) → substₑ (⇒⇒nf h) ≈ id → ⟦ h ⟧₁ ≈ id
+loop-trivial-from {X} h loop = begin
+  ⟦ h ⟧₁                          ≈˘⟨ identityʳ ⟩
+  ⟦ h ⟧₁ ∘ id                     ≈˘⟨ refl⟩∘⟨ _≅_.isoˡ (strictify X) ⟩
```
JacquesCarette: you can do this in 1 step with `intro`

src/Categories/Tactic/Monoidal/Coherence.agda
```diff
@@ -0,0 +1,357 @@
...
+  ⟦ h ⟧₁ ∘ (κ⁻¹ X ∘ _≅_.from (strictify X))  ≈⟨ sym-assoc ⟩
+  (⟦ h ⟧₁ ∘ κ⁻¹ X) ∘ _≅_.from (strictify X)
+    ≈⟨ (Equiv.trans (κ⁻¹-natural h)
```
JacquesCarette: Don't use `trans` in an equational proof

src/Categories/Tactic/Monoidal/Core.agda
```diff
@@ -0,0 +1,192 @@
...
+-- §1: The interpretation monoid homomorphism
+-- ----------------------------------------------------------------------------
+--
+-- `eval` sends a normal form (a word in the free monoid) to the canonical
+-- right-nested object; it is the unique monoid homomorphism extending
+-- `⟦_⟧ₐ`. Its multiplicativity `eval-homo` needs only the associator and left
+-- unitor — no pentagon — because a monoid homomorphism records the
+-- multiplicativity square, not its coherence.
+
+eval : List Atom → Obj
```
JacquesCarette, line 77: This is `foldr` -- please use taht.

src/Categories/Tactic/Monoidal/Core.agda
```diff
@@ -0,0 +1,192 @@
...
+eval []       = unit
+eval (x ∷ xs) = ⟦ x ⟧ₐ ⊗₀ eval xs
+
+eval-homo : (x y : List Atom) → eval (x ++ y) ≅ eval x ⊗₀ eval y
```
JacquesCarette: and this should be a corollary of the Monoid structure that you've already defined.

src/Categories/Tactic/Monoidal/Examples.agda
```diff
@@ -0,0 +1,104 @@
...
+open import Categories.Tactic.Monoidal.Core M ⟦_⟧ₐ
+open import Categories.Tactic.Monoidal using (solve; solve₀; solveMonoidal)
+
+open import Categories.Morphism C using (_≅_)
+open Category C hiding (_⇒_)
+open Monoidal M hiding (⊗) renaming (_⊗₁_ to _⊗₁ᶜ_)
+open Free Atom renaming (_∘_ to _∘ᶠ_)
+
+-- ----------------------------------------------------------------------------
+-- §1: Object-level coherence — the canonical isomorphism, for free
```
JacquesCarette: These should all be `private` since these are tests and not things to be exported.

src/Categories/Tactic/Monoidal.agda
```diff
@@ -0,0 +1,196 @@
...
+        )
+
+    private
+      module Coherence′ = Coherence.WithUIP M atom
+        (Decidable⇒UIP.≡-irrelevant (≡-dec-list _≟Fin_))
+
+    solveᶠ : ∀ {X Y} (f g : X ⇒ᶠ Y) → Sem.⟦_⟧₁ f ≈ Sem.⟦_⟧₁ g
+    solveᶠ = Coherence′.coherence-UIP
+
+  module Solver (A B C D E : Obj) where
```
JacquesCarette: Exactly 5 objects !?!?!

src/Categories/Tactic/Monoidal.agda
```diff
@@ -0,0 +1,196 @@
...
+  lastVisible (vArg x ∷ args) = maybe just (just x) (lastVisible args)
+  lastVisible (_ ∷ args)      = lastVisible args
+
+  getPath : Term → Maybe Term
+  getPath (def _ args) = lastVisible args
+  getPath (con _ args) = lastVisible args
+  getPath _            = nothing
+
+  getPaths : Term → Maybe (Term × Term)
+  getPaths goal with getArgs goal
```
JacquesCarette: use `maybe` here instead of `with`

JacquesCarette (changes requested):
On top of the requested changes, I will need you to
1. cite where the algorithm comes from
2. certify that no AI was used in the creation of this code.

LionOfJewdah:
Hi Jacques, thanks for the careful review. I’ll work through the requested changes.

The algorithm is the standard normal-form proof of Mac Lane coherence for monoidal categories: formal tensor expressions are normalized to words in the free strict monoidal category, with tensor interpreted as concatenation and the unit as the empty word, and structural morphisms are compared through the induced equality of those normal forms. I saw it first at https://en.wikipedia.org/wiki/Monoidal_category#Free_strict_monoidal_category. I’ll add citations to:

- Saunders Mac Lane, “Natural Associativity and Commutativity”, Rice University Studies 49, 1963, pp. 28-46.
- Saunders Mac Lane, Categories for the Working Mathematician, 2nd ed., Springer, 1998, Chapter VII.
- G. M. Kelly, “On MacLane’s Conditions for Coherence of Natural Associativities, Commutativities, etc.”, Journal of Algebra 1, 1964, pp. 397-402.

On AI: I should have checked the project policy first. I cannot honestly certify that no AI was used for the current branch: I designed and implemented the algorithm, but I did use Codex while working through typechecking and cleanup. I apologize for that.

I will not use LLM assistance for future contributions to Agda projects. If this makes the current PR unacceptable, I understand; I can close it and only come back with a manually prepared contribution. For that work I would use ordinary tooling only, such as Agda itself, editor commands, and CI/typechecking scripts, not LLM assistance. I have seen [agda/agda#8507](https://github.com/agda/agda/pull/8507) which discusses this.

src/Categories/Tactic/Monoidal/Core.agda
LionOfJewdah: Good point. I found something even more directly DRY-flavored: [`Data.List.Effectful.Foldable`](https://github.com/agda/agda-stdlib/blob/master/src/Data/List/Effectful/Foldable.agda), so when I update this branch it will have that

LionOfJewdah:
Hi Jacques, I reworked the PR manually and made it more readable and streamlined. I dropped some things that could go in a follow up PR but weren't used directly here, and I think my planned contributions of rigid and compact closed category lemmas and the trace construction on a compact closed category are more valuable than a braided or symmetric coherence solver.

I hope these contributions help :).

An aside: I'm pleased to see that the standard library is going [back to without-K](https://github.com/agda/agda-stdlib/pull/2967). Based on what you told me, I reverted my own project from cubical-compatible to without-K to save typechecking time, especially since I had no need for HoTT interoperability.

JacquesCarette:
I will accept that you take personal responsibility for all the code in this PR and not reject it out-of-hand for AI use.

src/Categories/Tactic/Monoidal/Core.agda
```diff
@@ -0,0 +1,461 @@
...
+-- The morphism-level theorem is the standard normal-form proof of Mac Lane
+-- coherence. See Mac Lane, "Natural Associativity and Commutativity", Rice
+-- University Studies 49 (1963), pp. 28-46, Categories for the Working
+-- Mathematician, Chapter VII; and Kelly, "On MacLane's Conditions for Coherence
+-- of Natural Associativities, Commutativities, etc.", Journal of Algebra 1
+-- (1964), pp. 397-402.
+--------------------------------------------------------------------------------
+
+open import Level using (Level; _⊔_)
+open import Categories.Category using (Category)
```
JacquesCarette: `Categories.Category.Core`

LionOfJewdah:
Thank you Jacques. I take full personal responsibility for all code in this and any future PRs, as well as my one previous one that added Obj-⊗-Monoid. I also learned better Agda style and idioms from this

src/Categories/Tactic/Monoidal/Core.agda
```diff
@@ -0,0 +1,461 @@
...
+-- `⟦_⟧₀` interprets a full object (with bracketing and units); `⟦ ⌜ w ⌝ ⟧₀`
+-- coincides definitionally with `eval w`. `strictify` is the canonical iso
+-- reassociating any object to (the image of) its normal form.
+
+⟦_⟧₀ : Ob → Obj
+⟦ ‹ x › ⟧₀ = ⟦ x ⟧ₐ
+⟦ I ⟧₀     = unit
+⟦ X ⊗ Y ⟧₀ = ⟦ X ⟧₀ ⊗₀ ⟦ Y ⟧₀
+
+strictify : (X : Ob) → ⟦ X ⟧₀ ≅ eval (nf X)
```
JacquesCarette: I think `strictify`'s statement should go in the other direction. The fact that the proofs involve a lot of `sym` is an indication of that as well.

src/Categories/Tactic/Monoidal/Core.agda
```diff
@@ -0,0 +1,461 @@
...
+strictify (X ⊗ Y) =
+  strictify X ⊗ᵢ strictify Y  ≅⟨ eval (nf X) ⊗₀ eval (nf Y) ⟩
+  ≅.sym (eval-homo (nf X) (nf Y))
+
+-- Object-level coherence: any two objects with the same normal form are
+-- canonically isomorphic.  No Mac Lane theorem is needed here; this is the free
+-- monoid interpretation transported along `strictify`.
+object-coherence : nf X ≡ nf Y → ⟦ X ⟧₀ ≅ ⟦ Y ⟧₀
+object-coherence {X} {Y} p =
+  strictify X                  ≅⟨ eval (nf X) ⟩
```
JacquesCarette: This proof is extremely weird -- it feels like it is exactly inside-out, with the justification on the left and the intermediate steps on the right. What is going on here!?!?

src/Categories/Tactic/Monoidal/Core.agda
```diff
@@ -0,0 +1,461 @@
...
+-- ----------------------------------------------------------------------------
+-- §5: Naturality of strictification
+-- ----------------------------------------------------------------------------
+
+private
+  assoc-shuffle
+    : α⇒ {P ⊗₀ Q} {R} {S} ∘ (α⇐ {P} {Q} {R} ⊗₁ id) ∘ α⇐ {P} {Q ⊗₀ R} {S}
+      ≈ α⇐ {P} {Q} {R ⊗₀ S} ∘ (id ⊗₁ α⇒ {Q} {R} {S})
+  assoc-shuffle = begin
+    α⇒ ∘ (α⇐ ⊗₁ id) ∘ α⇐                                 ≈˘⟨ refl⟩∘⟨ identityʳ ⟩
```
JacquesCarette: Given that the first 5 lines don't touch the `α⇒` at all, should pull out a sub-lemma in a `where` clause. Then I think the appropriate push/pull to use will become even more visible.

src/Categories/Tactic/Monoidal/Core.agda
```diff
@@ -0,0 +1,461 @@
...
+substₑ-∘ : {x y z : List Atom} (p : x ≡ y) (q : y ≡ z)
+  → substₑ (≡.trans p q) ≈ substₑ q ∘ substₑ p
+substₑ-∘ refl q = ⟺ identityʳ
+
+-- naturality of `μ⇒` along coercions
+μ-natural : {x x' y y' : List Atom} (p : x ≡ x') (q : y ≡ y')
+  → (substₑ p ⊗₁ substₑ q) ∘ μ⇒ x y ≈ μ⇒ x' y' ∘ substₑ (cong₂ _++_ p q)
+μ-natural {x = x} {y = y} refl refl = begin
+  (id ⊗₁ id) ∘ μ⇒ x y   ≈⟨ ⊗.identity ⟩∘⟨refl ⟩
+  id ∘ μ⇒ x y           ≈⟨ identityˡ ⟩
```
JacquesCarette: `id-comm` or `id-comm-sym` will do this in one step.

src/Categories/Tactic/Monoidal/Examples.agda
```diff
@@ -0,0 +1,67 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Level using (Level)
+open import Categories.Category using (Category)
+open import Categories.Category.Monoidal using (Monoidal)
+
+module Categories.Tactic.Monoidal.Examples
```
JacquesCarette, line 7: out of curiosity, what is the size of the `.agdai` file for this module, given that it basically has no content?

JacquesCarette (changes requested):
One more round of changes needed. There will be more, but it is useful to be incremental.

src/Categories/Tactic/Monoidal/Examples.agda
LionOfJewdah: On my Mac, _build/2.8.0/agda/src/Categories/Tactic/Monoidal/Examples.agdai, size 78,079 bytes, about 76 KiB.

JacquesCarette: I guess it must still store a bunch of stuff - but it's not too bad.

LionOfJewdah:
I think Data.List and its foldMap are the left adjoint to the forgetful functor from monoids to sets (or setoids, with some congruence lemmas). If desired, I could package the construction reusably as an auction between categories of monoids and set(oid)s at the appropriate level. I would prefer to do that in a follow up PR to do let this one remain focused, but the ingredients are here. I think I just need a foldMap composition lemma and naturality plumbing

JacquesCarette:
> I think Data.List and its foldMap are the left adjoint to the forgetful functor from monoids to sets

They are indeed. I even have two full proofs of that adjointness (in two different repos!) that I never managed to find time to submit a PR for. Definitely should be done in a separate PR.

src/Categories/Tactic/Monoidal/Core.agda
```diff
@@ -102,21 +102,21 @@ eval-homo = IsMonoidHomomorphism.homo eval-homomorphism
...
 
-strictify : (X : Ob) → ⟦ X ⟧₀ ≅ eval (nf X)
-strictify ‹ x ›   = ≅.sym unitorʳ
+strictify : (X : Ob) → eval (nf X) ≅ ⟦ X ⟧₀
+strictify ‹ x ›   = unitorʳ
 strictify I       = ≅.refl
 strictify (X ⊗ Y) =
-  strictify X ⊗ᵢ strictify Y  ≅⟨ eval (nf X) ⊗₀ eval (nf Y) ⟩
-  ≅.sym (eval-homo (nf X) (nf Y))
+  eval-homo (nf X) (nf Y)      ≅⟨ eval (nf X) ⊗₀ eval (nf Y) ⟩
```
JacquesCarette: This equational reasoning still seems inside out? Same with `object-coherence` below? Where does the `_≅⟨_⟩_` combinator come from?

JacquesCarette (changes requested):
Converging.

Now, for subst: take a look at [Relation.Binary.PropositionalEquality.Subst.Properties](https://agda.github.io/agda-categories/Relation.Binary.PropositionalEquality.Subst.Properties.html) in stdlib, especially `TransportOverQ`. It is used already (in particular in `Data.Quiver.Morphism`) for similar things.

src/Categories/Tactic/Monoidal/Core.agda
LionOfJewdah: Ah, I got that from `Categories.Category.Construction.Core.`, module `Shorthands` and `Commutationᵢ`. I wasn't as familiar with manipulating isomorphisms of objects as equivalences of morphisms, so when I found that (`_≅⟨_⟩_`; `connectᵢ` syntax) I used it. I changed it to ≅.trans chains inline -- I can try to present it in a different way if preferred.

LionOfJewdah: oh whoops I missed the `TransportOverQ` part -- will do that next

LionOfJewdah: Thanks — I've made a cleanup pass on Tactic/Monoidal/Core.agda.

The two `κ⁻¹-natural` cases you objected to (`α⇐ᶠ`, `ρ⇐ᶠ`) were doing the same thing twice, inside-out, via switch-fromtoˡ. I've factored the shared content into one lemma — naturality at an inverse generator follows from naturality at the generator by inverting the interpretation on the left and the coercion on the right — so both cases are now one line. That also killed `assocₑ⁻¹`, `unitʳₑ⁻¹`, and their two `isoʳ` lemmas. loop-trivial similarly lost two intermediate definitions that each had exactly one use site.

On `Relation.Binary.PropositionalEquality.Subst.Properties`: I tried it three ways and each came out worse to read, so I left it out. I think the reason is structural. In `Data.Quiver.Morphism` the transports are the subject — quivers have no composition, so `◂`/`▸` and their coherences are the only algebra available. Here the subject is composition in `𝒞`, and the coercions `substₑ p` are incidental: every one of them is discharged by matching `p` against `refl`, at which point `substₑ refl` is definitionally `id`. Routing through Transport/`TransportOverQ` costs a `cong eval` at each site to lift word-equalities to object-equalities, plus a `∘-vs-transport` bridge, and turns one-line `refl` matches into multi-step ≡-chains. Happy to be shown otherwise if you had a particular use in mind — I can push the attempts to a branch if it's useful to look at.

I will push my cleanup in a minute or two once `make test` returns green. I created a new file, `Categories.Category.Monoidal.Reassociation`, for reusable reassociation lemmas in monoidal categories. I made it its own file because the compact-close trace construction branch I'm working on uses it quite a bit, though with a lot more lemmas that weren't necessary for this PR

JacquesCarette (approved):
I think this is ready to go now.

PR #541, opened 2026-07-23
Add factorization structures for morphisms

t-wissmann:
This introduces the ℰ,ℳ-factorizations of morphisms in a category, following the definitions of the Adámek, Herrlich, Strecker book ("Joy of cats"). The record FactorizationStructure is Definition 14.1 of that book, together with the properties that the morphism classes ℰ and ℳ are each closed under morphism equality _≈_.

Beside the definition, the following concepts and results are added:

* Categories.Morphism.Lifts:
    * A new record `Diagonal`, which is like `Filler` but does not depend on a commutativity proof.
    * `MorphismClass` inclusion `_⊆_` and closure under morphism equality
    * A new record `MorphismClassMember` that allows a uniform treatment of morphisms in Epi, Mono, etc
* Categories.Morphism.FactorizationStructure.Core: The main definitions for a category 𝒞 and given morphism classes ℰ and ℳ
* Categories.Morphism.FactorizationStructure: The property of a category 𝒞 being [ ℰ , ℳ ]-structured
* Categories.Morphism.FactorizationStructure.Properties: Proofs of ℰ ∩ ℳ = Iso, ℰ∘ℰ⊆ℰ, ℳ∘ℳ⊆ℳ and that ℰ and ℳ determine each other by diagonalization.
* Categories.Morphism.FactorizationStructure.Duality: If 𝒞 is [ ℰ , ℳ ]-structured, then 𝒞ᵒᵖ is [ ℳ , ℰ ]-structured.
* Categories.Morphism.FactorizationStructure.Instance.Setoids: The category Setoids c ℓ is [ Surj , Inj ]-structured, where Surj resp. Inj are the classes of surjections resp. injections.

PR #542, opened 2026-07-24
Add the definition of factorization systems

t-wissmann:
This introduces the ℰ,ℳ-factorization system of morphisms in a category, following the definitions of the Adámek, Herrlich, Strecker book ("Joy of cats"). The record `FactorizationSystem` is Definition 14.1 of that book, together with the properties that the morphism classes ℰ and ℳ are each closed under morphism equality ≈.

Concretely, this adds the following:

* In `Categories.Morphism.Lifts`:
    * A new record `Diagonal`, which is like `Filler` but does not depend on a commutativity proof.
    * A new record `MorphismClassMember` that allows a uniform treatment of morphisms in Epi, Mono, etc.
* In `Categories.Morphism.FactorizationSystem.Core`: The main definitions of `Factorization` and `FactorizationSystem` for a category 𝒞 and morphism classes ℰ and ℳ (as module parameters).
* In `Categories.Morphism.FactorizationSystem`: Same without module parameters 𝒞, ℰ, ℳ  to define the syntax `[ ℰ , ℳ ]-structured 𝒞`.

### DIFF
--- a/src/Categories/Morphism/FactorizationSystem/Core.agda
+++ b/src/Categories/Morphism/FactorizationSystem/Core.agda
@@ -1,0 +1,72 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Categories.Category
+open import Categories.Morphism.Lifts using (MorphismClass)
+
+
+-- Consider a category 𝒞 and two morphism classes ℰ and ℳ.
+-- The level of the morphism equalities is called `eq`
+-- such that the identifier `e` stays fresh for members of ℰ.
+module Categories.Morphism.FactorizationSystem.Core
+       {o ℓ eq} (𝒞 : Category o ℓ eq)
+       {ℓℰ ℓℳ} (ℰ : MorphismClass 𝒞 ℓℰ) (ℳ : MorphismClass 𝒞 ℓℳ) where
+
+open import Level
+
+open Category 𝒞
+open Definitions 𝒞 using (CommutativeSquare)
+open import Categories.Morphism.Lifts 𝒞 using (MorphismClassMember; ≈-closed; UniqueDiagonal; Diagonal)
+open import Categories.Morphism {o} {ℓ} {eq} 𝒞 using (IsIso)
+open MorphismClassMember using (mor)
+
+private
+  variable
+    A B C D : Obj
+
+-- We write ->> / ↠ for morphisms in ℰ
+_↠_ : Obj → Obj → Set (ℓ ⊔ ℓℰ)
+_↠_ = MorphismClassMember ℰ
+
+-- We write >-> / ↣ for morphisms in ℳ
+_↣_ : Obj → Obj → Set (ℓ ⊔ ℓℳ)
+_↣_ = MorphismClassMember ℳ
+
+record Factorization (h : A ⇒ B) : Set (o ⊔ ℓ ⊔ eq ⊔ ℓℰ ⊔ ℓℳ) where
+  field
+    Im : Obj
+    e : A ↠ Im
+    m : Im ↣ B
+    m∘e≈h : mor m ∘ mor e ≈ h
+
+-- A factorization system (also called factorization structure) for
+-- morphisms in the category 𝒞, see section 14 of "Abstract and
+-- Concrete Categories - The Joy of Cats" by Adámek, Herrlich, Strecker, 2004.
+record FactorizationSystem : Set (o ⊔ ℓ ⊔ eq ⊔ ℓℰ ⊔ ℓℳ) where
+  field
+    ℰ-resp-≈ : ≈-closed ℰ
+    ℳ-resp-≈ : ≈-closed ℳ
+    -- 1. every morphism admits an (ℰ,ℳ)-factorization
+    factor : ∀ (f : A ⇒ B) → Factorization f
+    -- 2. ℰ and ℳ are closed under compositions with isomorphisms:
+    Iso∘ℰ : ∀ {h : B ⇒ C} → IsIso h → (e : A ↠ B) → ℰ (h ∘ mor e)
+    ℳ∘Iso : ∀ (m : B ↣ C) → {h : A ⇒ B} → IsIso h → ℳ (mor m ∘ h)
+    -- 3. every commutative square of the form
+    --      e
+    --   A ───>> B
+    --   │     / │
+    --   │  ∃!╱  │
+    -- f │   ╱   │ g
+    --   │  ╱    │
+    --   V V     V
+    --    C >──> D
+    --      m
+    --  has a unique diagonal filler.
+    diagonal-filler : {f : A ⇒ C} {g : B ⇒ D} (e : A ↠ B) (m : C ↣ D)
+                      (comm : CommutativeSquare (mor e) f g (mor m))
+                      → UniqueDiagonal (mor e) f g (mor m)
+  diag-unique₂ : {f : A ⇒ C} {g : B ⇒ D} (e : A ↠ B) (m : C ↣ D)
+            → (d₁ d₂ : Diagonal (mor e) f g (mor m))
+            → Diagonal.d d₁ ≈ Diagonal.d d₂
+  diag-unique₂ {f = f} {g = g} e m d₁ d₂ =
+    UniqueDiagonal.unique₂ (diagonal-filler e m (Diagonal.comm d₁)) d₁ d₂
+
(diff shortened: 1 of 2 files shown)


t-wissmann:
This is the first part of the big changeset of #541:
* I have applied the suggested 'import first, open later' pattern
* I have replaced `hiding` with `using`

Regarding the `Filler` record: Prior to the present PR, the entire file `Lifts.agda` is touched by only two commits, both by @TOTBWF  on Jan 2, 2022:

* b1289e8a85cdcc08f2abf70b931aaf27138d6457: creates `record Lifts`
* a734f2f7658702ee083f128db244a4656dfccaf6: renames `Lifts` to `Filler`

According to `grep -nr Filler src/` the only references to this record are in `src/Categories/Morphism/Lifts/Properties.agda`.

TOTBWF:
RE: `Diagonal` and `Filler`; it might be worth deprecating `Filler` and taking `Diagonal` as the blessed definition. We don't parameterize things like `is-pullback` by a square for the reasons you've noted.

src/Categories/Morphism/FactorizationSystem/Core.agda
```diff
@@ -0,0 +1,72 @@
...
+  variable
+    A B C D : Obj
+
+-- We write ->> / ↠ for morphisms in ℰ
+_↠_ : Obj → Obj → Set (ℓ ⊔ ℓℰ)
+_↠_ = MorphismClassMember ℰ
+
+-- We write >-> / ↣ for morphisms in ℳ
+_↣_ : Obj → Obj → Set (ℓ ⊔ ℓℳ)
+_↣_ = MorphismClassMember ℳ
```
TOTBWF, line 33: Is there a reason to bundle these? I suspect it's going to make goal display much worse

src/Categories/Morphism/FactorizationSystem/Core.agda
```diff
@@ -0,0 +1,72 @@
...
+-- Concrete Categories - The Joy of Cats" by Adámek, Herrlich, Strecker, 2004.
+record FactorizationSystem : Set (o ⊔ ℓ ⊔ eq ⊔ ℓℰ ⊔ ℓℳ) where
+  field
+    ℰ-resp-≈ : ≈-closed ℰ
+    ℳ-resp-≈ : ≈-closed ℳ
+    -- 1. every morphism admits an (ℰ,ℳ)-factorization
+    factor : ∀ (f : A ⇒ B) → Factorization f
+    -- 2. ℰ and ℳ are closed under compositions with isomorphisms:
+    Iso∘ℰ : ∀ {h : B ⇒ C} → IsIso h → (e : A ↠ B) → ℰ (h ∘ mor e)
+    ℳ∘Iso : ∀ (m : B ↣ C) → {h : A ⇒ B} → IsIso h → ℳ (mor m ∘ h)
```
TOTBWF, line 53: Likewise, these don't seem to be the most composable form of the closure condition; one side takes a bundled form of an E/M morphism, but we get an unbundled one. This means that chaining these lemmas is going to require a bunch of wrapping/unwrapping.

t-wissmann:
Thanks, @TOTBWF!

As suggested by @JacquesCarette in #541, I have switched to the name 'FactorizationSystem' instead of 'FactorizationStructure'. My original intention was to stay close to the terminology in the book by Adamek, Herrlich, Strecker (who use 'factorization structure'), but in all my own papers, I always refer to it as 'factorization system', so I'm happy with using '... system'.

src/Categories/Morphism/FactorizationSystem/Core.agda
t-wissmann: It makes definitions much more readable, and it matches the common pen&paper notation. Examples from #541:
```
Dom↠Im[_] : ∀ {X} {Y} (f : X ⇒ Y) → X ↠ Im[ f ]
Im[_]↣Codom : ∀ {X} {Y} (f : X ⇒ Y) → Im[ f ] ↣ Y

record Factorization (h : A ⇒ B) : Set (o ⊔ ℓ ⊔ eq ⊔ ℓℰ ⊔ ℓℳ) where
  field
    Im : Obj
    e : A ↠ Im
    m : Im ↣ B
    m∘e≈h : mor m ∘ mor e ≈ h

diagonal-filler : {f : A ⇒ C} {g : B ⇒ D} (e : A ↠ B) (m : C ↣ D)
```
etc.

Also, there are already bundled versions explicitly for Mono/Epi in Categories/Morphism.agda
```
record _↣_ (A B : Obj) : Set (o ⊔ ℓ ⊔ e) where
  field
    mor  : A ⇒ B
    mono : Mono mor


record _↠_ (A B : Obj) : Set (o ⊔ ℓ ⊔ e) where
  field
    mor : A ⇒ B
    epi : Epi mor
```

src/Categories/Morphism/FactorizationSystem/Core.agda
t-wissmann: The statement is how one would write it on paper:
"For every iso h and every epi e, the composition h ∘ e *is* an epi".

For chaining these lemmas, I can add `_∘ℰ_ : (B ↠ C) → (A ↠ B)` to `Properties.agda`. But the present PR is (intentionally) minimal.

t-wissmann: Or do you prefer these record fields to use the unbundled version?
```agda
    Iso∘ℰ : ∀ {h : B ⇒ C} → IsIso h → {e : A ⇒ B} → ℰ e → ℰ (h ∘ e)
    ℳ∘Iso : ∀ {m : B ⇒ C} → ℳ m → {h : A ⇒ B} → IsIso h → ℳ (m ∘ h)
```
I'm not sure whether Agda can derive `m` and `e` from `ℳ m` and `ℰ e` when ℳ and ℰ are instantiated to concrete properties

src/Categories/Morphism/FactorizationSystem/Core.agda
```diff
@@ -0,0 +1,72 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Categories.Category
```
JacquesCarette:
```suggestion
open import Categories.Category.Core using (Category)
```

src/Categories/Morphism/FactorizationSystem/Core.agda
```diff
@@ -0,0 +1,72 @@
...
+-- Consider a category 𝒞 and two morphism classes ℰ and ℳ.
+-- The level of the morphism equalities is called `eq`
+-- such that the identifier `e` stays fresh for members of ℰ.
+module Categories.Morphism.FactorizationSystem.Core
+       {o ℓ eq} (𝒞 : Category o ℓ eq)
+       {ℓℰ ℓℳ} (ℰ : MorphismClass 𝒞 ℓℰ) (ℳ : MorphismClass 𝒞 ℓℳ) where
+
+open import Level
+
+open Category 𝒞
```
JacquesCarette: We tend to put the pure `open`s after the `open import` clauses.

src/Categories/Morphism/FactorizationSystem/Core.agda
JacquesCarette: Maybe `↠` and `↣` (in this file) should be made `syntax` instead of defined functions?

src/Categories/Morphism/Lifts.agda
```diff
@@ -42,6 +44,62 @@ record Filler {A B X Y} {i : A ⇒ B} {f : A ⇒ X} {g : B ⇒ Y} {p : X ⇒ Y}
...
 Lifts : ∀ {A B X Y} → (i : A ⇒ B) → (p : X ⇒ Y) → Set (ℓ ⊔ e)
 Lifts i p = ∀ {f g} → (comm : CommutativeSquare i f g p) → Filler comm
+-- The diagonal in a square. The record does not require the square to commute
+-- but rather deduces this from the existence of the diagonal.
+-- The reason is that if the record is parametrized by a morphism equality,
+-- then record types differ if there are multiple (unequal) proofs of
+-- morphism equality. Then, properties like FactorizationSystem.diag-unique₂
+-- would become more complicated. Thus, Diagonal does not use
+-- above Filler record, which is parametric in a morphism equality.
+record Diagonal {A B X Y} (i : A ⇒ B) (f : A ⇒ X)
```
JacquesCarette: Either rename `X` and `Y` to `C` and `D` in the record or the diagram -- but these should match!

JacquesCarette (changes requested):
Will also want to hear back from @TOTBWF

t-wissmann:
Syntax for ↠ and ↣ does not work because:
```
syntax must use holes exactly once
```
And we can't make the morphism class an implicit parameter here because ↠ and ↣ really must denote different things.

JacquesCarette:
Ah right, good point about `syntax`. Sigh.

PR #538, opened 2026-07-17
Develop dual calculus for rigid monoidal categories

LionOfJewdah:
## Summary

Develop the core calculus of dual objects and morphisms in rigid monoidal
categories.

This adds:

- cups, caps, transposes, mates, and dual morphisms
- uniqueness, naturality, snake, cancellation, identity, and composition laws
- conversion between left and right rigidity using symmetry
- agreement of the induced left and right dual morphisms
- left- and right-rigid projections from a compact-closed structure
- supporting monoidal reasoning laws

## Motivation

The immediate goal is the canonical Joyal–Street–Verity trace of a compact
closed category. Its cap–cup construction requires a practical calculus for
moving between dual objects, transposed morphisms, cups, caps, and the two
handed forms of rigidity.

This PR establishes that foundational calculus. Tensor-product dual coherence
and the canonical symmetric double-dual isomorphism are separated into the
stacked `feat/rigid-tensor-and-double-dual-coherence` branch, reducing the scope
of this review without removing any results.

The subsequent branches use this infrastructure to formalize the compact-closed
trace of [Joyal–Street–Verity 1996](https://doi.org/10.1017/S0305004100074338),
its uniqueness following
[Hasegawa 2009](https://doi.org/10.1017/S0960129508007184), and the
traced star-autonomous recognition theorem of
[Hajgató–Hasegawa 2013](https://doi.org/10.70930/tac/ir3wltz0).

PR #544, opened 2026-07-31
Add tensor products of enriched categories and functors

LionOfJewdah:
## Summary

Construct tensor products of enriched categories and enriched functors over a
monoidal category equipped with interchange.

This adds:

- the enriched category tensor product `𝒜 ⊠ ℬ`
- tensor products of enriched functors `F ⊠F G`
- compatibility between tensor products and enriched opposites
- the symmetry functor `swapF`
- canonical factor inclusions `includeˡ` and `includeʳ`
- symmetric interchange identities needed by these constructions
- naturality, cancellation, and unit laws for cyclic rotation of three tensor
  factors

## Motivation

For categories enriched over a monoidal category `V`, the tensor-product
category has pairs as objects and tensor products of hom objects:

    (𝒜 ⊠ ℬ) [ (A , X) , (B , Y) ] = 𝒜 [ A , B ] ⊗ ℬ [ X , Y ].

Composition must exchange the two middle hom objects before composing in each
factor. The existing `HasInterchange` interface provides exactly the
naturality, associativity, and unit coherence required to construct this
category without assuming that `V` is braided.

The notation `⊠` was chosen because it is visibly related to the underlying
tensor product `⊗`, while remaining distinct from it, and was previously unused
in this library. It is not essential to the construction; other notation would
be equally viable if maintainers prefer a different convention.

The same interchange calculus makes the construction functorial. When `V` is
symmetric, it additionally supplies the enriched swap functor, compatibility
with enriched opposites, and the canonical inclusions of either tensor factor.

These constructions provide generic enriched-category infrastructure needed
for hom bifunctors, enriched profunctors, and extraordinary naturality. In
particular, they will support the later formalization of the enriched argument
in Hajgató–Hasegawa’s construction of the compact-closed structure on
traced star-autonomous categories, without coupling that application to this foundational PR.

Independent of PRs #538 and #540, but similarly motivated.

### DIFF
--- a/src/Categories/Category/Monoidal/Interchange/Symmetric.agda
+++ b/src/Categories/Category/Monoidal/Interchange/Symmetric.agda
@@ -15,17 +15,17 @@ import Categories.Category.Construction.Core C as Core
 import Categories.Category.Monoidal.Braided.Properties as BraidedProps
 open import Categories.Category.Monoidal.Interchange using (HasInterchange)
 import Categories.Category.Monoidal.Interchange.Braided as BraidedInterchange
-  using (module swapInner; swapInner-braiding)
+  using (module swapInner; swapInner-braiding; swapInner-unitˡ)
 import Categories.Category.Monoidal.Reasoning M as MonoidalReasoning
 import Categories.Category.Monoidal.Utilities M as MonoidalUtilities
 open import Categories.Functor using (_∘F_)
 open import Categories.NaturalTransformation.NaturalIsomorphism
   using (_≃_; niHelper)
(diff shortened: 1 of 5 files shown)


JacquesCarette (approved):
Very nice!

LionOfJewdah:
I realize I did not include tensor product of enriched natural transformations and natural isomorphisms -- I will include that in a followup PR. I may also include enriched bifunctors and a Symmetric Monoidal Category of V-enriched categories, with tensor unit CommMonoidAsMonoidal (Scalars V)

